PR #45, opened 2014-12-09
Modular library

DrRayWang:
* modified utils.rb of libraries with module 
* modified all affected files in recipes


DrRayWang:
take too long to merge this

PR #57, opened 2014-12-18
Changes to copy logs from nodes into HDFS

bijugs:
Changes in the PR will enable copying of log files from nodes in the cluster to a centralized location currently HDFS. There is an additional commit to use node attribute to store HDFS URL.

**Note** Any user trying to copy logs from nodes to HDFS or change the current log locations, need to update the recipe for the corresponding component to set the correct key values for the [copylog node attribute](https://github.com/bijugs/chef-bach/blob/copy_logs/cookbooks/bcpc-hadoop/attributes/default.rb#L143-L151)

**Future work:**
1) Need to identify the access control to  read and copy syslog, auth.log and chef log to HDFS.
2) Identify the correct components to copy the logs from the bootstrap node into HDFS.
3) When changes are made to dynamically create `` site.xml `` and `` env.sh `` the file locations for copylog need to be updated as well.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/copylog.rb
+++ b/cookbooks/bcpc-hadoop/recipes/copylog.rb
@@ -1,0 +1,97 @@
+#
+# Recipe to copy log data from nodes into HDFS
+# Flume is used to copy data and runs as an agent on the nodes
+# 
+#**** Attention ***** Attention ****
+# This should be the last recipe in the runlist for any role since 
+# individual recipes can make requests to copy log files to HDFS
+# This can be achieved by adding the Copylog role as the last role to the node
+# Since flume writes into HDFS nodes running this recipe should have HDFS
+# client components installed on them.
+
+
+dpkg_autostart "flume-agent" do
+  allow false
+end
+
+%w{flume flume-agent}.each do |p|
+  package p do
+    action :upgrade
+  end
+end
+
+service "flume-agent" do
+  action [:stop, :disable]
+end
+
+bash "make_shared_logs_dir" do
+  code <<EOH
+  hdfs dfs -mkdir -p #{node['bcpc']['hadoop']['hdfs_url']}/user/flume/logs/ && \
+  hdfs dfs -chown -R flume #{node['bcpc']['hadoop']['hdfs_url']}/user/flume/
+EOH
+  user "hdfs"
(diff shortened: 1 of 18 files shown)


cookbooks/bcpc-hadoop/recipes/copylog.rb
```diff
@@ -0,0 +1,70 @@
...
+  hdfs dfs -mkdir -p #{node['bcpc']['hadoop']['hdfs_url']}/user/flume/logs/ && \
+  hdfs dfs -chown -R flume #{node['bcpc']['hadoop']['hdfs_url']}/user/flume/
+EOH
+  user "hdfs"
+  not_if "hdfs dfs -test #{node['bcpc']['hadoop']['hdfs_url']}/user/flume/logs/", :user => "hdfs"
+end
+
+service "flume-agent" do
+  supports :status => true, :restart => true, :reload => false
+  action [:stop]
```
cbaenziger: Do we want to disable too? It seems this might be a case for disabling this package via dpkg_autostart?

cookbooks/bcpc-hadoop/recipes/copylog.rb
```diff
@@ -0,0 +1,70 @@
...
+    service "flume-agent-multi" do
+      supports :status => true, :restart => true, :reload => false
+      action [:enable, :start]
+      subscribes :restart, "template[/etc/flume/conf/flume-env.sh]", :delayed
+    end
+  end
+else
+  service "flume-agent-multi" do
+    supports :status => true, :restart => true, :reload => false
+    action [:disable]
```
cbaenziger: Why does it make sense to only disable and not stop the service here?

cookbooks/bcpc-hadoop/recipes/copylog.rb
bijugs: Yes. We can use `` dpkg_autostart `` to disable autostart.

cookbooks/bcpc-hadoop/recipes/copylog.rb
bijugs: This logic is to allow users to disable the **copy logs** feature by setting the  `` default['bcpc']['hadoop']['copylog_enable'] = true ``

kiiranh:
LGTM

cbaenziger:
Some upstream issues captured here:
* [BIGTOP-1581](https://issues.apache.org/jira/browse/BIGTOP-1581)
* [FLUME-2573](https://issues.apache.org/jira/browse/FLUME-2573)

Issue #69, opened 2015-01-20
bcpc_jmxtrans::default returns invalid date


cbaenziger:
On a number of physical clusters we are seeing:
````
  * service[restart jmxtrans on dependent service manual restart] action restart
================================================================================
Error executing action `restart` on resource 'service[restart jmxtrans on dependent service manual restart]'
================================================================================


ArgumentError
-------------
invalid date


Cookbook Trace:
---------------
/var/chef/cache/cookbooks/bcpc_jmxtrans/libraries/utils.rb:50:in `parse'
/var/chef/cache/cookbooks/bcpc_jmxtrans/libraries/utils.rb:50:in `block (2 levels) in process_require_restart?'
/var/chef/cache/cookbooks/bcpc_jmxtrans/libraries/utils.rb:48:in `each'
/var/chef/cache/cookbooks/bcpc_jmxtrans/libraries/utils.rb:48:in `block in process_require_restart?'
/var/chef/cache/cookbooks/bcpc_jmxtrans/libraries/utils.rb:44:in `each'
/var/chef/cache/cookbooks/bcpc_jmxtrans/libraries/utils.rb:44:in `process_require_restart?'
/var/chef/cache/cookbooks/bcpc_jmxtrans/recipes/default.rb:88:in `block (2 levels) in from_file'
````

cbaenziger:
Instrumenting the exception I see the issue:
````invaid data target process: 20:19
, dependent process:  2014````

We need to expect all variants of ps's semantics; or pick a format which does not change. From `ps`(1) I see for `start_time`:
````
start_time  START     starting time or date of the process.  Only the year will be displayed if the process was not started the same year ps was invoked, or "MmmDD" if it was not started the same day, or "HH:MM" otherwise.  See also bsdstart, start, lstart, and stime.
````

I think `lstart` would be a better field choice, as it returns strings like `Fri Aug  8 12:15:42 2014`
Sadly, its entry in `ps`(1) is not quite descriptive of its return though:
`time the command started.`

Issue #68, opened 2015-01-19
Need mapred user on namenode for mapreduce usage


cbaenziger:
On a test cluster, one currently sees the following errors on the namenode when trying to run YARN jobs:
````2015-01-16 23:57:32,456 WARN org.apache.hadoop.security.UserGroupInformation: No groups available for user mapred
2015-01-16 23:57:32,457 ERROR org.apache.hadoop.security.UserGroupInformation: PriviledgedActionException as:mapred (auth:SIMPLE) cause:org.apache.hadoop.security.AccessControlException: Permission denied: user=mapred, access=READ_EXECUTE, inode="/user/history/done_intermediate/ubuntu":ubuntu:mapred:drwxrwx---````

This is due to the default in our clusters of [resolving a user's group on the namenode](http://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/HdfsPermissionsGuide.html#Group_Mapping).

Issue #63, opened 2015-01-09
Restarting jmxtrans takes a long time and is done often


cbaenziger:
It seems that we restart jmxtrans quite often when updating configuration files. Instead of restarting jxmtrans once if a service had restarted, we seem to restart it once per service. I suspect we need to set the service resource name to be the same at [bcpc_jmxtrans::default.rb#L74-L79](https://github.com/bloomberg/chef-bach/blob/c835b8e86f9ded3fc02d8abff96f3f7923da02c2/cookbooks/bcpc_jmxtrans/recipes/default.rb#L85-L89) and [bcpc_jmxtrans::default.rb#L74-L79](https://github.com/bloomberg/chef-bach/blob/c835b8e86f9ded3fc02d8abff96f3f7923da02c2/cookbooks/bcpc_jmxtrans/recipes/default.rb#L85-L89) opposed to service specific.

We could use a [log](https://docs.chef.io/resource_log.html) resource to report which services for which we are bouncing jmx first to still give that insight. And we could create a block of subscribes and guard conditions for a single service resource.

I question if we are not restarting jmxtrans when we should not need to as well.

cbaenziger:
An example where we restart jmxtrans twice is in this case where the datanode and regionserver were believed to have been restarted:

````
Recipe: jmxtrans::default
  * user[jmxtrans] action create (up to date)
  * ark[jmxtrans] action installRecipe: <Dynamically Defined Resource>
  * directory[/opt/jmxtrans-latest] action create (up to date)
  * remote_file[/var/chef/cache/jmxtrans-latest.zip] action create (up to date)
  * execute[unpack /var/chef/cache/jmxtrans-latest.zip] action nothing (skipped due to action :nothing)
  * execute[set owner on /opt/jmxtrans-latest] action nothing (skipped due to action :nothing)
  * link[/opt/jmxtrans] action create (up to date)
  * template[/etc/profile.d/jmxtrans.sh] action create (skipped due to only_if)
  * ruby_block[adding '/opt/jmxtrans-latest/bin' to chef-client ENV['PATH']] action run (skipped due to only_if)
 (up to date)
Recipe: jmxtrans::default
  * file[/opt/jmxtrans/jmxtrans.sh] action touch
    - update utime on file /opt/jmxtrans/jmxtrans.sh

  * template[/etc/init.d/jmxtrans] action create (up to date)
  * template[/etc/default/jmxtrans] action create (up to date)
[...]

Issue #49, opened 2014-12-16
Add topology script


cbaenziger:
We should have a way to provide a cluster topology script via a configurable template for HDFS rack awareness.

PR #70, opened 2015-01-20
Fix for #49, YARN-2516, #63, #68, #69 and tunables

cbaenziger:
Provides fixes for:
* HDFS Topology Hooks (Issue #49)
* Heap size tuning for the Mapred Job History Server
* Subscribes for YARN and Mapred Daemons
* HDFS blocksize as a tunable (it will not retroactive change blocksizes if changed, however)
* Sets a minimum share to cause YARN Fair Scheduler pre-emption
* ~~Enables HBase snapshot support~~
* Adds mapred user on namenode for mapreduce usage, via installing the hadoop-mapreduce package (Issue #68)
* Modifies bcpc_jmxtrans to:
  * Use `ps`(1)'s `lstart` instead of `start_time` (Issue #69)
  * Use one `service[restart jmxtrans on dependent service]` to avoid restarting JMX Trans multiple times needlessly (Issue #63)


kiiranh:
Looks good.

cbaenziger:
I removed the HBase snapshot commit.

Issue #72, opened 2015-01-20
Vagrant Baremetal Bring-Up Assumes HTTP(S) Proxy


cbaenziger:
In [kinfe.rb.erb](https://github.com/bloomberg/chef-bach/blob/1552ee81e6b015dda37fc54d606eabbedfac89c0/cookbooks/bcpc/templates/default/knife.rb.erb#L22-L23) used by the [bcpc::bootstrap_knife](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc/recipes/bootstrap_knife.rb) recipe, it is assumed the environment is using http_proxies. We should make this dynamic based on the environment.

PR #73, opened 2015-01-22
Fix for Vagrant Baremetal Bring-Up Assumes HTTP(S) Proxy #72

cbaenziger:
A fix for Vagrant Baremetal Bring-Up Assumes HTTP(S) Proxy #72


kiiranh:
Looks good.

Issue #79, opened 2015-02-05
get_req_node_attributes fails a chef run with unhelpful "undefined method `[]' for nil:NilClass"


cbaenziger:
If one is searching a node for an attribute it does not have, instead of properly erroring out and reporting the node object was incomplete, we today see the following failure:
````
================================================================================
Recipe Compile Error in /var/chef/cache/cookbooks/kafka-bcpc/recipes/setattr.rb
================================================================================


NoMethodError
-------------
undefined method `[]' for nil:NilClass


Cookbook Trace:
---------------
  /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:239:in `block (3 levels) in get_req_node_attributes'
  /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:239:in `each'
  /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:239:in `reduce'
  /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:239:in `block (2 levels) in get_req_node_attributes'
  /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:238:in `each'
  /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:238:in `block in get_req_node_attributes'
  /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:236:in `each'
  /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:236:in `get_req_node_attributes'
  /var/chef/cache/cookbooks/kafka-bcpc/recipes/setattr.rb:16:in `from_file'


Relevant File Content:
----------------------
/var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:

232:  # [ { :node_number => "val", :hostname => "nameval" }, ...]
233:  #
234:  def get_req_node_attributes(node_objects,srch_keys)
235:    result = Array.new
236:    node_objects.each do |obj|
237:      temp = Hash.new
238:      srch_keys.each do |name, key|
239>>       val = key.split('.').reduce(obj) {|memo, key| memo[key]}
240:        temp[name] = val
241:      end
242:      result.push(temp)
243:    end
244:    return result
245:  end
246:
247:  #
248:  # Restarting of hadoop processes need to be controlled in a way that all the nodes
````

PR #80, opened 2015-02-05
Fix for #79: get_req_node_attributes fails a chef run with unhelpful "undefined method `[]' for nil:NilClass"

cbaenziger:
This will print out a log message showing the node object and which key is missing for easier debugging. The Chef run will stop and error at that point as well.


kiiranh:
Does the log message have to be of level 'FATAL'? Apart from that the change looks good.

cbaenziger:
@kiiranh since this aborts the Chef run, it seems appropriate that it be a `FATAL` message.

PR #81, opened 2015-02-09
Changes to install correct hive config files on all worker nodes

bijugs:
Currently default config files are getting installed on worker nodes resulting in hive actions ending up using local Derby DB instead of the common mysql metastore. This is a fix for the chef-bcpc issue [#338](https://github.com/bloomberg/chef-bcpc/issues/338).


kiiranh:
looks good

kiiranh:
I tested this on VM cluster and hive-config seems to have run on vm3: /etc/hive/conf structure and content is same as that on vm2. Not sure what to verify to make sure hive is working correctly.

PR #84, opened 2015-02-10
macOS_sed_fix

RayAtBloomberg:
Description: tests/automated_install has a encode issue with MAC OS. Because MAC OS uses BSD version SED, while ubuntu uses Unix version SED. The character '\n' encodes different in these two versions.

My solution: use two patterns in SED. The first will put an '\n' in needing newline place, second pattern will replace this \n with real newline character.


amithkanand:
It breaks on Linux environment. Adds proxy after ``bootstrap:`` instead of  putting it inside the curly brace ``bootstrap: {``.

kiiranh:
I tested this on OSX and I see extraneous files generated in the chef-bcpc folder viz. Vagrantfile"" and proxy_setup.sh"". Would it be possible to not have these created ?

kiiranh:
Tested in LL2 ubuntu box. The environment file was modified as expected and no extraneous files were generated.

bijugs:
Thanks @kiiranh for additional testing.

PR #85, opened 2015-02-12
Changes to copy Oozie sharelibs to HDFS

bijugs:
In the current code Oozie sharelibs are not being copied into HDFS. This PR is to fix the issue.


kiiranh:
I tested this change on my VM cluster. I see the libraries being put in /user/oozie/share/lib directory.

PR #87, opened 2015-02-17
Programmatic hadoop disks

cbaenziger:
This pull request allows us to have dynamic configuration of Hadoop disks -- where `node[:bcpc][:hadoop][:disks]` and `node[:bcpc][:hadoop][:mounts]` change within a single Chef run. This can occur for a few different reasons, one is inspection of a machine's storage controller after having setup the disks within the Chef run or handling a failed disk within a Chef run.


cbaenziger:
Squashed into one commit per popular demand.

cbaenziger:
Need to still fix-up at least journalnode.rb

cbaenziger:
The dynamically created resources do not seem to end up in the `resource_collection` it would be good to understand if they affect the Chef run's metrics (e.g. for the Chef graphite handler).

PR #90, opened 2015-02-17
Stash Maven in bins to fix #88

### LINKED ISSUE #88: Hannibal Pulls in Maven and Breaks Internet Disconnected Clusters
Today, if one is running a bootstrap node copying only the bins directory then they will bomb out with:
````
[2015-02-17T19:57:35+00:00] ERROR: ark[maven] (maven::default line 28) had an error: Chef::Exceptions::MultipleFailures: Multiple failures occurred:
* Errno::ETIMEDOUT occurred in chef run: remote_file[/var/chef/cache/maven-3.1.1.tar.gz] (/var/chef/cache/cookbooks/ark/providers/default.rb line 45) had an error: Errno::ETIMEDOUT: Error connecting to http://apache.mirrors.tds.net/maven/maven-3/3.1.1/binaries/apache-maven-3.1.1-bin.tar.gz - Connection timed out - connect(2)
````

This is because today we have a [`Hannibal-Build`](https://github.com/bloomberg/chef-bach/commit/7d98fde7acb829282d407e70b007d3894c764b0f) role which is not guarded on having a pre-built Hannibal.

Ideally, this would be idempotent for a bootstrap node on the Internet (and building Hannibal) or off the Internet (with a pre-built Hannibal).

cbaenziger:
This fixes issues #88 by stashing Maven in `/home/vagrant/chef-bcpc/bins` for later use.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/maven_config.rb
+++ b/cookbooks/bcpc-hadoop/recipes/maven_config.rb
@@ -1,3 +1,15 @@
-# Override maven attributes
+require 'pathname'
 
 node.override['maven']['install_java'] = false
+
+internet_download_url = node['maven']['3']['url']
+maven_file = Pathname.new(node['maven']['3']['url']).basename
+
+node.override['maven']['3']['url'] = "#{get_binary_server_url}/#{maven_file}"
+
+# download Maven only if not already stashed in the bins directory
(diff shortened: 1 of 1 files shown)


kiiranh:
Tested this on my test VM. Maven tar is only downloaded once and stored in bins. Also confirmed that hannibal is built

Issue #92, opened 2015-02-25
Chef test vms in parallel in automated_install.sh


kiiranh:
Currently, in tests/automated_install.sh lines 79-82, there is a for loop which chefs (cluster-assign-roles) each Hadoop vm serially. This does not setup the services on the vms correctly. 
Replace the for loop with a single cluster-assign-roles statement without specifying a VM.
This is supposed to set /etc/hosts correctly on all the nodes without having to chef them in a particular order.

cbaenziger:
Fixed in #94

PR #94, opened 2015-02-27
Modify automated_install to chef all vms in parallel

kiiranh:
This PR enables chef'ing all the VM nodes in parallel instead of serially in a for loop. Refer issue #92 

EDIT: This is supposed to set /etc/hosts correctly on all the nodes without having to chef them in a particular order.

### DIFF
--- a/tests/automated_install.sh
+++ b/tests/automated_install.sh
@@ -76,10 +76,8 @@ printf "Snapshotting post-Cobbler\n"
 printf "#### Chef all the nodes\n"
 vagrant ssh -c "sudo apt-get install -y sshpass"
 
-for i in 1 2 3; do
-  printf "#### Chef machine bcpc-vm${i}\n"
-  vagrant ssh -c "cd chef-bcpc; ./cluster-assign-roles.sh $ENVIRONMENT Hadoop bcpc-vm$i"
-done
+printf "#### Chef machine bcpc-vms\n"
+vagrant ssh -c "cd chef-bcpc; ./cluster-assign-roles.sh $ENVIRONMENT Hadoop"
 
 printf "Snapshotting post-Cobbler\n"
 VBoxManage snapshot bcpc-vm1 take Full-Shoes


kiiranh:
Tested this on OSX and Ubuntu. 

* OSX: 
** Sometimes the VIP doesn't come up correctly on VM1 after the initial chef, and so all services do not start.
** Had to rechef the cluster by running cluster-assign-roles again. After this, the cluster was happy (verified using [Verification Steps|https://gist.github.com/bijugs/e01bfbd81f8d09f7e8c2]).

* Ubuntu:
** The VIP was up, but had to rechef twice: the first time hdfs came up and the second time hbase came up.
** Post this verified that the cluster functions using the above mentioned verification steps.

bijugs:
@kiiranh , the VIP not coming up is consistent from past observations when ``cluster-asign-role`` is run to chef all the nodes in parallel. Some time this requires manual intervention i.e. stop start ``keepalived``, ``haproxy`` manually by logging on to the head nodes. In the previous approach the state was consistent for a user even though they need to run ``cluster-assign-roles`` multiple times. We need to weigh which one to follow if we are not able to resolve the ``VIP`` issue.

tests/automated_install.sh
```diff
@@ -76,10 +76,8 @@ printf "Snapshotting post-Cobbler\n"
 printf "#### Chef all the nodes\n"
 vagrant ssh -c "sudo apt-get install -y sshpass"
 
-for i in 1 2 3; do
-  printf "#### Chef machine bcpc-vm${i}\n"
-  vagrant ssh -c "cd chef-bcpc; ./cluster-assign-roles.sh $ENVIRONMENT Hadoop bcpc-vm$i"
-done
+printf "#### Chef machine bcpc-vms\n"
+vagrant ssh -c "cd chef-bcpc; ./cluster-assign-roles.sh $ENVIRONMENT Hadoop"
```
amithkanand: This will still behave the same way. With look or without loop machine is only being chef'ed once. We will still have to re-chef the machines.

kiiranh:
@bijugs It is not consistent. As noted above, on ubuntu VM cluster, VIP was up. So we can observe this during our runs and then see how it works out. We've been using the previous approach for about a year now. Lets use this approach for a few weeks and then decide what is to be done.
Also, even in this parallel approach, I'm seeing that we need to rechef twice. I haven't had luck getting all services up in just a single run.

bijugs:
@kiiranh, I meant my past observations. :-).

kiiranh:
On the VM cluster, I see that with a single parallel chef run on all the 3 nodes, the /etc/hosts entries are made for the 3 IPs for the 3 interfaces (float, storage, management)

PR #95, opened 2015-03-03
Support arbitrary file system and mout point for Hadoop disks

cbaenziger:
This moves what was in `bcpc-hadoop::disks` to `bcpc::disks` and provides:
* Arbitrary mountpoint location for the disks
* Arbitrary file system support for the disks
  * ext4 & xfs supported out of the box
  * easily modifiable mount options

This is dependent on chef-bach #87 being accepted first.


cookbooks/bcpc-hadoop/recipes/datanode.rb
```diff
@@ -76,40 +80,46 @@
 end
 
 # Setup datanode and nodemanager bits
-if node[:bcpc][:hadoop][:mounts].length <= node[:bcpc][:hadoop][:hdfs][:failed_volumes_tolerated]
-  Chef::Application.fatal!("You have fewer #{node[:bcpc][:hadoop][:disks]} than #{node[:bcpc][:hadoop][:hdfs][:failed_volumes_tolerated]}! See comments of HDFS-4442.")
+ruby_block "verify Hadoop mounts" do
+  block do
+    if node[:bcpc][:storage][:mounts].length <= node[:bcpc][:hadoop][:hdfs][:failed_volumes_tolerated]
```
cbaenziger: If the machine looses too many disks and drops below `node[:bcpc][:hadoop][:hdfs][:failed_volumes_tolerated]` disks, we will refuse to Chef anything beyond this in the Chef converge phase. I don't know if we want a `raise` or simply a `Chef::Log.warn` as HDFS will do the right thing -- and this was to prevent manual mis-specification.

cookbooks/bcpc-hadoop/recipes/journalnode.rb
```diff
@@ -17,32 +19,35 @@
     group "hdfs"
     user 0644
     content Base64.decode64(get_config("namenode_txn_fmt"))
-    not_if { node[:bcpc][:hadoop][:mounts].all? { |d| File.exists?("/disk/#{d}/dfs/jn/#{node.chef_environment}/current/VERSION") } }
+    not_if { lazy { node[:bcpc][:storage][:mounts].all? { |d| File.exists?("#{mount_root}/#{d}/dfs/jn/#{node.chef_environment}/current/VERSION") } } }
```
cbaenziger: This should have a `call()`?

cbaenziger:
Though this offered some nice features, the dynamic resource creation of so many resources was REALLY ugly it seemed. I would much prefer a cleaner approach than a compile, converge (with a recompile and converge ruby block approach).

PR #97, opened 2015-03-05
Configure heap size for Hive processes

amithkanand:
This PR will allow configuring a custom heap size for Hive processes `hive-metastore` and `hiveserver2`.

### DIFF
--- a/cookbooks/bcpc-hadoop/templates/default/hv_hive-env.sh.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/hv_hive-env.sh.erb
@@ -6,3 +6,4 @@ export HIVE_CONF_DIR=/etc/hive/conf.<%= node.chef_environment %>
 # export HIVE_AUX_JARS_PATH=
 export HIVE_AUX_JARS_PATH=/usr/lib/hcatalog/share/hcatalog/hcatalog-core.jar
 
+export HADOOP_HEAPSIZE=<%=node[:bcpc][:hive][:heap][:size] %>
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc-hadoop/templates/default/hv_hive-env.sh.erb
```diff
@@ -6,3 +6,4 @@ export HIVE_CONF_DIR=/etc/hive/conf.<%= node.chef_environment %>
 # export HIVE_AUX_JARS_PATH=
 export HIVE_AUX_JARS_PATH=/usr/lib/hcatalog/share/hcatalog/hcatalog-core.jar
 
+export HADOOP_HEAPSIZE=<%=node[:bcpc][:hadoop][:heapsize] %>
```
kiiranh: Since the heap size is for hadoop and not specifically for hive, why not make it in bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb ?

amithkanand: I am not sure if we really want to do that. Specifying it in `hdp_hadoop-env.sh.erb` will impact other hadoop daemons as well. With this PR we still set it to default causing no impact to any other process. Since `hive-env.sh` will include this as environment overridden value, it will only impact hive processes.

kiiranh: Sounds good.

cbaenziger:
Please use a different name from `node.bcpc.hadoop.heapsize` as this is Hive specific.

PR #98, opened 2015-03-12
Fix solr poll loop to only run for ZK nodes

amithkanand:
This PR fixes installation of Kafka. Solr update should only be checked in case of `Kafka Zookeeper` installation and not for `Kafka Server` installation.


cbaenziger:
Look good

PR #99, opened 2015-03-16
Disable the Chef-Server WebUI

cbaenziger:
The Chef Web UI is not something we use and as [Noah Kantrowitz](https://coderanger.net/chef-webui/) points out it is going away as it currently stands. Let's disable it.


kiiranh:
LGTM

amithkanand:
LGTM as well.

Issue #101, opened 2015-03-17
Use Berkshelf to manage cookbook dependencies


kiiranh:
Currently, cookbook dependencies are downloaded in a a shell script: `setup_chef_cookbooks.sh`. Instead, we should use Berkshelf to manage these dependencies.

PR #103, opened 2015-03-18
Remove unused inherited bcpc-centos cookbook

kiiranh:
This PR removes unused code inherited from bcpc. Some changes are as simple as deleting cookbooks entirely and some changes will require carefully modifying references to deleted code in the used code(e.g.: library code in `bcpc` uses `Role[BCPC-Headnode]` but that role is not used in chef-bach).

This fixes issue #102

PR #104, opened 2015-03-20
Manage cookbook dependencies using berkshelf

kiiranh:
Added Berksfile in all the cookbooks to correctly manage the cookbook dependencies.

This fixes issue #101


kiiranh:
Currently berkshelf does not correctly handle transitive dependencies for cookbooks cloned from git. Hence, a cookbook, say 'G', downloaded from git needs to be specified in Berksfile of cookbook, say 'A' that needs the git cookbook and in other cookbooks, say 'B' and 'C' where the dependency relation is of the form C -> B -> A -> G (where '->' == 'depends on').

kiiranh:
Should this PR also cleanup the code in setup_chef_cookbooks.sh which downloads these dependent cookbooks. If yes. we need to add a step to berks install and vendor the cookbooks before setting up the cluster.

kiiranh:
The changes have been tested by building a vbox vm cluster and by my Jenkins server. Good for review and merge. I can squash commits further if desired.

cbaenziger:
The centos cookbook should be gone from #103 now

tests/automated_install.sh
```diff
@@ -38,6 +43,18 @@ $SEDINPLACE 's/vb.gui = true/vb.gui = false/' Vagrantfile
...
 [[ -n "$PROXY" ]] && $SEDINPLACE -e "s#\(\"bootstrap\": {\)#\1\\\n\"proxy\" : \"http://$PROXY\",#" -e $'s/\\\\n/\\\n/' environments/${ENVIRONMENT}.json
 
+# download dependent/third-party cookbooks: begin
+[[ -n "$PROXY" ]] && export http_proxy=http://$PROXY; export https_proxy=http://$PROXY;
+export SSL_CERT_FILE="$PWD/bins/cacert.pem"
+rm -rf /tmp/cookbooks
+pushd cookbooks
+find . -maxdepth 1 -type d '!' -name "*bcpc*" -a '!' -name "hannibal" -a '!' -name "hdfsdu" -a '!' -name '\.'| xargs rm -rf
+popd
+find ./cookbooks -name "Berksfile" -exec berks -b {} \; -exec berks vendor /tmp/cookbooks -b {} \; -exec rsync -ar /tmp/cookbooks . \;
```
bijugs: @kiiranh . What will happen if any of the community cookbook doesn't use Berkshelf? Also we may download more cookbooks if any of the external cookbooks have groups defined in Berksfile as in the case of [kafka-cookbook](https://github.com/mthssdrbrg/kafka-cookbook/blob/master/Berksfile#L5-L8) where a ``test`` group is defined which is supposed to be skipped using ``berks install --except test``.

tests/automated_install.sh
```diff
@@ -6,6 +6,11 @@
 # simple few environment variables to tune booting for fast testing of Chef-BCPC
 # Run this script in the root of the git repository
 #
+# Pre-requisite software:
+# 1. curl
+# 2. virtualbox
+# 3. vagrant
+# 4. chefdk (w/ berkshelf)
```
cbaenziger: We want to be as lightweight as possible on the hypervisor host; please restrict to installing and running berks on the bootstrap node.

http-418:
PR #616 supercedes this one.

PR #107, opened 2015-03-23
chef vault download and install

RayAtBloomberg:
* download chef vault gem package in recipe, and chef vault cookbook in setup chef cookbook
* install chef vault gem package in recipe
* add chef vault download and install recipe in runlist of BCPC-Bootstrap role

### DIFF
--- a/cookbooks/bcpc/recipes/chef_vault_download.rb
+++ b/cookbooks/bcpc/recipes/chef_vault_download.rb
@@ -1,0 +1,55 @@
+#
+# Cookbook Name:: bcpc
+# Recipe:: chef vault install
+#
+# Copyright 2015, Bloomberg Finance L.P.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+target_filename = node['bcpc']['chefvault']['filename']
+link_filename = "chef-vault.gem"
+package_path = "#{node['bcpc']['bin_dir']['gems']}/#{target_filename}"
+checksum = node['bcpc']['chefvault']['checksum']
+
+
+if !File.exists?(package_path)
+
+# download rubygem from website
+  remote_file "#{ target_filename }" do
+    source "#{ node['bcpc']['gem_source'] }/#{ target_filename }"
+    checksum checksum
+    group "root"
+    owner "root"
+    mode "755"
+    path package_path
+    action :nothing
+  end.run_action(:create_if_missing)
+  
+  link "#{node['bcpc']['bin_dir']['gems']}/#{ link_filename }" do
+    owner 'root'
+    to "#{node['bcpc']['bin_dir']['gems']}/#{ target_filename }"
+    action :nothing
+  end.run_action(:create)
+  
+  bash 'build_bins' do
+    action :nothing
+    user 'root'
+    cwd '/home/vagrant/chef-bcpc'
+    code <<-EOH
+      ./build_bins.sh
+    EOH
+    umask 0002
+  end.run_action(:run)
+
+end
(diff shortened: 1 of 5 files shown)


cbaenziger:
Hi Ray, these two commits need to be rebased into one commit please; on cursory read through this looks good.

cookbooks/bcpc/recipes/chef_vault_download.rb
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+target_filename = "chef-vault-2.2.4.gem"
+link_filename = "chef-vault.gem"
+package_path = "#{node['bcpc']['bin_dir']['gems']}/#{target_filename}"
+checksum = "8d89c96554f614ec2a80ef20e98b0574c355a6ea119a30bd49aa9cfdcde15b4a"
```
cbaenziger: Please make `target_filename` and `checksum` attributes.

cookbooks/bcpc/recipes/chef_vault_download.rb
```diff
@@ -0,0 +1,50 @@
...
+
+# download rubygem from website
+remote_file "#{ target_filename }" do
+  source "#{ node['bcpc']['gem_source'] }/#{ target_filename }"
+  checksum checksum
+  group "root"
+  owner "root"
+  mode "755"
+  path package_path
+  action :nothing
```
cbaenziger: I would think we want a guard here to not keep downloading the file (e.g. a cluster which does not have internet access but does have bins pre-stashed).

cookbooks/bcpc/recipes/chef_vault_download.rb
```diff
@@ -0,0 +1,50 @@
...
+end.run_action(:create)
+
+bash 'build_bins' do
+  action :nothing
+  user 'root'
+  cwd '/home/vagrant/chef-bcpc'
+  code <<-EOH
+    ./build_bins.sh
+  EOH
+  umask 0002
```
cbaenziger: Again, this is a somewhat long operation (~2 minutes on some machines) a guard would be useful here.

cookbooks/bcpc/recipes/chef_vault_install.rb
```diff
@@ -0,0 +1,38 @@
...
+require 'pathname'
+require 'rubygems'
+
+gem_path = Pathname.new(Gem.ruby).dirname.join("gem").to_s
+chef_gemspec = "/opt/chef/embedded/lib/ruby/gems/1.9.1/specifications/chef-vault-2.2.4.gemspec"
+
+gem_package "chef-vault" do
+  gem_binary gem_path
+  version ">=0.0.0"
+  action :nothing
```
cbaenziger: Let's see if you can set a `umask` attribute on a `gem_package` resource to work around the restrictive [knife.rb](https://github.com/bloomberg/chef-bach/blob/5c07b64905beb52b32e876b0ac8a5f2a0de14b95/setup_chef_cookbooks.sh#L55) umask. I would like that more than a file action to simply change the permissions; only I suspect umask is not going to be supported.

Otherwise, if the `gem_package` does not support umask, could you add a comment on line 34 that the file resource is only to set the permissions? Further, it seems the `Gem.clear_paths` should be under the file action as that is really only when the gem is "deployed" -- though here we are running as root so get away with it before the file action.

RayAtBloomberg:
* put target_filename and checksum in node attributes
* put the guard, and not to download the chef vault and build bins twice
* put Gem.clear_paths at the bottom
* gem_package does not have umask attribute
* add comment for file resource to set the permissions.

setup_chef_cookbooks.sh
```diff
@@ -71,3 +71,4 @@ if [[ ! -d kafka ]]; then
 fi
 [[ -d jmxtrans ]] || git clone https://github.com/jmxtrans/jmxtrans-cookbook.git jmxtrans
 [[ -d cobblerd ]] || git clone https://github.com/cbaenziger/cobbler-cookbook.git cobblerd -b cobbler_profile
+[[ -d chef-vault ]] || git clone https://github.com/opscode-cookbooks/chef-vault.git chef-vault
```
cbaenziger: Ray, is there any reason the git clone of `chef-vault` is needed rather than a `knife cookbook download` of it? Usually this is only done for cookbooks which are not in the [supermarket](https://supermarket.chef.io/cookbooks/chef-vault); right now both look to be at version 1.3.0.

cbaenziger:
BREAKING: On a pre-existing cluster one needs provide the `chef-vault` gem

Issue #108, opened 2015-03-24
Time on Graphite graphs are off by 1 hour 


bijugs:
Currently the current time in the X axis of graphs created in Graphite is delayed by one hour. This is due to the default value of  [America/Chicago for Timezone parameter](http://graphite.readthedocs.org/en/latest/config-local-settings.html#general-settings) which is not being changed during the install.

bijugs:
Fixed as part of PR #109 .

PR #109, opened 2015-03-24
Change timezone to show the correct time on Graphite graphs

bijugs:
This is to fix issue #108.

### DIFF
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -149,6 +149,7 @@ default['bcpc']['keepalived']['config_template'] = "keepalived.conf_openstack"
 default['bcpc']['graphite']['relay_port'] = 2013
 default['bcpc']['graphite']['web_port'] = 8888
 default['bcpc']['graphite']['log']['retention'] = 15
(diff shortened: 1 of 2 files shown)


kiiranh:
LGTM. Verified this on my VM cluster. Graphite shows graph in EST.

cookbooks/bcpc/attributes/default.rb
```diff
@@ -149,6 +149,7 @@
 default['bcpc']['graphite']['relay_port'] = 2013
 default['bcpc']['graphite']['web_port'] = 8888
 default['bcpc']['graphite']['log']['retention'] = 15
+default['bcpc']['graphite']['timezone'] = "'America/New_York'"
```
cbaenziger: Could this come from `/etc/timezone`?

cbaenziger:
I had asked could the timezone come from `/etc/timezone`; but on reflection, we should likely keep all TZ selection in Chef as we are not necessarily writing a generic Graphite cookbook here -- and what if one has a Graphite cluster which spans machines with different timezones (then perhaps one would sanely use UTC anyways).

PR #111, opened 2015-03-25
Fix embedded ruby block to set oozie java_home

kiiranh:
Fixes the embedded ruby code which set Java home in oozie environment script.


cbaenziger:
Good find; thanks Kiran.

PR #113, opened 2015-03-30
Support Trusty APT mirrors

cbaenziger:
This adds support for Trusty APT mirrors. Ideally, we would follow something using the `mirrors.txt` file (akin to this [Gist](https://gist.github.com/PeteLawrence/9052977) by Pete Lawrence) but it seems `apt-mirror` on Precise reports the following using `mirrror://` with a `mirrors.txt`:
````
Proceed indexes: [Psh: 1: cannot open mirrors.ubuntu.com/mirrors.txt//dists/trusty/main/binary-amd64/Packages.gz: No such file
apt-mirror: can't open index in proceed_index_gz at /usr/bin/apt-mirror line 460.
````


cbaenziger:
Also, this looks like it will add around 75GB to the size of one's mirror.

PR #116, opened 2015-04-04
Bonding

cbaenziger:
This commit allows one to create a bonded interface for your machines.

To use this code, one provides a `slaves` attribute in their network's definition. One should also provide which interface to use (e.g. `bond0`, `bond1`).

For example:
````
"floating": {
        "interface": "bond0",
        "slaves": ["eth1", "eth2"],
        "netmask": "255.255.0.0",
        "cidr": "192.168.0.0/24",
        "gateway": "192.168.0.1",
        "vip": "192.168.0.6",
        "mtu": "9000"
      }
````

### DIFF
--- a/cookbooks/bcpc/recipes/networking.rb
+++ b/cookbooks/bcpc/recipes/networking.rb
@@ -98,13 +98,15 @@ bash "enable-mellanox" do
     only_if "lspci | grep Mellanox"
 end
 
-bash "enable-bonding" do
+if ["floating", "storage", "management"].select{|i| node[:bcpc][i].attribute?("slaves")}.any?
+  bash "enable-bonding" do
     user "root"
     code <<-EOH
-        modprobe bonding
-        echo 'bonding' >> /etc/modules
+      modprobe bonding
+      echo 'bonding' >> /etc/modules
     EOH
     not_if "grep -e '^bonding' /etc/modules"
+  end
 end
 
 bash "enable-8021q" do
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by cbaenziger** [comment]
Further, if one no longer has a storage network (e.g. your two NICs make a bonded floating network) you can set storage's interface to the same as management's and it will be skipped (see [lines 154-155 in network.rb](https://github.com/cbaenziger/chef-bach/blob/bonding_take_2/cookbooks/bcpc/recipes/networking.rb#L154-L155)).

File: cookbooks/bcpc/recipes/networking.rb [hunk 1]
```diff
@@ -123,105 +131,93 @@
...
+      :mtu => node[:bcpc][iface][:mtu],
+      :metric => i*100
+    )
+  end
+  
+  bash "#{iface} up" do
+    code <<-EOH
+      ifup #{device_name} #{node[:bcpc][iface].has_key("slaves") and node[:bcpc][iface][:slaves].join(" ")}
+    EOH
+    not_if "ip link show up | grep #{device_name}"
```
cbaenziger: One thing to be aware of, if one has a change to an `/etc/network/interfaces.d/*` file, this `ifup` will only run if the device was down. If the device was reconfigured (e.g. new MTU) then that change won't take affect until something external (e.g. `ifdown`/`ifup` or a reboot).

amithkanand:
LGTM

PR #100, opened 2015-03-16
Remove hardcoded cookbook versions

cbaenziger:
- This reverts the change from chef-bcpc [issue 266](https://github.com/bloomberg/chef-bcpc/issues/266) which seems to be resolved in the current apt cookbook.
- This reverts the change from our initial Chef 11 [commit](https://github.com/bloomberg/chef-bach/commit/d6e431e0c18e7d2897028bb59cd3c736397ca580#diff-95d29768b114a017d5a3307f0e57eb09) which hard coded chef-client to version 3.0.6
- This also updates the list of cookbooks in `setup_chef_cookbooks.sh` to have each cookbook on its own line to aide in `git blame` being useful for seeing when things changed per cookbook

PR #117, opened 2015-04-06
Kerberos implementation added

amithkanand:
This PR# adds `Kerberos` to current stack.  After this one should be able to `kerberize` a hadoop cluster.


cbaenziger:
Since this is only a commit for Kerberos can the following be pulled out for an independent commit:
* [`.gitignore`](https://github.com/bloomberg/chef-bach/blob/0fe5e168bb16ff6065ad73aee1c4c84f6fd42366/.gitignore)
* [`Vagrantfile`](https://github.com/bloomberg/chef-bach/blob/0fe5e168bb16ff6065ad73aee1c4c84f6fd42366/Vagrantfile)
* [`vbox/.gitignore`](https://github.com/bloomberg/chef-bach/blob/0fe5e168bb16ff6065ad73aee1c4c84f6fd42366/vbox/.gitignore)
* The cookbook changes which could just be merging PR #100

cbaenziger:
Can you provide something in `[cookbooks/bach_krb5/README.md](https://github.com/bloomberg/chef-bach/blob/87856b67dcc178e47cc42bf2adbc8ab7eb1c0579/cookbooks/bach_krb5/README.md) `?

cbaenziger:
Similarly, all `vagrantfile` changes not necessary for Kerberos should ideally go into their own PR to make this more tractable.

build_bins.sh
```diff
@@ -40,7 +40,7 @@ mkdir -p $APT_REPO_BINS
 apt-get -y update
 
 # Install tools needed for packaging
-apt-get -y install git rubygems make pbuilder python-mock python-configobj python-support cdbs python-all-dev python-stdeb libmysqlclient-dev libldap2-dev ruby-dev gcc patch rake ruby1.9.3 ruby1.9.1-dev python-pip python-setuptools dpkg-dev apt-utils haveged libtool autoconf unzip rsync autogen
+apt-get -y install git rubygems make pbuilder python-mock python-configobj python-support cdbs python-all-dev python-stdeb libmysqlclient-dev libldap2-dev ruby-dev gcc patch rake ruby1.9.3 ruby1.9.1-dev python-pip python-setuptools dpkg-dev apt-utils haveged
```
cbaenziger: Could this code clean-up go into its own PR?

cookbooks/bach_krb5/libraries/utils.rb
```diff
@@ -0,0 +1,18 @@
+def principal_exists?(pname) 
+  require 'rkerberos'
+  kadm5_connected = false
+  princ_found = false
+
+  begin
+    kadm5 = Kerberos::Kadm5.new(:principal => "#{node[:krb5][:admin_principal]}", :password => "#{get_config("krb5-admin-password")}")
+    kadm5_connected = true
+  rescue Kerberos::Kadm5::Exception
```
cbaenziger: Do we not want to bail out if there is a connection error?

cookbooks/bach_krb5/recipes/gems.rb
```diff
@@ -0,0 +1,32 @@
...
+require 'rubygems'
+gem_path = Pathname.new(Gem.ruby).dirname.join("gem").to_s
+
+gem_package "rake-compiler" do
+  gem_binary gem_path
+  version ">=0.0.0"
+  action :nothing
+end.run_action(:install)
+
+node['krb5']['devel']['packages'].each do |pkg|
```
cbaenziger: Is `node['krb5']['devel']['packages']` specified anywhere?

cookbooks/bach_krb5/recipes/gems.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+gem_package "rkerberos" do
+  gem_binary gem_path
+  version ">=0.0.0"
+  action :nothing
+end.run_action(:install)
+
+execute "correct-permissions-for-rkerberos" do
+  command "chmod 644 /opt/chef/embedded/lib/ruby/gems/1.9.1/specifications/rkerberos-0.1.3.gemspec"
```
cbaenziger: Can not this be a [file](https://docs.chef.io/resource_file.html) resource with a `:touch` action?

cookbooks/bach_krb5/recipes/gems.rb
```diff
@@ -0,0 +1,32 @@
...
+  action :nothing
+end.run_action(:run)
+
+execute "correct-permissions-for-rake-compiler" do
+  command "chmod 644 /opt/chef/embedded/lib/ruby/gems/1.9.1/specifications/rake-compiler-0.9.5.gemspec"
```
cbaenziger: Same as above for using a touch action

cookbooks/bach_krb5/recipes/keytabs.rb
```diff
@@ -0,0 +1,42 @@
+get_all_nodes().each do |h|
+  # Create directories to store keytabs
```
cbaenziger: This creates a directory not multiple directories right?

environments/Test-Laptop.json
```diff
@@ -41,6 +42,28 @@
         "hdfs" : {
           "dfs_replication_factor" : 1
         },
+        "realm" : "BCPC.EXAMPLE.COM",
+        "enable_kerberos": false,
+        "kerberos": {
+          "data": {
```
cbaenziger: Is the `keytab` key in these hashes effectively the file name since the hash data structure is really always keyed off the key like `namenode`, `datanode`, etc.? If so, couldn't the `keytab` entry be omitted an generated to make this data structure more simple?

cookbooks/bach_krb5/recipes/keytabs.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    krb5_principal "#{va['principal']}/#{float_host(h[:fqdn])}@#{node[:bcpc][:hadoop][:realm]}" do
+      action :create
+      randkey true
+      not_if {principal_exists?("#{va['principal']}/#{float_host(h[:fqdn])}@#{node[:bcpc][:hadoop][:realm]}")}
+    end
+  end
+
+  # Generate keytabs
+  node[:bcpc][:hadoop][:kerberos][:data].each do |ke, va|
```
cbaenziger: Nit: could `ke` and `va` be more descriptive? Or the relevant data needed out of `va` be stored into a descriptive variable name to make it obvious what is actually at use below without lots of reading. (It is just using the keytab file name right?)

cookbooks/bcpc-hadoop/attributes/default.rb
```diff
@@ -31,7 +31,7 @@
 default["bcpc"]["hadoop"]["namenode"]["jmx"]["port"] = 10111
 default["bcpc"]["hadoop"]["namenode"]["rpc"]["port"] = 8020
 default["bcpc"]["hadoop"]["namenode"]["http"]["port"] = 50070 
-default["bcpc"]["hadoop"]["namenode"]["https"]["port"] = 50080 
+default["bcpc"]["hadoop"]["namenode"]["https"]["port"] = 50470
```
cbaenziger: Shouldn't this port change be done in its own PR?

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -27,8 +27,8 @@
 # to be included in the result. Attribute hierarchy can be expressed as a dot seperated string. User the following
 # as an example
 #
-HOSTNAME_ATTR_SRCH_KEYS = {'hostname' => 'hostname'}
-HOSTNAME_NODENO_ATTR_SRCH_KEYS = {'hostname' => 'hostname', 'node_number' => 'bcpc.node_number'}
+HOSTNAME_ATTR_SRCH_KEYS = {'hostname' => 'fqdn'}
+HOSTNAME_NODENO_ATTR_SRCH_KEYS = {'hostname' => 'fqdn', 'node_number' => 'bcpc.node_number'}
```
cbaenziger: I recall us discussing why this was needed; mind adding why here for posterity?

cookbooks/bach_krb5/recipes/gems.rb
amithkanand: Yes, this parameter is defined in upstream cookbook `krb5` and `bach_krb5` depends on it.

cookbooks/bach_krb5/recipes/keytabs.rb
amithkanand: This will create multiple directories one for each host as keytabs for each host are stored in host specific directory.

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,59 @@
...
+    mode 0755  
+  end
+
+  # Download and create all keytabs
+  node[:bcpc][:hadoop][:kerberos][:data].each do |srvc, srvdat|
+
+    config_host = srvdat['princhost'] == "_HOST" ?  float_host(node[:hostname]) : srvdat['princhost'].split('.')[0]
+    file "#{node[:bcpc][:hadoop][:kerberos][:keytab][:dir]}/#{srvdat['keytab']}" do
+      action :delete
+      only_if {File.exists?("#{node[:bcpc][:hadoop][:kerberos][:keytab][:dir]}/#{srvdat['keytab']}") && node[:bcpc][:hadoop][:kerberos][:keytab][:recreate] == true}
```
cbaenziger: Isn't this unnecessary if the guard on line 99 has the `!File.exsists?` removed and we trust Chef to tell if a file needs updated -- or would there be a time we would want to run with an outdated keytab? (It seems we would very quickly rev a user's credentials cache any ways in the loop starting on line 104.)

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,59 @@
...
+      owner "#{srvdat['owner']}"
+      group "root"
+      mode "#{srvdat['perms']}"
+      content Base64.decode64(get_config("#{config_host}-#{srvc}"))
+      only_if { !File.exists?("#{node[:bcpc][:hadoop][:kerberos][:keytab][:dir]}/#{srvdat['keytab']}") && user_exists?("#{srvdat['owner']}")}
+    end
+  end
+  
+  # Initialize keytbas
+  node[:bcpc][:hadoop][:kerberos][:data].each do |srvc, srvdat|
```
cbaenziger: Why does this loop need to be separate than the one above?

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,59 @@
...
+
+    config_host = srvdat['princhost'] == "_HOST" ? float_host(node[:fqdn]) : srvdat['princhost']
+
+    next if srvdat['principal'] == "HTTP"
+
+    execute "kdestroy-for-#{srvdat['owner']}" do
+      command "sudo -u #{srvdat['owner']} kdestroy"
```
cbaenziger: Did the `user` attribute not do something that `sudo -u` is doing here?

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,59 @@
...
+      action :run
+      only_if { user_exists?("#{srvdat['owner']}") }
+    end
+
+    execute "kinit-for-#{srvdat['owner']}" do
+    command "sudo -u #{srvdat['owner']} kinit -kt #{node[:bcpc][:hadoop][:kerberos][:keytab][:dir]}/#{srvdat['keytab']} #{srvdat['principal']}/#{config_host}"
+      action :run
+      only_if { File.exists?("#{node[:bcpc][:hadoop][:kerberos][:keytab][:dir]}/#{srvdat['keytab']}") && user_exists?("#{srvdat['owner']}") }
```
cbaenziger: It seems this guard's `File.exists?` is only protecting us from Chef delivering a file on line 94 above?

cookbooks/bcpc-hadoop/recipes/journalnode.rb
```diff
@@ -22,12 +22,8 @@
 end
 
 node[:bcpc][:hadoop][:mounts].each do |d|
-# Per chef-documentation for directory resource's recursive attribute:
```
cbaenziger: Why remove this comment?

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
```diff
@@ -60,12 +60,14 @@
...
 end
 
+zks = node[:bcpc][:hadoop][:zookeeper][:servers].map{|zkh| "#{zkh}:#{node[:bcpc][:hadoop][:zookeeper][:port]}"}.join(",")
+
 bash "format-zk-hdfs-ha" do
   code "yes | hdfs zkfc -formatZK"
   action :run
   user "hdfs"
   notifies :restart, "service[generally run hadoop-hdfs-namenode]", :delayed
-  not_if { zk_formatted? }
```
cbaenziger: Yea code clean-up; could you remove the function from `utils.rb` too?

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,59 @@
...
+      mode "#{srvdat['perms']}"
+      content Base64.decode64(get_config("#{config_host}-#{srvc}"))
+      only_if { !File.exists?("#{node[:bcpc][:hadoop][:kerberos][:keytab][:dir]}/#{srvdat['keytab']}") && user_exists?("#{srvdat['owner']}")}
+    end
+  end
+  
+  # Initialize keytbas
+  node[:bcpc][:hadoop][:kerberos][:data].each do |srvc, srvdat|
+
+    config_host = srvdat['princhost'] == "_HOST" ? float_host(node[:fqdn]) : srvdat['princhost']
```
cbaenziger: Do we need to bounce daemons when we update the credentials caches?

cookbooks/bcpc-hadoop/templates/default/hdp_core-site.xml.erb
```diff
@@ -76,4 +79,41 @@
...
+    <name>hadoop.security.authorization</name>
+    <value>true</value>
+    <description>Enable authorization for different protocols.</description>
+  </property>
+
+  <property>
+    <name>hadoop.security.auth_to_local</name>
+    <value>
+      <% node[:bcpc][:hadoop][:kerberos][:data].each do|ke,va| %>
+      RULE:[2:$1@$0](<%= va['principal']%>@.*<%= node[:bcpc][:hadoop][:realm] %>)s/.*/<%=va['owner'] %>/
```
cbaenziger: Should these be attributes or some such. For sites that want to protect cluster one from cluster two `.*<realm>` might be a bit fast and loose?

cookbooks/bcpc-hadoop/templates/default/hdp_hdfs-site.xml.erb
```diff
@@ -154,4 +144,140 @@
...
+  </property>
+  <% else %>
+  <property>
+    <name>dfs.datanode.kerberos.principal</name>
+    <value><%= node[:bcpc][:hadoop][:kerberos][:data][:datanode][:principal] %>/<%=node[:bcpc][:hadoop][:kerberos][:data][:datanode][:princhost] == "_HOST" ? "_HOST" : node[:bcpc][:hadoop][:kerberos][:data][:datanode][:princhost]%>@<%= node[:bcpc][:hadoop][:realm] %></value>
+  </property>
+
+  <property>
+    <name>dfs.datanode.address</name>
+    <value>0.0.0.0:1004</value>
```
cbaenziger: Should not this and htp address be the floating IP?

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -173,7 +163,46 @@
   </property>
 
   <property>
-    <name>yarn.resourcemanager.scheduler.class</name>
+    <name>yarn.resourcemanager.echeduler.class</name>
```
cbaenziger: Typo `echeduler` -> `scheduler`

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -87,16 +87,6 @@
   <% end %>
 
   <property>
-    <name>mapreduce.jobhistory.address</name>
```
cbaenziger: The Job History addresses are removed but not added back anywhere?

roles/BCPC-Hadoop-Head.json
```diff
@@ -15,10 +15,10 @@
     "recipe[bcpc::diamond]",
     "recipe[bcpc::graphite]",
     "recipe[bcpc::powerdns]",
+    "recipe[bach_krb5::krb5_client]",
     "recipe[bcpc-hadoop::disks]",
     "recipe[bcpc-hadoop::hdp_repo]",
     "recipe[bcpc-hadoop::configs]",
-    "recipe[pam::default]",
```
cbaenziger: If we remove `pam::default`, then we will be unable to set ulimits as `pam::default` is what actually writes out the Pam provider modified config files?

cbaenziger:
Any reason to add the descriptions to the *.xml files? I'd prefer we refer to the relevant Hadoop version's files which provide definitions accurate for that version of the software.

cbaenziger:
It's worth commenting, this does not secure ZK data but makes it possible to secure. For example, the following are some of the recommended HBase ACL settings:
````
setAcl /hbase world:anyone:r,sasl:hbase:cdrwa
setAcl /hbase/backupmasters sasl:hbase:cdrwa
setAcl /hbase/draining sasl:hbase:cdrwa
setAcl /hbase/flushtableproc sasl:hbase:cdrwa
setAcl /hbase/hbaseid world:anyone:r,sasl:hbase:cdrwa
setAcl /hbase/master world:anyone:r,sasl:hbase:cdrwa
setAcl /hbase/metaregionserver world:anyone:r,sasl:hbase:cdrwa
setAcl /hbase/namespace sasl:hbase:cdrwa
setAcl /hbase/onlinesnapshot sasl:hbase:cdrwa
setAcl /hbase/regionintransition sasl:hbase:cdrwa
setAcl /hbase/recoveringregions sasl:hbase:cdrwa
setAcl /hbase/replication sasl:hbase:cdrwa
setAcl /hbase/rs sasl:hbase:cdrwa
setAcl /hbase/running sasl:hbase:cdrwa
setAcl /hbase/splitWAL sasl:hbase:cdrwa
setAcl /hbase/table sasl:hbase:cdrwa
setAcl /hbase/tablelock sasl:hbase:cdrwa
setAcl /hbase/tokenauth sasl:hbase:cdrwa
````

File: cookbooks/bach_krb5/recipes/gems.rb (hunk 3 above)
Comment (cbaenziger) [inline cookbooks/bach_krb5/recipes/gems.rb]: Ah! Thanks I missed the lack of bcpc/bach in there.

File: cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb (hunk 21 above)
Comment (amithkanand) [inline cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb]: It has been added to `hdp_mapred-site.xml.erb`.

File: environments/Test-Laptop.json (hunk 7 above)
Comment (amithkanand) [inline environments/Test-Laptop.json]: Consider a scenario where a single `keytab` is generated for  `HDFS` services (`namenode`, `datanode`, `journalnode`) and a single `keytab` for `yarn` services (`resourcemanage` and `nodemanager`), in such situation we need to know which service uses which `keytab` file. If each service is always using a separate `keytab` then using hash structure keys like `namenode`, `datanode` etc can be used to generate `ketyab` file name.

File: cookbooks/bach_krb5/libraries/utils.rb (hunk 2 above)
Comment (amithkanand) [inline cookbooks/bach_krb5/libraries/utils.rb]: Added `Chef::Application.fatal` to log and bail out.

File: cookbooks/bcpc-hadoop/recipes/namenode_master.rb (hunk 16 above)
Comment (amithkanand) [inline cookbooks/bcpc-hadoop/recipes/namenode_master.rb]: Removed from `utils.rb`

File: cookbooks/bcpc-hadoop/recipes/journalnode.rb (hunk 15 above)
Comment (amithkanand) [inline cookbooks/bcpc-hadoop/recipes/journalnode.rb]: Added that comment back.

File: cookbooks/bach_krb5/libraries/utils.rb (hunk 2 above)
Comment (cbaenziger) [inline cookbooks/bach_krb5/libraries/utils.rb]: I was mainly wondering. However, I really would rather us `raise` or if the error message provided by the Kerberos code is good enough, let it raise for us. (See [What Makes A Good Chef Cookbook?](http://www.slideshare.net/JulianDunn/what-makes-a-good-cookbook-35029540) for recommendations on how not to use `Chef::Application.fatal` -- a [not recommended](https://tickets.opscode.com/browse/CHEF-4982) API to use.)

**Comment by cbaenziger** [comment]
I'm a bit uncomfortable without some more documentation around the whole kerberos process in that we can easily forget how some of these knobs work and I suspect anyone not us would not grok this. Any ideas on what can be done for the envisioned process, recipe and attribute use? (An example is my question around the keytabs to filename data structure which I agree with your reasoning on but I could not see that without the explanation.)

File: cookbooks/bach_krb5/recipes/keytabs.rb [hunk 23]
```diff
@@ -0,0 +1,54 @@
+keytab_dir = node[:bcpc][:hadoop][:kerberos][:keytab][:dir]
+realm = node[:bcpc][:hadoop][:realm]
```
cbaenziger: Should not the realm be part of `node[:bcpc][:hadoop][:kerberos]`?

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,69 @@
...
+include_recipe "java::oracle_jce"
 
 %w{zookeeper}.each do |pkg|
   package pkg do
     action :upgrade
   end
 end
+
+# Create Keytabs (if kerberos is eanbled)
+if node[:bcpc][:hadoop][:enable_kerberos] == true then
```
cbaenziger: Should not `enable_kerberos` simply be `node[:bcpc][:hadoop][:kerberos][:enable]`; it seems rather kerberos specific to not be under the kerberos hierarchy?

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,69 @@
...
+                node[:bcpc][:hadoop][:kerberos][:keytab][:recreate] == true
+              }
+    end 
+
+    # Create the keytab file
+    file "#{keytab_dir}/#{keytab_file}" do
+      owner "#{srvdat['owner']}"
+      group "root"
+      mode "#{srvdat['perms']}"
+	  action :create_if_missing
```
cbaenziger: Erm, indentation?

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,69 @@
...
+      content Base64.decode64(get_config("#{config_host}-#{srvc}"))
```
cbaenziger: This should likely use `get_config!` to raise if the keytab is missing?

cookbooks/bach_krb5/libraries/utils.rb
```diff
@@ -0,0 +1,18 @@
+def principal_exists?(pname) 
+  require 'rkerberos'
+  kadm5_connected = false
+  princ_found = false
+
+  begin
+    kadm5 = Kerberos::Kadm5.new(:principal => "#{node[:krb5][:admin_principal]}", :password => "#{get_config("krb5-admin-password")}")
```
cbaenziger: Same point about `get_config!` instead of `get_config` to raise if the password is not found.

cbaenziger:
Where we expect a value from a data bag `get_config!` should be used to raise if the data bag item can not be found, opposed to letting a missing value propagate downstream. E.g. instead of:
````
NoMethodError: undefined method `[]' for nil:NilClass
````

One could see something akin to*:
````
RuntimeError
------------
Failed to find value for <hostname>-namenode!
````

 * Except for the fact that `get_config!` uses `Chef::Application.fatal` instead of `raise`.

File: cookbooks/bach_krb5/recipes/upload_keytabs.rb [hunk 28]
```diff
@@ -0,0 +1,22 @@
+require 'base64'
+# Upload keytabs to chef-server
+get_all_nodes().each do |h|
+  node[:bcpc][:hadoop][:kerberos][:data].each do |srvc, srvdat|
+    # Set host based on configuration 
+    config_host=srvdat['princhost'] == "_HOST" ? float_host(h[:hostname]) : srvdat['princhost'].split('.')[0]
+    keytab_host=srvdat['princhost'] == "_HOST" ? float_host(h[:fqdn]) : srvdat['princhost']
+
+    # Delete existing configuration item (if requested)
+    delete_config("#{config_host}-#{srvdat['principal']}") if node[:bcpc][:hadoop][:kerberos][:keytab][:recreate] == true
```
cbaenziger: This is from the `bcpc` cookbook but `bach_krb5/metadata.rb` does not depend on that cookbook; one will get:
`No resource or method named `delete_config' for `Chef::Recipe "upload_keytabs"'`

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -98,11 +98,18 @@ def get_config(key, item=node.chef_environment, bag="configs")
...
     end
   end
+
+def delete_config(key)
+  init_config if $dbi.nil?
+  $dbi.delete(key) if $dbi.has_key?(key)
+  $dbi.save
+end
+
 end
```
cbaenziger: This ends `get_config()` up above and masks `delete_config()` even after adding `bcpc` as a dependency for `bach_krb5`.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
cbaenziger: Ah, yes already there. Thank you

environments/Test-Laptop.json
```diff
@@ -32,6 +32,7 @@
       "bootstrap": {
         "interface" : "eth0",
         "pxe_interface" : "eth1",
+        "hostname" : "bcpc-bootstrap",
```
http-418: This change was mistakenly omitted from PR #189
Breaks the build with:
```
Bringing machine 'bootstrap' up with 'virtualbox' provider...
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The hostname set for the VM should only contain letters, numbers,
hyphens or dots. It cannot start with a hyphen or dot.
```

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
```diff
@@ -72,12 +72,14 @@
...
+  not_if { znode_exists?("/hadoop-ha/#{node.chef_environment}",zks) }
```
cbaenziger: This won't work; and this will make it appear as though the namenode is unreachable through an amazingly painful failure process (as I've recently found on a few clusters). So, `zks` is of the form:
````
[{"hostname"=>"cluster-r1n02.example.com", "node_number"=>"9"}, {"hostname"=>"cluster-r1n03.example.com", "node_number"=>"10"}, {"hostname"=>"cluster-r1n06.example.com", "node_number"=>"13"}]
````
However, the function `znode_exists?` expects a data structure of the form `
+zks = node[:bcpc][:hadoop][:zookeeper][:servers].map{|zkh| "#{zkh[:hostname]}:#{node[:bcpc][:hadoop][:zookeeper][:port]}"}.join(",")`

File: cookbooks/bcpc-hadoop/recipes/namenode_master.rb [hunk 32]
```diff
@@ -72,12 +72,14 @@
   not_if { node[:bcpc][:hadoop][:mounts].any? { |d| File.exists?("/disk/#{d}/dfs/nn/current/VERSION") } }
 end
 
+zks = node[:bcpc][:hadoop][:zookeeper][:servers].map{|zkh| "#{zkh}:#{node[:bcpc][:hadoop][:zookeeper][:port]}"}.join(",")
```
cbaenziger: I would prefer to see this in the bash code block that's using this. I think it'd still be run in the compile phase though. I'd really like to see this in converge just in case one has to use `chef-shell` to debug this, it will be much easier.

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -62,9 +64,69 @@
...
+      action :create_if_missing
+      content Base64.decode64(get_config!("#{config_host}-#{srvc}"))
+      only_if { user_exists?("#{srvdat['owner']}")  }
+    end
+
+    princ_host = srvdat['princhost'] == "_HOST" ? float_host(node[:fqdn]) : srvdat['princhost']
+
+    next if srvdat['principal'] == "HTTP"
+
+    execute "kdestroy-for-#{srvdat['owner']}" do
```
cbaenziger: Do we need to `kdestroy`? The `kinit` will do a `kdestroy` if necessary as I understand it.

cookbooks/bach_krb5/recipes/gems.rb
```diff
@@ -0,0 +1,33 @@
...
+end
+
+gem_package "rkerberos" do
+  gem_binary gem_path
+  version ">=0.0.0"
+  action :nothing
+end.run_action(:install)
+
+file "correct-permissions-for-rkerberos" do
+  path "/opt/chef/embedded/lib/ruby/gems/1.9.1/specifications/rkerberos-0.1.3.gemspec"
```
cbaenziger: I'm opposed to hardcoding the gemspec paths, please programmatically find these paths.

cookbooks/bcpc-hadoop/recipes/configs.rb
cbaenziger: While `kinit` seems to destroy the CC on the machine; Lev says `kdestroy` could be useful since `kinit` doing a `kdestroy` is not mandatory.

cookbooks/bach_krb5/recipes/krb5_server.rb
```diff
@@ -0,0 +1,26 @@
...
+include_recipe "krb5::kadmin_init"
+
+kdc_resource = resources("service[krb5-kdc]")
+kdc_resource.supports(:status => true, :restart => true, :reload => false)
+kdc_resource.action([:enable, :start])
+kdc_resource.subscribes(:restart, "template #{node['krb5']['data_dir']}/kdc.conf", :delayed) 
+
+kadm_resource = resources("service[krb5-admin-server]")
+kadm_resource.supports(:status => true, :restart => true, :reload => false)
+kadm_resource.action([:enable, :start])
```
cbaenziger: You were mentioning that there is some behavior with `krb5-admin-server` not starting the first time?

cbaenziger:
If we could, clusters deployed with some sort of pre-existing kerberos will not need the `krb5_server` recipe and it could be difficult to scrape off. I am less worried about recipes one may need to add but really do not like including things which one likely does not want in a full install (akin to `powerdns`). Could we remove the `krb5_server` and as such `krb5_client` from the roles?

amithkanand:
I will create another PR for dynamically finding `gemspec` path.

cookbooks/bach_krb5/recipes/gems.rb
amithkanand: This has been implemented as a file resource.

cookbooks/bach_krb5/recipes/gems.rb
amithkanand: This has been added as a file resource.

cookbooks/bach_krb5/recipes/keytabs.rb
amithkanand: Added values to descriptive variable names.  Please see [here](https://github.com/bloomberg/chef-bach/blob/kerberos/cookbooks/bach_krb5/recipes/keytabs.rb#L17-L19).

cookbooks/bcpc-hadoop/attributes/default.rb
amithkanand: This was merged as part of PR #189.

cookbooks/bcpc-hadoop/libraries/utils.rb
amithkanand: This was changed to `FQDN` as per hadoop kerberos implementation documentation.

cookbooks/bcpc-hadoop/recipes/configs.rb
amithkanand: Changes implemented as requested.

cookbooks/bcpc-hadoop/recipes/configs.rb
amithkanand: Moved to single loop.

cookbooks/bcpc-hadoop/recipes/configs.rb
amithkanand: Added user attribute.

cookbooks/bcpc-hadoop/recipes/configs.rb
amithkanand: Changed implemented as requested.

cookbooks/bcpc-hadoop/recipes/configs.rb
amithkanand: No we don't need to bounce daemons. Credential cache is populated to perform `hdfs` operations during `chef` run.

cookbooks/bcpc-hadoop/templates/default/hdp_hdfs-site.xml.erb
amithkanand: Changed to floating IP.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
amithkanand: Corrected typo.

roles/BCPC-Hadoop-Head.json
amithkanand: Added `pam::default` to the runlist. Not sure why it was removed.

cookbooks/bach_krb5/recipes/keytabs.rb
amithkanand: Moved under `node[:bcpc][:hadoop][:kerberos]`.

cookbooks/bcpc-hadoop/recipes/configs.rb
amithkanand: Moved to `node[:bcpc][:hadoop][:kerberos]` and changed the attribute to be `enable`.

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
amithkanand: This is fixed as part of PR #232

cookbooks/bach_krb5/recipes/upload_keytabs.rb
amithkanand: This does depend on `BCPC`. Please see [here](https://github.com/bloomberg/chef-bach/blob/kerberos/cookbooks/bach_krb5/metadata.rb#L9-L11)

cookbooks/bcpc-hadoop/recipes/configs.rb
amithkanand: It is now using `get_config!`.

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
amithkanand: This was fixed under PR #232.

cookbooks/bach_krb5/recipes/gems.rb
amithkanand: As discussed this will be fixed as enhancement PR.

cookbooks/bach_krb5/recipes/krb5_server.rb
amithkanand: For some reason, extending these resources doesn't start the services on first chef-run. I will  have to spend more time on this to debug it. This not a blocker at the moment though.

environments/Test-Laptop.json
amithkanand: This was fixed in PR #202.

cookbooks/bcpc-hadoop/recipes/configs.rb
amithkanand: Fixed indentation.

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
cbaenziger: Committed in Chef-BACH #232

cookbooks/bcpc-hadoop/templates/default/hdp_core-site.xml.erb
cbaenziger: This is just a mapping, for splitting clusters' principals we should do that in the mapping of the environment, e.g. [Test-Laptop](https://github.com/bloomberg/chef-bach/blob/kerberos/environments/Test-Laptop.json#L48-L61).

cbaenziger:
After reams of design and code review I think we have finally beaten in full kerberos support!

cookbooks/bcpc-hadoop/templates/default/hb_hbase-site.xml.erb
```diff
@@ -94,4 +94,71 @@
...
 
+  <% if node[:bcpc][:hadoop][:kerberos][:enable] == true then %>
+  <!-- HBase Authorization Configuration -->
+  <property> 
+    <name>hbase.security.authorization</name> 
+    <value>true</value> 
+  </property> 
+
+  <property>    
+    <name>hbase.superuser</name>
```
DrRayWang: This potentially breaks the pull request and make hbase not working found by Lev. If build this with new cluster, the hbase will not work, and got some error like :  
```
ERROR: org.apache.hadoop.hbase.security.AccessDeniedException: Insufficient permissions for user 'rwang238' (global, action=CREATE)
```

amithkanand:
@DrRayWang: You need to add a user to [this](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/attributes/default.rb#L48) to add it to list of users allowed to perform `HBase` operations. Once added, [`hbase-site.xml`](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/templates/default/hb_hbase-site.xml.erb#L109-L113) will populate the correct value.

Issue #91, opened 2015-02-23
Vagrant baremetal does not regenerate admin.pem


cbaenziger:
Today, there are two issues which can bite one while setting up clusters from a generic bootstrap node image (e.g. using the Vagrant baremetal approach).

If one mis-specifies the `node[:chef_client][:server_url]` attribute (e.g. to another Chef server) one wil get a failure on provisioning at best; or will [reconfigure](https://github.com/bloomberg/chef-bach/commit/1552ee81e6b015dda37fc54d606eabbedfac89c0#diff-5c4334c15ae90f8cf136f0764cf59afbR17) an innocent Chef server. There should be an interlock between `node[:bcpc][:bootstrap][:server]` and `node[:chef_client][:server_url]` to ensure they are the same.

The risk of reconfiguring an innocent Chef server spawns from not re-generating the `/etc/chef-server/admin.pem` and associated key on the Chef server as well.

cbaenziger:
Fixed in #118

PR #118, opened 2015-04-07
Roll the Chef admin user key on baremetal reconfigure

cbaenziger:
Today one using Vagrantfile.baremetal to reconfigure a Vagrant box for baremetal bootstrap may think they are cleaning all cruft leftover but the admin user's PEM file is not today regenerated. This should fix that.


cbaenziger:
This should fix Chef-BACH #91

amithkanand:
LGTM

PR #119, opened 2015-04-07
Disable chef-client certs verification

kiiranh:
Latest chef-client enables server certs verification by default (https://github.com/opscode-cookbooks/chef-client/commit/7b1e1266a5e6f38bd9b040ee473e43b11f715a40)
Disabling this by default in the environment.

This has been tested on local VM cluster and production clusters.


bijugs:
@kiiranh .. LGTM... BTW as an FYI.. In version 12 of Chef the cert verification is set to true by default.

PR #121, opened 2015-04-08
Dynamically set oozie(catalina) jvm opts

kiiranh:
This PR adds an attribute to set catalina memory opts. This is tested and deployed on prod cluster with override in environment to "\"-Xmx2048m -XX:MaxPermSize=256m\"".

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/default.rb
+++ b/cookbooks/bcpc-hadoop/attributes/default.rb
@@ -13,6 +13,7 @@ default["bcpc"]["repos"]["hortonworks"] = 'http://public-repo-1.hortonworks.com/
 default["bcpc"]["repos"]["hdp_utils"] = 'http://public-repo-1.hortonworks.com/HDP-UTILS-1.1.0.20/repos/ubuntu12'
 default["bcpc"]["hadoop"]["disks"] = []
 default["bcpc"]["hadoop"]["oozie"]["admins"] = []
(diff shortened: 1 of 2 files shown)


bijugs:
@kiiranh LGTM. Do we want to add ``-XX:MaxPermSize=256m`` to the new attribute value so that users can be reminded of the additional parameter they can change incase there is a OOM issue.

kiiranh:
@bijugs I've changed the PR to say one can update JVM opts. Users could also specify Xms, MinPermSize etc. I could add MaxPermSize, but will need to set to some default value (hortonworks said it is 64m by default). Let me know if this is required.

bijugs:
@kiiranh The suggestion is purely on usability perspective and not required technically. By adding another parameter it provides a visual clue and reminder to the user that they can add other JVM parameters to that string attribute. With a single parameter as it is currently users may be a bit hesitant.

kiiranh:
Hortonworks said they have the following setting in oozie-env.sh OPTs by default in HDP 2.2:
`-Xmx2048m -XX:MaxPermSize=256m`

kiiranh:
Changed code to set hdp2.2 defaults for oozie.

kiiranh:
I got OOM errors today even with the settings hortonworks suggested. Will update this PR when I can get this fixed.

cbaenziger:
Kiran, mind moving the quotes out of the attribute and instaed into ooz_oozie-env.sh.erb? Otherwise, looks good to me and let's get this off our PR list

kiiranh:
Moved quotes out of attribute to the template and tested on vbox VM cluster. Good to go!

Issue #122, opened 2015-04-10
Increase Flume channel capacity


bijugs:
The default channel capacity for ``Flume`` agents is 100. When entries are written by processes like ``HBase`` or ``HDFS`` at a faster rate, this capacity is not sufficient to store the entries. We need to increase the capacity to higher value for e.g. 10000.

bijugs:
Fixed as part of PR #123. Thanks @kiiranh & @amithkanand .

PR #123, opened 2015-04-10
Set a higher value for Flume channel 

bijugs:
This change is to change the Flume channel capacity to a higher value so that ``flume agents`` can store entries when logs they are monitoring are written at a higher rate. Refer issue #122 .

### DIFF
--- a/cookbooks/bcpc-hadoop/templates/default/flume_flume-conf.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/flume_flume-conf.erb
@@ -1,5 +1,6 @@
 <%= @agent_name %>.channels = memoryChannel
 <%= @agent_name %>.channels.memoryChannel.type = memory
+<%= @agent_name %>.channels.memoryChannel.capacity = 10000
 <%= @agent_name %>.sources = pstream
 <%= @agent_name %>.sources.pstream.channels = memoryChannel
 <%= @agent_name %>.sources.pstream.type = exec


cookbooks/bcpc-hadoop/templates/default/flume_flume-conf.erb
```diff
@@ -1,5 +1,6 @@
 <%= @agent_name %>.channels = memoryChannel
 <%= @agent_name %>.channels.memoryChannel.type = memory
+<%= @agent_name %>.channels.memoryChannel.capacity = 10000
```
kiiranh: How do we decide what this value will be (e.g.: 10000 in this case)? random/recommended/empirical?

bijugs: @kiiranh the memory channel need to have enough capacity to offset the entries which will be added to it in the time delay for the HDFS sink to write data to HDFS. The default value of channel capacity is 100 which is too low when HDFS writes takes some delay but should not take more than few seconds in worst case scenario. If we assume that the worst case scenario is 20 secs and we can expect that the memory channel will have to hold around 2000 entries max. Hence the decision to go with a number of 10000 which will be able to handle a much bigger delay in HDFS write completion.

kiiranh: Ok Sounds good.

bijugs:
@kiiranh . Will you be able to provide a +1.

kiiranh:
+1

amithkanand:
LGTM.

PR #131, opened 2015-04-16
Add an attribute to hannibal hbase_site.xml to point to the hbase.regionserver.info.port

kiiranh:
This PR adds an bcpc-hadoop attribute for specifying hbase region server info port. In HDP2.2, we set a different port(e.g.: 60300) than what the default(60030) is assumed by hannibal.
It also cleans up some unused templates.


kiiranh:
Created PR on wrong branch; should goto bloomberg:hdp2.2. Closing this one.

Issue #124, opened 2015-04-10
Update "Namenode Format UUID file" correctly


kiiranh:
Currently, in bcpc-hadoop::namenode_master recipe [block](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/namenode_master.rb#L99-L109) "namenode_txn_fmt" data bag item is only updated on restarting hadoop-hdfs-namenode service. This block should be changed to re-generate that item if the version file differs from that on disk (maybe using a hashing function) -- this would mean HDFS updates would need to happen on the NN running on the `namenode_master.rb` running machine first.

Also, in bcpc-hadoop::journalnode recipe, [this block](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/journalnode.rb#L25-L31) is redundant. In the same recipe, [this block](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/journalnode.rb#L41-L43) can be made simpler by using chef::bash->cwd to change dir and keeping the code value to just the tar command.
[Resource bash](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/journalnode.rb#L39) name can be changed to include the disk# for readability and debugging.

cbaenziger:
Further, [`namenode_master.rb`](https://github.com/bloomberg/chef-bach/blob/1e66d8709d08842e0027508a750ab581f5ad9ec3/cookbooks/bcpc-hadoop/recipes/namenode_master.rb#L104) uses [`make_config`](https://github.com/bloomberg/chef-bach/blob/d77fb5e28ac11e3949bea012408baf67ae4381ae/cookbooks/bcpc/libraries/utils.rb#L58) which does not overwrite the data bag item -- resulting in a tar file without the `jn/Test-Laptop/current/VERSION` file and only the `nn/current/VERSION` file. (For reference, it appears the asymmetry in path names is due to the JN and NN code, not our recipes.)

cbaenziger:
I would prefer to see us create a `make_config!` which overwrites the data bag if present (it may have to be written as a data bag delete and recreate admittedly).

PR #132, opened 2015-04-17
Update namenode format UUID correctly

amithkanand:
This PR fixes issue #124 Update namenode format UUID correctly.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/namenode_master.rb
+++ b/cookbooks/bcpc-hadoop/recipes/namenode_master.rb
@@ -1,5 +1,8 @@
-include_recipe 'dpkg_autostart'
 require "base64"
+require "digest"
+require 'mixlib/shellout'
+
+include_recipe 'dpkg_autostart'
 include_recipe 'bcpc-hadoop::hadoop_config'
 
 #
@@ -41,6 +44,11 @@ end
 
 template "/etc/init.d/hadoop-hdfs-namenode" do
   source "hdp_hadoop-hdfs-namenode-initd.erb"
+  mode 0655
+end
+
+template "/etc/init.d/hadoop-hdfs-zkfc" do
+  source "hdp_hadoop-hdfs-zkfc-initd.erb"
   mode 0655
 end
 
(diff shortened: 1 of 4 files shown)


cookbooks/bcpc-hadoop/recipes/namenode_master.rb
```diff
@@ -101,16 +109,55 @@
...
+    cmdStrLayVersn = "zgrep -a -i layoutVersion #{Chef::Config[:file_cache_path]}/nn_fmt.tgz|uniq|cut -d'=' -f2"
+
+    cmd = Mixlib::ShellOut.new(cmdStrCount, :timeout => 10).run_command
+    cmd.error!
+    Chef::Log.debug("Total number of version lines : #{cmd.stdout}") 
+    if cmd.stdout.to_i != 2
+      Chef::Log.fatal("Couldn't find required number of layoutVersion records");
+    end
+
+    cmd = Mixlib::ShellOut.new(cmdStrUnqCount, :timeout => 10).run_command
```
cbaenziger: I'm not fully sure we want this check. In steady-state I would expect this -- and we see this. However, during an HDFS upgrade I am not sure what the state of each NN and JN's version would be.

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
```diff
@@ -101,16 +109,55 @@
...
+  block do
+    cmdStrCount = "zgrep -a -i layoutVersion #{Chef::Config[:file_cache_path]}/nn_fmt.tgz|wc -l"
+    cmdStrUnqCount = "zgrep -a -i layoutVersion #{Chef::Config[:file_cache_path]}/nn_fmt.tgz|uniq|wc -l"
+    cmdStrLayVersn = "zgrep -a -i layoutVersion #{Chef::Config[:file_cache_path]}/nn_fmt.tgz|uniq|cut -d'=' -f2"
+
+    cmd = Mixlib::ShellOut.new(cmdStrCount, :timeout => 10).run_command
+    cmd.error!
+    Chef::Log.debug("Total number of version lines : #{cmd.stdout}") 
+    if cmd.stdout.to_i != 2
+      Chef::Log.fatal("Couldn't find required number of layoutVersion records");
```
cbaenziger: After every `Chef::Log.fatal` we should `raise` to stop processing this block; perhaps set the resource attribute `ignore_failure` so we continue on but at least the problem will be logged.

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
```diff
@@ -101,16 +109,55 @@
...
+    cmd Mixlib::ShellOut.new(cmdStrLayVersn, :timeout => 10).run_command
+    cmd.error!
+    Chef::Log.debug("layoutVersion stored in node is : #{node_layout_version}")
+    Chef::Log.debug("layoutVersion stored in the file is #{cmd.stdout.to_i}")
+
+    if ( get_config("namenode_txn_fmt").nil? ) || ( cmd.stdout.to_i < node_layout_version )
+      make_config!("namenode_txn_fmt", Base64.encode64(IO.read("#{Chef::Config[:file_cache_path]}/nn_fmt.tgz")));
+      node[:bcpc][:hadoop][:hdfs][:layoutVersion] = cmd.stdout.to_i
+      node.save
+    end
```
cbaenziger: We should likely log if the version increases for some reason.

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
cbaenziger: Strange this was to the old diff... Thanks for adding the `raise` in.

cookbooks/bcpc-hadoop/recipes/namenode_master.rb
```diff
@@ -101,16 +109,61 @@
...
+    cmd.error!
+    Chef::Log.debug("layoutVersion stored in node is : #{node_layout_version}")
+    Chef::Log.debug("layoutVersion stored in the file is #{cmd.stdout.to_i}")
+
+    if ( get_config("namenode_txn_fmt").nil? ) || ( cmd.stdout.to_i < node_layout_version )
+      make_config!("namenode_txn_fmt", Base64.encode64(IO.read("#{Chef::Config[:file_cache_path]}/nn_fmt.tgz")));
+      node.set[:bcpc][:hadoop][:hdfs][:layoutVersion] = cmd.stdout.to_i
+      node.save
+    elsif cmd.stdout.to_i > node_layout_version
+      Chef::Log.fatal("New HDFS layoutVersion is higher than old HDFS layoutVersion")
```
cbaenziger: Just to document it; the layout version appears to be an increasingly negative number so higher versions are older layouts.

PR #133, opened 2015-04-17
Dynamically set hannibal hbase.regionserver.info.port

kiiranh:
This PR provides a way to set hbase region server info port for hannibal. Default port in hannibal is 60030.
It also cleans up some unused templates.


kiiranh:
Tested on my vbox VM cluster.

cbaenziger:
Please split the mass-deSQLing into its own commit.

kiiranh:
Done

PR #134, opened 2015-04-17
Override hannibal hbase parameters to support HDP 2.2

kiiranh:
This fixes issue #130. The PR updates hbase version from 0.96 to 0.98 and overrides the hbase.regionserver.info.port to 60300 which we default to in HDP 2.2.

This PR depends on PR #133.

PR #138, opened 2015-04-22
HDP2.2 code base added

amithkanand:
This PR adds HDP2.2 code base to Chef-Bach.


kiiranh:
PR #133 #134 will be needed for hannibal to work correctly with HDP 2.2. #133 is a change in master and #134  in hdp2.2 branch.

cbaenziger:
Looks good to me on a read through. I do, however, agree with Kiran we should merge #133 and re-tag.

PR #141, opened 2015-04-23
Spark cookbook added

amithkanand:
This PR adds ``spark`` installation to the stack.

### DIFF
--- a/cookbooks/bach_spark/recipes/spark_package.rb
+++ b/cookbooks/bach_spark/recipes/spark_package.rb
@@ -1,0 +1,47 @@
+spark_tar_file = "#{node[:spark][:download][:file][:name]}.#{node[:spark][:download][:file][:type]}"
+spark_file_name = node[:spark][:download][:file][:name]
+spark_tar_url = node[:spark][:download][:url]
+spark_download_dir = node[:spark][:download][:dir]
+spark_install_dir = node[:spark][:package][:base]
+spark_pkg_prefix = node[:spark][:package][:prefix]
+spark_pkg_version = node[:spark][:package][:version]
+spark_extract_dir = "/tmp"
+
+gem_package "fpm" do
+   action :install
+end
+
+remote_file "#{spark_download_dir}/#{spark_tar_file}" do
+  source "#{spark_tar_url}/#{spark_tar_file}"
+  not_if { File.exists?("#{spark_download_dir}/#{spark_tar_file}") || File.exists?("#{spark_download_dir}/#{spark_pkg_prefix}_#{spark_pkg_version}_amd64.deb") }
+end
+
+execute "extract spark tar" do
+  command "tar xvf #{spark_tar_file} -C #{spark_extract_dir}"
+  cwd spark_download_dir
+end
+
+bash "build_spark_package" do
+  cwd "#{spark_extract_dir}/#{spark_file_name}"
+  user 'root'
+  group 'root'
+  code %Q{
+    fpm -s dir -t deb --prefix #{spark_install_dir}/#{spark_pkg_version} -n #{spark_pkg_prefix} -v #{spark_pkg_version} --description "Spark Package with Hadoop" -p #{spark_download_dir} *
+  }
+  umask 0002
+  not_if { File.exists?("#{spark_download_dir}/#{spark_pkg_prefix}_#{spark_pkg_version}_amd64.deb") }
+  notifies :run, "bash[build_bins]", :delayed
+end
+
+# Call build_bins.sh so the package is added to the apt-repo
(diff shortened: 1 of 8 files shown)


cbaenziger:
Could you also add this to the worker run_list? I would ideally like us to ship this as a package (e.g. built with [fpm](https://github.com/jordansissel/fpm/wiki#overview)) as we do Hannibal, etc.

amithkanand:
Added `bach_spark` to `BCPC-Hadoop-Worker` role.

roles/BCPC-Hadoop-Worker.json
```diff
@@ -20,7 +20,8 @@
     "recipe[bcpc-hadoop::oozie_client]",
     "recipe[bcpc::zabbix-work]",
     "recipe[bcpc::diamond]",
-    "recipe[bcpc_jmxtrans]"
+    "recipe[bcpc_jmxtrans]",
+    "recipe[bach_spark::default]"
```
cbaenziger: Perhaps install Spark with the other high level languages (like Hive and Pig)?

cbaenziger:
Please FPM'ize the tarball.

cookbooks/bach_spark/templates/default/spark-env.sh.erb
```diff
@@ -0,0 +1,5 @@
+export HADOOP_CONF_DIR="/etc/hadoop/conf"
+export HDP20_BASE_DIR="/usr/lib/hadoop"
```
cbaenziger: No need to support HDP2.0 we have moved on to the HDP2.2 milestone for this codebase.

cookbooks/bach_spark/recipes/spark_package.rb
```diff
@@ -0,0 +1,36 @@
+spark_tar_file = node[:spark][:download][:file]
+spark_tar_url = node[:spark][:download][:url]
+spark_download_dir = node[:spark][:download][:dir]
+spark_install_dir = node[:spark][:package][:base]
+spark_pkg_prefix = node[:spark][:package][:prefix]
+spark_pkg_version = node[:spark][:package][:version]
+
+gem_package "fpm" do
+   action :nothing
+end.run_action(:install)
```
cbaenziger: You should not need to `run_action` here since Chef is not needing to load the Gem, this will install before the `bash` action [below](https://github.com/bloomberg/chef-bach/pull/141/files#diff-4f690f0cb3248f506a62d866e231e0eeR17)?

cookbooks/bach_spark/recipes/spark_package.rb
```diff
@@ -0,0 +1,36 @@
...
+  source "#{spark_tar_url}/#{spark_tar_file}"
+  not_if { File.exists?("#{spark_download_dir}/#{spark_tar_file}") }
+end
+
+bash "build_spark_package" do
+  cwd spark_download_dir
+  user 'root'
+  group 'root'
+  code %Q{
+    fpm -s tar -t deb --prefix #{spark_install_dir} -n #{spark_pkg_prefix} -v #{spark_pkg_version} --description "Spark Package with Hadoop 2.6" #{spark_tar_file}
```
cbaenziger: I would not mention Hadoop 2.6 since we have nothing which is 2.6 dependent and this will likely become stale.

cookbooks/bach_spark/templates/default/spark-env.sh.erb
```diff
@@ -0,0 +1,3 @@
+export HADOOP_CONF_DIR="/etc/hadoop/conf"
+export HDP22_BASE_DIR="/usr/hdp/current/hadoop-client"
```
cbaenziger: The dependency on the hadoop basedir has been frustrating me. Perhaps follow [Using Spark's "Hadoop Free" Build](http://spark.apache.org/docs/latest/hadoop-provided.html) instructions? (Particularly, `hadoop classpath`.)

cookbooks/bach_spark/recipes/spark_package.rb
```diff
@@ -0,0 +1,36 @@
...
+  }
+  umask 0002
+  not_if { File.exists?("#{spark_pkg_prefix}_#{spark_pkg_version}_amd65.deb") }
```
cbaenziger: `amd65`?

roles/BCPC-Hadoop-Worker.json
amithkanand: Hive and PIG are installed separately. Hive is role and PIG installs as a separate recipe. We can introduce a new role for Hadoop Eco System components.

cbaenziger:
This looks good

Issue #82, opened 2015-02-10
Generic bootstrap does not regenerate UFW rules to allow TFTP access


cbaenziger:
If one setups a VM bootstrap image and then uses `Vagrant.baremetal` to re-provision the UFW rules are not initially updated. The rules from `ufw.rb` do not update on the re-provision but seem to update subsequently. Disabling the `ufw` service and re-chefing the bootstrap seems to add the new IP in for the TFTP rule (though the old IP is left in as well).

If one sees TFTP boot failing, simply check `iptables -L` under the `ufw-user-input` chain for TFTP and ensure the IP address listed is for the correct machine.

amithkanand:
This is fixed in PR #143

Issue #39, opened 2014-11-28
Add dfs.balance.bandwidthPerSec parameter to hdfs-site.xml


amithkanand:
We have experienced issues w.r.t to how HDFS data is stored on different nodes. We have seen that some of the nodes are consuming more disk space causing disk usage to go upto 100% where other nodes stay at very low disk usage of around 10%. To re-balance the data across all the nodes we can use `hdfs balancer` and to use that we must specify network bandwidth that can be used by balancer utility. I recommend adding `dfs.balance.bandwidthPerSec` parameter to `hdfs-site.xml` to control the network bandwidth while re-balancing the cluster.

amithkanand:
This is fixed in PR #143

Issue #140, opened 2015-04-23
Removing zabbix-work recipe 


bijugs:
This is related to [chef-bcpc issue #299](https://github.com/bloomberg/chef-bcpc/issues/299).

The ``zabbix-head`` recipe seem to include [zabbix-work](
https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc/recipes/zabbix-head.rb#L139) recipe while we also include both the recipes in the [BCPC-Hadoop-Head role](https://github.com/bloomberg/chef-bach/blob/master/roles/BCPC-Hadoop-Head.json#L12-L13). Can we remove the include in ``zabbix-head`` recipe so that we can avoid installing ``zabbix-agent`` by removing the ``zabbix-work`` recipe from all the roles.

amithkanand:
This is fixed in PR #143

amithkanand:
Removal of `zabbix-agent` removes `zabbix-sender` and that breaks sending data from `graphite` to `zabbix`. Since we have been using building `Zabbix` using source, we should probably move to using [`zabbix debian`](http://repo.zabbix.com/zabbix/2.2/ubuntu/pool/main/z/zabbix/) repository to manage the installation of different packages in a better way.

bijugs:
As per previous comment closing this issue.

PR #143, opened 2015-04-23
Generic issues

amithkanand:
This PR takes care of issues given below:
Issue #82 
Issue #39 
Issue #154 
Issue #153 
Issue #152 
Issue #155 
Issue #156 

This PR also removes all `BCPC` specific roles.


cbaenziger:
I think you will want to pull the zabbix-agent out of [`build_bins.sh`](https://github.com/amithkanand/chef-bach/blob/739e7f0234098bf63006cbb2151f0abda7443c70/build_bins.sh#L172-L192).

cookbooks/bcpc-hadoop/templates/default/hdp_hdfs-site.xml.erb
```diff
@@ -15,6 +15,11 @@
   </property>
 
   <property>
+    <name>dfs.datanode.balance.bandwidthPerSec</name>
```
cbaenziger: Could this have a node attribute too? (100MB is a drop in the bucket for nodes with 20 and 40GB throughput.)

amithkanand: Yes, can be converted into a node attribute. I need to figure out a way to get max speed for network card to be able to calculate it based on a ratio.

cbaenziger: It looks like `ethtool` supports general hardware but `mii-tool` does not support all the NICs I could find.

cbaenziger:
@amithkanand Awaiting at least manual configuration of `dfs.datanode.balance.bandwidthPerSec` to an arbitrary value. I'm happy to just have an attribute, no need to make it a ratio of interface speed.

cbaenziger:
The zabbix-agent removal also needs to strip references from:
* [keepalived-vip_change.erb](https://github.com/bloomberg/chef-bach/blob/be32e86ca5e961824c8b094e694bffc804a84103/cookbooks/bcpc/templates/default/keepalived-vip_change.erb#L19-L20)
* [zabbix-head.rb](https://github.com/bloomberg/chef-bach/blob/8afefe5dfab92333b7611005889f1d7f903f7d78/cookbooks/bcpc/recipes/zabbix-head.rb#L169-L175)

cbaenziger:
Due to [graphite_to_zabbix.rb](https://github.com/bloomberg/chef-bach/blob/a278f13657b5caa4d6507ddf93036abfd2aa6657/cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb) we need zabbix_sender on head nodes. Thanks for dropping #140 from this for now.

PR #147, opened 2015-04-24
Fir journal node directory permissions

kiiranh:
Under #132, `directory "/disk/#{d}/dfs/jn/" do` was removed thinking it was redundant. But because of that the directory was created with owner as root and journal node service failed to come up. This PR adds the directory resource back.

Tested on vbox vm cluster. Chef completes successfully.


cbaenziger:
LGTM

PR #158, opened 2015-05-21
chef vault cobbler root and web password

DrRayWang:
* encrypt cobbler root/web password with chef vault into databag "os"
* add facility methods in bcpc library to create a data bag if missing and access the secret in chef vault

### DIFF
--- a/cluster-assign-roles.sh
+++ b/cluster-assign-roles.sh
@@ -34,7 +34,7 @@ KNIFE_ADMIN="-u admin -k /etc/chef-server/admin.pem"
 # Argument: $1 - a string of role!IP!FQDN pairs separated by white space
 # Will install the machine with role $role in the order passed (left to right)
 function install_machines {
-  passwd=`sudo knife data bag show configs $ENVIRONMENT $KNIFE_ADMIN | grep "cobbler-root-password:" | awk ' {print $2}'`
+  passwd=`sudo knife vault show os cobbler "root-password" --mode client | grep "root-password:" | awk ' {print $2}'`
   for h in $(sort <<< ${*// /\\n}); do
     [[ "$h" =~ $REGEX ]]
     local run_list="${BASH_REMATCH[1]}"
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -22,6 +22,17 @@ require 'base64'
 require 'thread'
 require 'ipaddr'
 
+# create databag if missing
+def create_databag(name)
+  if !Chef::DataBag.list.key?(name)
+    bag = Chef::DataBag.new
+    bag.name(name)
+    bag.create
+  end
+end
+
+# retrieve secret by key in chef vault item from chef vault data bag
+
 #
 # Constant string which defines the default attributes which need to be retrieved from node objects
 # The format is hash { key => value , key => value }
@@ -67,11 +78,26 @@ def make_config(key, value)
 	end
 end
 
-def get_config(key)
-	init_config if $dbi.nil?
-	Chef::Log.info  "------------ Fetching value for key \"#{key}\""
-	value = node['bcpc']['encrypt_data_bag'] ? $edbi[key] : $dbi[key]
-	return value
+# get value for data bag/chef-vault item with key
+# bag: databag name of chef vault item; 
+# item: chef vault item; 
+# key: the key to retrieve password in the item
+def get_config(key, item=node.chef_environment, bag="configs")
+  require 'chef-vault'
+  begin
+    entry = ChefVault::Item.load(bag, item)
+    return entry[key]
+  rescue ChefVault::Exceptions::KeysNotFound
+    if bag != "configs"
+      entry = Chef::DataBagItem.load(bag,item)
+      return entry[key]
+    else
+      init_config if $dbi.nil?
+      Chef::Log.info  "------------ Fetching value for key \"#{key}\""
+      value = node['bcpc']['encrypt_data_bag'] ? $edbi[key] : $dbi[key]
+      return value
+    end
+  end
 end
 
 def get_config!(key)
--- a/cookbooks/bcpc/recipes/cobbler.rb
+++ b/cookbooks/bcpc/recipes/cobbler.rb
@@ -18,13 +18,32 @@
 #
 
 require 'digest/sha2'
+require 'chef-vault'
 
 make_config('cobbler-web-user', "cobbler")
-make_config('cobbler-web-password', secure_password)
-make_config('cobbler-root-password', secure_password)
-make_config('cobbler-root-password-salted', "#{get_config('cobbler-root-password')}".crypt("$6$" + rand(36**8).to_s(36)) )
+
+create_databag("os")
+
+# backward compatibility
+web_passwd = get_config('cobbler-web-password')
+if web_passwd.nil?
+  web_passwd = secure_password
+end
+root_passwd = get_config('cobbler-root-password')
+if root_passwd.nil?
+  root_passwd = secure_password
+end
+
+chef_vault_secret "cobbler" do
+  data_bag 'os'
+  raw_data({ 'web-password' => web_passwd, 'root-password' => root_passwd, 'root-password-salted' => root_passwd.crypt("$6$" + rand(36**8).to_s(36)) })
+  admins node[:hostname]
+  search '*:*'
+  action :nothing
+end.run_action(:create_if_missing)
+
 node.default[:cobbler][:web_username] = get_config('cobbler-web-user')
-node.default[:cobbler][:web_password] = get_config('cobbler-web-password')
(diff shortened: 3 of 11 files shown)


amithkanand:
For vault changes following file require a change as well:
* [cobbler.bcpc_ubuntu_host.preseed.erb](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb#L58)
*

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -22,6 +22,22 @@
...
+  if !Chef::DataBag.list.key?(name)
+    bag = Chef::DataBag.new
+    bag.name(name)
+    bag.create
+  end
+end
+
+# load chef vault item from vault bag
+def get_secret(bag,item)
+require 'chef-vault'
```
kiiranh: Indentation: Should there be a 'tab' before require 'chef-vault'?

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -22,6 +22,22 @@
...
+  entry = ChefVault::Item.load(bag, item)
```
kiiranh: What happens if there is no entry in the bag by the name 'item'?

cookbooks/bcpc/recipes/cobbler.rb
```diff
@@ -20,11 +20,21 @@
...
+chef_vault_secret "cobbler" do
+  data_bag 'os'
+  raw_data({ 'web-password' => secure_password, 'root-password' => secure_password })
+  admins node[:hostname]
+  search '*:*'
+  action :nothing
+end.run_action(:create_if_missing)
+
+cobbler = get_secret("os","cobbler")
+make_config('cobbler-root-password-salted', "#{cobbler['root-password']}".crypt("$6$" + rand(36**8).to_s(36)) )
```
kiiranh: What if cobbler is nil?

cookbooks/bcpc/libraries/utils.rb
DrRayWang: Chef API will handle that, please check https://github.com/Nordstrom/chef-vault/blob/master/lib/chef-vault/item_keys.rb#L114-L127

cookbooks/bcpc/recipes/cobbler.rb
DrRayWang: If cobbler is nil, the script will fail at the next step : make_config(), and whole chef run should exit by this failure

cookbooks/bcpc/libraries/utils.rb
DrRayWang: Fixed, thanks

pu239ppy:
Looks sane and straightforward, however I would use a single function to retrieve an encrypted or non-encrypted item.  

get_config will accept three parameters a key an item and a data bag, if only the key is provided the default will be searched, if data bag and item are provided a hash will be returned and a single value will be provided if all three parameters are identified and hopefully match something

You can use Chef::DataBag and Chef::DataBagItem and ChefVault to load all data bags and items "pointers" into a single "namespace" in init_config including the node.chef_environment.  As you are loading you can use the Vault library depending on weather the data bag is an instance of a Vault or a regular data bag.  Also per discussion make sure to have a "default" key pointing to configs["chef_environment"] identifying that this will the default search space if an unqualified item is requested.  

This will allow you to preserve the backward compatibility for recipe developers, ensure that no one has to modify existing recipers,  and at the same time support multi item/multi databag configuration

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -67,11 +78,26 @@ def make_config(key, value)
...
+# bag: databag name of chef vault item; 
+# item: chef vault item; 
+# key: the key to retrieve password in the item
+def get_config(bag="configs",item=node.chef_environment,key)
+  require 'chef-vault'
+  begin
+    entry = ChefVault::Item.load(bag, item)
+    return entry[key]
+  rescue ChefVault::Exceptions::KeysNotFound
+    if bag != "configs"
```
pu239ppy: This can be used for things in config as well

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -67,11 +78,26 @@ def make_config(key, value)
...
+      entry = Chef::DataBagItem.load(bag,item)
+      return entry[key]
+    else
```
pu239ppy: This is obsoleted we can just get_config from any (bag,item,key) however it does have a hook for encrypted databage and implies that init.  And even that can be done in second case so the flow would be

take out if bag != "configs"
check if encrypted databag from node['bcpc']['encrypt_data_bag'] use Chef::EncryptedDataBagItem.load('configs', node.chef_environment) otherwise use Chef::DataBagItem.load('configs', node.chef_environment)

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -67,11 +78,26 @@ def make_config(key, value)
...
-def get_config(key)
-	init_config if $dbi.nil?
-	Chef::Log.info  "------------ Fetching value for key \"#{key}\""
-	value = node['bcpc']['encrypt_data_bag'] ? $edbi[key] : $dbi[key]
-	return value
+# get value for data bag/chef-vault item with key
+# bag: databag name of chef vault item; 
+# item: chef vault item; 
+# key: the key to retrieve password in the item
+def get_config(bag="configs",item=node.chef_environment,key)
```
amithkanand: Shouldn't the order for parameters be (key, item, bag)? A `bag` can have multiple `items` and an `item` can have multiple `keys`. Consider a scenario where default bag `configs` has two items and I want to fetch a particular `key` from a particular `item` leaving the data bag to default value. The call the get_config method will look something like:
````
get_config(item,key)
````

With current method signature the call will change the bag to the item name passed instead of changing the item parameter. Refer to the code below:

````
def get_config(bag="configs",item="chef_environment",key)
  puts "Bag is : #{bag}"
  puts "Item is : #{item}"
  puts "Key is : #{key}"
end
get_config("myitem","test")
````
The output from above I get is:
````
Bag is : myitem
Item is : chef_environment
Key is : test
````

**Comment by pu239ppy** [comment]
The only thing here that maybe no negotiable is that init creates  bags as well, however I see little value in creating a bag with an empty item it it never existed in the first place.  Recipes will look there for values and fails when those values are not there.

File: cookbooks/bcpc/libraries/utils.rb (hunk 4 above)
Comment (amithkanand) [inline cookbooks/bcpc/libraries/utils.rb]: In the `rescue` block we should only keep following two lines:
````
entry = Chef::DataBagItem.load(bag,item)
return entry[key]
````

Everything else can go away.

File: cookbooks/bcpc/libraries/utils.rb [hunk 7]
```diff
@@ -22,6 +22,17 @@
 require 'thread'
 require 'ipaddr'
 
+# create databag if missing
+def create_databag(name)
```
amithkanand: I don't thin we need `create_databag` method. Also `init` method is not required anymore. The signature for `make_config` should change to `(key, item, bag)`  and `item` and `bag` should default to `node.chef_environment` and `configs` respectively. The `make_config` method should handle the logic for creation of `data bags` if it doesn't exist.

pu239ppy:
Ok to merge

cbaenziger:
I am happy to see this taking shape! One issue is that this can not build a cluster, as it stands today. Various shell-scripts rely on `cobbler-root-password` and a cluster build fails as such. Please recommend this to merge when one can build a cluster from the code.

cluster-assign-roles.sh
```diff
@@ -34,7 +34,7 @@ KNIFE_ADMIN="-u admin -k /etc/chef-server/admin.pem"
 # Argument: $1 - a string of role!IP!FQDN pairs separated by white space
 # Will install the machine with role $role in the order passed (left to right)
 function install_machines {
-  passwd=`sudo knife data bag show configs $ENVIRONMENT $KNIFE_ADMIN | grep "cobbler-root-password:" | awk ' {print $2}'`
+  passwd=`sudo knife vault show os cobbler "root-password" --mode client | grep "root-password:" | awk ' {print $2}'`
```
amithkanand: Although this works and not a big deal, the syntax can be changed to:
`sudo knife vault show os cobbler --mode client | grep "root-password`

DrRayWang: sudo knife vault show os cobbler "root-password" --mode client
because this will return something like:
id:            cobbler
root-password: ************

So we still need grep "root-password" to strip id item out.

amithkanand:
I was able to build a cluster using this commit. LGTM.

DrRayWang:
Thanks, Amit.

cbaenziger:
BREAKING: If running this on a pre-existing cluster one needs to ensure the `chef-vault` cookbook is available

PR #159, opened 2015-05-22
fix balancer bandwidth typo

DrRayWang:
* fixed our bcpc-hadoop attribute syntax error issue for `default["bcpc"]["hadoop"]["balancer"]["bandwidth]" = 1048576`


cbaenziger:
Please split into two commits

PR #166, opened 2015-05-29
Update cluster-assign-roles to error if a worker fails to chef

cbaenziger:
This fixes the issue where if a worker node fails to Chef then `cluster-assign-roles.sh` and particularly `tests/automated_install.sh` will think all is fine and well. If a machine fails to Chef one will get output akin to:
````
Install failed for machines:
bcpc-vm3.bcpc.example.com
````
Further, `cluster-assign-roles.sh` will return non-zero (specifically 1) to signal the failure.

### DIFF
--- a/cluster-assign-roles.sh
+++ b/cluster-assign-roles.sh
@@ -286,8 +286,8 @@ INSTALL_TYPE=$2
 MATCHKEY=${3-}
 
 shopt -s nocasematch
-if [[ ! "$INSTALL_TYPE" =~ (openstack|hadoop|kafka) ]]; then
-  printf "Error: Need install type of OpenStack, Hadoop or Kafka\n" > /dev/stderr
+if [[ ! "$INSTALL_TYPE" =~ (openstack|hadoop|kafka|basic) ]]; then
+  printf "Error: Need install type of OpenStack, Hadoop, Kafka or Basic\n" > /dev/stderr
   exit 1
 fi
 shopt -u nocasematch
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cluster-assign-roles.sh [hunk 1]
```diff
@@ -181,10 +181,24 @@ function hadoop_install {
   done
 
   printf "Installing workers...\n"
+  status_file=${TMPDIR:-/tmp}/status_file_$$
+  function clean_up_status_file {
+    rm -f $status_file
+  }
+  trap clean_up_status_file EXIT
+  tempfile -n $status_file
```
kiiranh: `tempfile` command is not recognized by the following bash versions:
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin14)
GNU bash, version 3.2.25(1)-release (x86_64-redhat-linux-gnu)

On, both versions, `mktemp $status_file` works.

cluster-assign-roles.sh
```diff
@@ -181,10 +181,24 @@ function hadoop_install {
...
+    install_machines $m || ( [[ "$m" =~ $REGEX ]]; local fqdn="${BASH_REMATCH[3]}"; echo "$fqdn" >> $status_file ) &
   done
   wait
+  failures=$(wc -l $status_file | sed 's/ .*//')
+  if [[ $failures -ne 0 ]]; then
+    printf "Install failed for machines:\n$(cat $status_file)"
+    rm $status_file
+    exit 1
+  fi
+  rm -f $status_file
```
kiiranh: Call function `clean_up_status_file ` instead in case you wish to do more cleanup in the future?

cluster-assign-roles.sh
kiiranh: On the Ubuntu bootstrap node, with GNU bash, version 4.2.25(1)-release (x86_64-pc-linux-gnu), tempfile is available and the code does work.

cbaenziger:
This also now includes a `Basic` cluster option for Chefing a cluster to create all node objects necessary for Chef vaulting or to create a node object for a replacement machine when re-installing machines.

amithkanand:
LGTM.

Issue #169, opened 2015-06-03
Run chef-client daemon with custom client.rb


kiiranh:
One can have a node chef'd periodically by starting a chef-client process on the node (include `recipe[chef-client::init_service]` in node's runlist). All this works fine when the node's client key is present in `/etc/chef/` directory; for e.g.; `/etc/chef/client.pem`. But if the node's key is not present in that location and you use a [custom client.rb script](https://github.com/bloomberg/chef-bach/blob/master/setup_chef_cookbooks.sh#L25-L56) as with [chef-bach's bootstrap node](https://github.com/bloomberg/chef-bach/blob/master/setup_chef_bootstrap_node.sh#L32), chef-client fails because it cannot find node's key in /etc/chef/ directory.
This only affects the bootstrap node. There should be a way to have chef-client use the [bootstrap nodes' key](https://github.com/bloomberg/chef-bach/blob/master/setup_chef_cookbooks.sh#L35) at a different location than /etc/chef/...

cbaenziger:
Merged

PR #170, opened 2015-06-03
Enable chef-client to run with custom client.rb config on bootstrap node

kiiranh:
This fixes issue #169. 

The pull request modifies the BCPC-Bootstrap role from json to ruby role format. It maintains the same run_list, but adds override_attrbiutes list to setup a custom `/etc/chef/client.rb` through the [chef-client cookbook](https://github.com/opscode-cookbooks/chef-client/blob/master/attributes/default.rb#L24-L32) using the current [knife.rb](https://github.com/bloomberg/chef-bach/blob/master/setup_chef_cookbooks.sh#L26-L56)


kiiranh:
Unit tested this and chef-client runs happily on the bootstrap node. Need to test this by setting up a cluster from scratch.

kiiranh:
Closing this PR since, this approach would required the BCPC-Bootstrap role to have access to the bootstrap node IP and http(s)_proxy variables which are not easily accessible to it.

PR #172, opened 2015-06-05
Run chef-client with custom client.rb

kiiranh:
This fixes issue #169.

The PR updates the version of chef-client pinned in [setup_chef_cookbooks.sh](https://github.com/bloomberg/chef-bach/blob/master/setup_chef_cookbooks.sh#L60). It also creates a symlink in /etc/chef/client.d/ to the [custom knife.rb](https://github.com/bloomberg/chef-bach/blob/master/setup_chef_cookbooks.sh#L26-L56) we use on the bootstrap node.

### DIFF
--- a/setup_chef_bootstrap_node.sh
+++ b/setup_chef_bootstrap_node.sh
@@ -30,3 +30,8 @@ fi
 
 knife node run_list add $(hostname -f) 'role[BCPC-Bootstrap]' -c .chef/knife.rb
 sudo chef-client -c .chef/knife.rb
+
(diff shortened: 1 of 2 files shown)


kiiranh:
This has been tested on a vbox vm cluster and chef-client is running happily on the bootstrap node.

cbaenziger:
This look good. For reference thanks for pinning chef-client at release 4.2.4 to avoid opscode-cookbooks/chef-client#286 which removes support for our old Chef server version.

cbaenziger:
Erm I just realized, the symlink could have been done in `bcpc::bootstrap` to ensure idempotency so we do not have any manual re-work on existing clusters.

Issue #171, opened 2015-06-05
Issue starting mysql: Error executing action `start` on resource 'service[mysql]'


pu239ppy:
Creating a cluster today fails with MySQL failing to start due to what seems to be upstream changes in Percona.

 ````
================================================================================
Error executing action `start` on resource 'service[mysql]'
 ================================================================================


Additional information:
Mixlib::ShellOut::ShellCommandFailed
 ------------------------------------
Expected process to exit with [0], but received '1'
---- Begin output of /etc/init.d/mysql start ----
STDOUT: THIS IS WHAT I GOT
/var/lib/mysql
/var/lib/mysql/bcpc-vm1.pid
 * Starting MySQL (Percona XtraDB Cluster) database server mysqld
 * Stale PID file found: /var/lib/mysql/bcpc-vm1.pid with 19050
 * Please take a look at the syslog.
   ...fail!
STDERR:
---- End output of /etc/init.d/mysql start ----
Ran /etc/init.d/mysql start returned 1


Resource Declaration:
 ---------------------
 # In /var/chef/cache/cookbooks/bcpc/recipes/mysql.rb

 93: service "mysql" do
 94:   supports :status => true, :restart => true, :reload => false
 95:   action [ :enable, :start ]
 96: end
 97:
````

pu239ppy:
The issues happens is triggered in /etc/init.d/mysql Line126
Manually executing the command reveals the cause
ubuntu@bcpc-vm1:~$ mysqladmin ping
mysqladmin: unknown variable 'max_allowed_packet=32M'

While max_allowed_packet is a legal configuration option mysqladmin does not like it.

Possible workarounds:
Comment out lines 126,127, and 128 in /etc/init.d/mysql
OR
in the [client] section of /etc/mysql/conf.d/wsrep.cnf
modify max_allowed_packet=32M to be loose-max_allowed_packet=32M

pu239ppy:
Closing since the request has been merged

PR #173, opened 2015-06-08
mark max_allowed_packet as loose

pu239ppy:
prepending loose- in front of configuration parameters that mysqladmin is not aware of will prevent it form exiting with error.  Instead it will only issue a warning.  Issue #171


kiiranh:
I've tested this workaround on my vbox vm cluster. Mysql does come up and is operational

PR #174, opened 2015-06-12
Code to start ApplicationTimeLineServer 

arunvoma:
It contains all the related code to deploy and start the ApplicationTimelineserver.


amithkanand:
Can you please squash your commits into one? All the changes belong to `ATS`.

cookbooks/bcpc-hadoop/attributes/default.rb
```diff
@@ -4,6 +4,8 @@
 #
 #############################################
 
+
+default['yarn']['timelineservice']['hostname']="f-bcpc-vm2.bcpc.example.com"
```
cbaenziger: This should not be a node attribute but a search using `get_nodes_for()` so that it is found dynamically.

arunvoma:
ATS HA is not supported in this current version but its a requirement in next version of  ATS v2.Reference
https://issues.apache.org/jira/browse/YARN-2928

pu239ppy:
Work will be completed in #717

PR #175, opened 2015-06-15
Changes to increase HDFS resources for HBase

bijugs:
These are some of the recommended increases in HDFS for clusters running HBase.

**Note**: Changes were tested on a VM cluster and the parameter values were tested on a physical cluster.


cbaenziger:
Biju, thanks for the PR.

One concern I have is that VMs won't have 4GB if the datanode should get memory hungry. Can this follow the [same type](https://github.com/bloomberg/chef-bach/commit/cb797d0923f62d1618f3e68a438ee35089696ff9#diff-7a2d0681c0ee04a430b036dcdd1eba2c) of auto-tuning using a ratio or fixed setting that components like the YARN node manager do?

bijugs:
@cbaenziger . Thanks for reviewing. I have made changes for the comments. Let me know if that looks good and I can squash the commits.

cbaenziger:
While finally realizing that Xceivers are "threads"(?) and something akin to 500 threads per core seems like a concerningly high upper bound -- I suppose a thread could be doing input to the datanode and blocking on being written to disk or some such; or at least not all vieing to be on the run queue. Further, for tuning advise with HBase there's the nice Cloudera [article](http://blog.cloudera.com/blog/2012/03/hbase-hadoop-xceivers/).

cbaenziger:
Sorry to language lawyer but I would feel much better using something named:
````
...["datanode"]["xmx"]["max_size"] = 4096
...["datanode"]["xmx"]["max_ratio"] = 0.25
````
This disambiguates that these are maximum values for the data node hopefully versus the YARN nodemanager [settings](https://github.com/bloomberg/chef-bach/blob/cb797d0923f62d1618f3e68a438ee35089696ff9/cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb#l147-149) which are greedy and use minimum values.

bijugs:
@cbaenziger , Changes made for the review comments and the commits squashed.

PR #180, opened 2015-06-19
Serial console support

pu239ppy:
After the hosts have been built we want to be able to login via serial ports

### DIFF
--- a/build_bins.sh
+++ b/build_bins.sh
@@ -42,7 +42,7 @@ apt-get -y update
 # Install tools needed for packaging
 apt-get -y install git rubygems make pbuilder python-mock python-configobj python-support cdbs python-all-dev python-stdeb libmysqlclient-dev libldap2-dev ruby-dev gcc patch rake ruby1.9.3 ruby1.9.1-dev python-pip python-setuptools dpkg-dev apt-utils haveged libtool autoconf unzip rsync autogen
 if [[ -z `gem list --local fpm | grep fpm | cut -f1 -d" "` ]]; then
-  gem install fpm --no-ri --no-rdoc
+  gem install fpm --no-ri --no-rdoc -v 1.3.3
 fi
 
 # Download jmxtrans zip file
--- a/cookbooks/bcpc/attributes/default.rb
+++ b/cookbooks/bcpc/attributes/default.rb
@@ -22,6 +22,8 @@ default['bcpc']['vms_key'] = nil
 
 default['bcpc']['encrypt_data_bag'] = false
 
+default['bcpc']['bootstrap']['preseed']['add_kernel_opts'] = "console=ttyS0"
+
 default['bcpc']['bootstrap']['preseed']['late_command'] = "true"
 
 default['bcpc']['bootstrap']['admin_users'] = []
--- a/cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
+++ b/cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
@@ -88,6 +88,7 @@ d-i     grub-installer/with_other_os    boolean true
 d-i     finish-install/keep-consoles    boolean false
 d-i     finish-install/reboot_in_progress       note
 d-i     cdrom-detect/eject      boolean true
+d-i     debian-installer/add-kernel-opts string <%= @node[:bcpc][:bootstrap][:preseed][:add_kernel_opts] %> 
 d-i     debian-installer/exit/halt      boolean false
 d-i     debian-installer/exit/poweroff  boolean false
 d-i     pkgsel/include string openssh-server lldpd
--- a/enroll_cobbler.sh
+++ b/enroll_cobbler.sh
@@ -38,7 +38,7 @@ for i in bcpc-vm1 bcpc-vm2 bcpc-vm3; do
   fi 
   echo "Registering $i with $MAC for ${subnet}.${node}"
   if hash vagrant 2>/dev/null; then
-    vagrant ssh -c "sudo cobbler system remove --name=$i; sudo cobbler system add --name=$i --hostname=$i --profile=bcpc_host --ip-address=${subnet}.${node} --mac=${MAC}"
+    vagrant ssh -c "sudo cobbler system remove --name=$i; sudo cobbler system add --name=$i --hostname=$i --profile=bcpc_host --ip-address=${subnet}.${node} --mac=${MAC} --kopts=\"console=ttyS0\""
   else
     ssh -t -i $KEYFILE ubuntu@$1 "sudo cobbler system remove --name=$i; sudo cobbler system add --name=$i --hostname=$i --profile=bcpc_host --ip-address=${subnet}.${node} --mac=${MAC}"
   fi
--- a/setup_chef_cookbooks.sh
+++ b/setup_chef_cookbooks.sh
@@ -56,7 +56,7 @@ EOF
 cd cookbooks
 
 # allow versions on cookbooks via "cookbook version"
-for cookbook in "apt 2.4.0" python build-essential ubuntu cron "chef-client 4.2.4" chef-vault ntp yum logrotate yum-epel sysctl chef_handler 7-zip windows ark sudo ulimit pam ohai "poise 1.0.12" graphite_handler java maven; do
+for cookbook in "apt 2.4.0" python build-essential ubuntu cron "chef-client 4.2.4" chef-vault ntp yum logrotate yum-epel sysctl chef_handler 7-zip "windows 1.36.6" ark sudo ulimit pam ohai "poise 1.0.12" graphite_handler java maven; do
   if [[ ! -d ${cookbook% *} ]]; then
      # unless the proxy was defined this knife config will be the same as the one generated above
     knife cookbook site download $cookbook --config ../.chef/knife.rb
--- a/vbox_create.sh
+++ b/vbox_create.sh
@@ -177,6 +177,9 @@ function create_bootstrap_VM {
           # Add the bootable mini ISO for installing Ubuntu 12.04
           $VBM storageattach $vm --storagectl "IDE Controller" --device 0 --port 0 --type dvddrive --medium ubuntu-12.04-mini.iso
           $VBM modifyvm $vm --boot1 disk
+          # Add serial ports
+          $VBM modifyvm $vm --uart1 0x3F8 4
+          $VBM modifyvm $vm --uartmode1 server /tmp/serial-${vm}-ttyS0
       fi
     done
   fi
@@ -219,6 +222,9 @@ function create_cluster_VMs {
 
           # Set hardware acceleration options
           $VBM modifyvm $vm --largepages on --vtxvpid on --hwvirtex on --nestedpaging on --ioapic on
+          # Add serial ports
+          $VBM modifyvm $vm --uart1 0x3F8 4
+          $VBM modifyvm $vm --uartmode1 server /tmp/serial-${vm}-ttyS0
       fi
   done
 }


cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,59 @@
...
+            device_baud = `stty -F /dev/#{serial_dev_type}#{port_number} speed`
+            device_baud = device_baud.chomp
+            if not numeric?(device_baud)
+                next
+            end
+            device_name = "#{serial_dev_type}#{port_number}"
+            serial_devices[device_name] = device_baud
+        end
+    end
+end
```
DrRayWang: Could you wrap line 5 - 29 into a method in library util.rb? I think this recipe likes clean chef code.

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,59 @@
...
+    block do
+        serial_devices.each do |serial_device, serial_baud|
+            Chef::Log.info("Serial device found: #{serial_device}:#{serial_baud}" )
+        end
+    end
+end
+
+serial_devices.each_key do |serial_device|
+    service "#{serial_device}" do
+       action :nothing
```
DrRayWang: Could you explain why you roll over all the services but do nothing?

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,59 @@
+def numeric?(data)
+        Float(data) != nil rescue false
+end
+
+serial_devices = {}
+# we are not interested in USB or any other serial at the moment
+serial_dev_type = "ttyS"
+serial_driver_info = "/proc/tty/driver/serial"
+File.open(serial_driver_info, "r") do |serials|
```
cbaenziger: Hi Lev, could you add a comment showing the expected format you expect from the serial file? (I don't suspect it will change but if it does, it will help some future code spelunker understand your logic for a fix.)

cookbooks/bcpc/recipes/serial_console.rb
cbaenziger: Ray, do you expect other consumers of this code outside the serial console recipe?

cbaenziger:
While it is very authoritative, could you provide a more meaningful commit message than "this will the new role to all hosts"?

cookbooks/bcpc/recipes/serial_console.rb
DrRayWang: Just think it is more clean to wrap a large piece of pure ruby code in library method. Maybe in future it can be used for other purpose.

pu239ppy:
I have added comments to specify what expected file format or command out put needs to look like.  Also all work now happens at convergence time in a ruby_block resource

amithkanand:
Do we need three commits? can these commits be squashed?

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,71 @@
+# declare a placeholder template
+template "/etc/init/ttySx.conf" do
+    source "ttySx.erb"
```
amithkanand: Can we keep indentation to 2 spaces to keep it in align with all other recipes we have?

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,71 @@
...
+        end
+        # we are not interested in USB or any other serial at the moment
+        # Sample file format where the first column
+        # number x corresponds to ttySx
+        # serinfo:1.0 driver revision:
+        # 0: uart:16550A port:000003F8 irq:4 tx:165 rx:0 CTS|DSR|CD
+        # 1: uart:unknown port:000002F8 irq:3
+        # 2: uart:unknown port:000003E8 irq:4
+        serial_dev_type = "ttyS"
+        serial_driver_info = "/proc/tty/driver/serial"
```
amithkanand: Will it be a better idea to put this into an attribute?

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,71 @@
...
+                    if not numeric?(device_baud)
+                        next
+                    end
+                    device_name = "#{serial_dev_type}#{port_number}"
+                    Chef::Log.info("Serial device found: #{device_name}:#{device_baud}")
+                    Chef::Log.info("Running template for: #{device_name}:#{device_baud}")
+                    # load the template resource and modify
+                    tmpl = run_context.resource_collection.find("template[/etc/init/ttySx.conf]")
+                    tmpl.path("/etc/init/#{device_name}.conf")
+                    tmpl.source("ttySx.erb")
```
amithkanand: Isn't this already defined in the template resource?

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,71 @@
...
+                    tmpl.variables(:serial_params => {
+                        "serial_device" => device_name,
+                        "device_baud" => device_baud })
+                    tmpl.owner("root")
+                    tmpl.group("root")
+                    tmpl.mode(00644)
```
amithkanand: Variables `owner`, `group` and `mode` are already defined in the template resource. Do we need to set/define again?

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,71 @@
...
+                    device_baud = device_baud.chomp
+                    if not numeric?(device_baud)
+                        next
+                    end
+                    device_name = "#{serial_dev_type}#{port_number}"
+                    Chef::Log.info("Serial device found: #{device_name}:#{device_baud}")
+                    Chef::Log.info("Running template for: #{device_name}:#{device_baud}")
+                    # load the template resource and modify
+                    tmpl = run_context.resource_collection.find("template[/etc/init/ttySx.conf]")
+                    tmpl.path("/etc/init/#{device_name}.conf")
```
amithkanand: If template path is derived in the ruby block, can we change the template name above to be something else instead of path?

cookbooks/bcpc/recipes/serial_console.rb
pu239ppy: I don't know if this is important enough to persist in attribute space

cookbooks/bcpc/recipes/serial_console.rb
pu239ppy: The "dummy" template resource is only there as a placeholder such that it can be loaded with tmpl = run_context.resource_collection.find("template[/etc/init/ttySx.conf]") and assigned a more specific path of attributes such as the actual file path overriding the "name" attribute. Note that the action on the template is :nothing.  Because I may have to potentially call on a template in a loop (once per serial port discovered) it is either at compile (see previous revisions) or convergence time.

cookbooks/bcpc/recipes/serial_console.rb
pu239ppy: Is redundant I will fix it in one place or another

cookbooks/bcpc/recipes/serial_console.rb
pu239ppy: Sure, I will come up with something more descriptive

pu239ppy: Why, the template name tells us exactly what the intent is

amithkanand: If I look at the template resource, I would assume that the template will be triggered by some other resource and it will create `/etc/init/ttySx.conf`  file. But later  in the ruby block the template is being changed dynamically to `tmpl.path("/etc/init/#{device_name}.conf")` which may never create the originally defined file `/etc/init/ttySx.conf`.

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,71 @@
...
+                end
+                line_parts = line.split()
+                uart = line_parts[1]
+                if not uart.include? "uart:unknown"
+                    port_assignment = line_parts[0]
+                    port_number = port_assignment.chop
+                    # get device baud we expect
+                    # root@bcpc-vm2:/proc/tty/driver# stty -F /dev/ttyS0 speed
+                    # 115200
+                    device_baud = `stty -F /dev/#{serial_dev_type}#{port_number} speed`
```
amithkanand: The OS command can be executed using Mixlib::ShellOut in more controlled way and better error handling. One of the example is [`namenode_master.rb`](https://github.com/amithkanand/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/namenode_master.rb#L126-L136)

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,71 @@
...
+        File.open(serial_driver_info, "r") do |serials|
+            serials.each_line do |line|
+                if line.start_with? "serinfo:"
+                    next
+                end
+                line_parts = line.split()
+                uart = line_parts[1]
+                if not uart.include? "uart:unknown"
+                    port_assignment = line_parts[0]
+                    port_number = port_assignment.chop
```
amithkanand: Since `port_assignment` is not used anywhere else the code can be changed to `port_number = line_parts[0].chop`.

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,68 @@
...
+    File.open(serial_driver_info, "r") do |serials|
+      serials.each_line do |line|
+        if line.start_with? "serinfo:"
+          next
+        end
+        line_parts = line.split()
+        uart = line_parts[1]
+        if not uart.include? "uart:unknown"
+          port_assignment = line_parts[0]
+          port_number = port_assignment.chop
```
amithkanand: Since port_assignment is not used anywhere else the code can be changed to port_number = `line_parts[0].chop`.

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,68 @@
...
+          # get device baud we expect
+          # root@bcpc-vm2:/proc/tty/driver# stty -F /dev/ttyS0 speed
+          # 115200
+          device_baud = `stty -F /dev/#{serial_dev_type}#{port_number} speed`
```
amithkanand: The OS command can be executed using Mixlib::ShellOut in more controlled way and better error handling. One of the example is [`namenode_master.rb`](https://github.com/amithkanand/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/namenode_master.rb#L126-L136)

cookbooks/bcpc/recipes/serial_console.rb
```diff
@@ -0,0 +1,72 @@
...
+            "device_baud" => device_baud })
+          tmpl.owner("root")
+          tmpl.group("root")
+          tmpl.mode(00644)
+          tmpl.run_action(:create)
+          # load the service resource and create
+          Chef::Log.info("Creating service for: #{device_name}:#{device_baud}")
+          svc = run_context.resource_collection.find("service[ttySx_svc]")
+          svc.service_name(device_name)
+          svc.run_action(:start)
```
amithkanand: Should we also `enable` the service to start it automatically?

pu239ppy: There is no need, the job already defines run levels.  Unlike SysV init, in upstart  there are no run level specific directories to create symlinks

cbaenziger:
Lev did you want to modify grub and the cobbler recipe to enable this for boot too?

cbaenziger:
So strange to see inittab gone: https://help.ubuntu.com/community/SerialConsoleHowto

http-418:
Works for me using socat.

PR #183, opened 2015-06-23
Changes to tune Linux kernel tcp parameters

bijugs:
There are two commits in this PR.

1) A generic definition which can be used to install/uninstall kernel modules.
2) Changes to kernel TCP parameters to reduce/reuse sockets in TIME_WAIT status which may be causing connection failures to datanodes from Hbase region servers when there is high volume read requests.

**Note**: At this stage this PR is made available for review and comments. These changes are tested on VM cluster. Will be performing tests on a physical cluster.

### DIFF
--- a/cookbooks/bcpc/definitions/kernel_module.rb
+++ b/cookbooks/bcpc/definitions/kernel_module.rb
@@ -1,0 +1,18 @@
+define :kernel_module, :action => :install, :search_name => nil do
+
+  params[:search_name] ||= params[:name]
+
+  if params[:action] == :install
+    bash "install module #{params[:name]}" do
+      code "modprobe #{params[:name]}"
+      not_if "lsmod | grep #{params[:name]}"
+    end
+  end
+
+  if params[:action] == :uninstall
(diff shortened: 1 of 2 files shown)


http-418:
Rui, maybe we should have your kernel modules definition add the module to /etc/modules, so it is loaded at boot.

If we don't want to own the /etc/modules file, that could be done with augeas.

cookbooks/bcpc/definitions/kernel_module.rb
```diff
@@ -0,0 +1,18 @@
+define :kernel_module, :action => :install, :search_name => nil do
+
+  params[:search_name] ||= params[:name]
+
+  if params[:action] == :install
+    bash "install module #{params[:name]}" do
+      code "modprobe #{params[:name]}"
+      not_if "lsmod | grep #{params[:name]}"
+    end
+  end
```
cbaenziger: Why not add to `/etc/modules` too?

vt0r:

Issue #185, opened 2015-06-23
Incorrect configuration location in mapred-site.xml


amithkanand:
In `mapred-site.xml` the parameter `mapreduce.application.classpath` defines `/etc/hadoop/conf/secure` as location for `hadoop` configuration files. This causes incorrect location to be picked up within `mapper` of a `MR` job. 
For example:

```
FileSystem.get(new Configuration()) Within mapper of a MR Job defaults to local 
filesystem and this causes the following code to fail 
fs = FileSystem.get(new Configuration()) 
fs.exists(hdfs://user/xxx/a.txt)
```

The fix is to remove `/etc/hadoop/conf/secure` from the classpath.

cbaenziger:
Hrm, it is strange an empty directory in the CLASSPATH would break things so badly? Does any empty directory do this?

PR #186, opened 2015-06-23
Remove incorrect configuration file location from mapred-site.xml

amithkanand:
This PR takes care of removing `/etc/hadoop/conf/secure` from `mapreduce.application.classpath`. Please refer Issue #185 for more details.


cbaenziger:
As mentioned on #185, I find it strange that this would cause the effect we are seeing but this is certainly wrong either way.

Issue #177, opened 2015-06-16
Extend current graphite and Zabbix monitoring framework to support non-jmx data


kiiranh:
Currently, the graphite data is grouped in [logical groups](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc_jmxtrans/attributes/default.rb#L283) for Hadoop. The [graphite queries ](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/templates/default/graphite.query_graphite.config.erb#L6) which fetch data, do it considering these logical groups and send aggregate data for all hosts in that group to Zabbix. In Zabbix, [hosts are the logical groups](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/attributes/default.rb#L124) which encompass individual hosts of same type; for e.g.: hbase_master represents all hbase masters. All this was designed to work with JMX data for hadoop components. 

Now, if one wants to monitor and alarm on data collected by other means (say diamond) and wishes to trigger for individual hosts, the current framework does not work. Changes need to be made so arbitrary hosts can be monitored using graphite and Zabbix.

kiiranh:
PR #187 has been closed. Resolving this.

PR #187, opened 2015-06-25
Extend graphite_to_zabbix to support arbitrary metrics from graphite

kiiranh:
This PR extends the graphite-zabbix monitoring framework to support arbitrary metrics(e.g.: diamond) in addition to the originally defined jmx metrics. The structure used to define the queries and triggers is now modeled so its easy to override a particular key-value in a recipe or the environment file.

This fixes issue #177.


cbaenziger:
Thanks for breaking the attributes out to their own file. However, can the service specific checks be added in by those recipes somehow? (E.g. what if we are not running HDFS or HBase but still deploying the monitoring and alarming framework?)

kiiranh:
To confirm, by service specific checks, you mean if hbase/hdfs is not installed then do not configure the related triggers in Zabbix and graphite.

kiiranh:
We could set the `trigger_enable` flag to false by default and enable them in environments in which they are relevant.

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,109 @@
...
+default["bcpc"]["hadoop"]["zabbix"]["mail_source"] = "zabbix.zbx_mail.sh.erb"
+default["bcpc"]["hadoop"]["zabbix"]["cookbook"] = nil 
+
+# Graphite queries which specify property to query and alarming trigger, severity(maps to zabbix's api -> trigger -> priority)
+# and owner who the trigger is routed to for resolution
+# Queries are structured as they appear in graphite dashboard
+# Refer Zabbix API(https://www.zabbix.com/documentation/2.2/manual/api/reference/) for the various attributes used
+# The following list is supposed to be a skeleton which should be extended/edited according to the deployment scenario
+# By default items, triggers actions are disabled and won't be monitored. To enable them, set "enable=true" in the json structure
+default["bcpc"]["hadoop"]["graphite"]["queries"] = {
```
cbaenziger: Can these be added to the germane recipe for the query? (E.g. could the namenode* recipes deploy the namenode queries as we do for the [copylogs](https://github.com/bloomberg/chef-bach/blob/88a319e7520f30442a4e1dfe3ef2e9016e5cc9ca/cookbooks/bcpc-hadoop/recipes/namenode_master.rb#L11-L19) parameters?)

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,109 @@
...
+         'trigger_cond' => "=0",
+         'trigger_name' => "NodeAvailability",
+         'enable' => true,
+         'trigger_desc' => "Node seems to be down",
+         'severity' => 5,
+         'route_to' => "admin"
+      },
+      'disk_space_avail' => {
+         'type' => "servers",
+         'query' => "diskspace._disk_0.byte_avail",
```
cbaenziger: Can I do a glob pattern e.g. for disk_* a minimum is reaching 0 on this machine?

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,109 @@
...
+         'trigger_val' => "max(61,0)",
+         'value_type' => 3,
+         'trigger_cond' => "=0",
+         'trigger_name' => "DiskAvailability",
+         'enable' => false,
+         'trigger_desc' => "Disk seems to be down",
```
cbaenziger: The disk is not only down, it may be full, right?

cookbooks/bcpc-hadoop/attributes/monitoring.rb
kiiranh: The problem with that is, one cannot easily view all the triggers defined for the cluster. If we encourage users to write triggers based on what makes sense to them, it would be good to have them write in the attributes file rather than modify the namenode and other important recipes. That works for copylogs because it is internal and users won't be modifying it.

cookbooks/bcpc-hadoop/attributes/monitoring.rb
kiiranh: Yes, this is just a graphite query and one can specify a glob to club data from all disks.

cookbooks/bcpc-hadoop/attributes/monitoring.rb
kiiranh: diamond collects the following metrics(among others) related to disk space:
server.diskspace.disk#.byte_avail
server.diskspace.disk#.byte_free
server.diskspace.disk#.byte_used

I had assumed byte_avail = total disk capacity, byte_free = free disk capacity, byte_used = used disk capacity. But I'm seeing byte_avail == byte_free in graphite and using python os.statvfs(). Considering that, Yes, this condition would mean either disk is down or full. I'll edit the description.

cookbooks/bcpc-hadoop/attributes/monitoring.rb
cbaenziger: But we don't want them modifying the default attribute file either. Can't they nullify these changes if here or in a recipe?

kiiranh: Yes, I see your point. Will make changes.

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,121 @@
...
+         'trigger_name' => "DiskSpaceUsage",
+         'enable' => false,
+         'trigger_desc' => "More than 15G disk used",
+         'severity' => 3,
+         'route_to' => "tenant"
+      },
+      'chef_client_time' => {
+         'type' => "chef",
+         'query' => "elapsed_time",
+         'trigger_val' => "max((#{node['chef_client']['interval']}.to_i + #{node['chef_client']['splay']}.to_i ) / 60 ,0)",
```
cbaenziger: The Zabbix [max](https://www.zabbix.com/documentation/2.2/manual/appendix/triggers/functions) function seems to take seconds, here you are dividing seconds (the Chef run interval) by 60?

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,121 @@
...
+         'route_to' => "admin"
+      },
+      'disk_space_used' => {
+         'type' => "servers",
+         'query' => "diskspace._disk_*.byte_used",
+         'trigger_val' => "max(61,0)",
+         'value_type' => 3,
+         'trigger_cond' => ">15G",
+         'trigger_name' => "DiskSpaceUsage",
+         'enable' => false,
```
cbaenziger: Why not enable these?

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,121 @@
...
+         'severity' => 5,
+         'route_to' => "admin"
+      }
+   },
+   'bcpc-vm' => {
+      # Entry in suffix array added to hostname as suffix and 
+      # to key name, trigger name as suffix "_<hostname with suffix>"
+      # If specified, should be *first* entry for a host
+      'suffix' =>  [ "1", "2", "3" ], 
+      'memory_active' => {
```
cbaenziger: No `value_type`? (What is a `value_type`?)

kiiranh: [value_type](https://www.zabbix.com/documentation/2.2/manual/api/reference/item/object) is `Type of information of item`. [Default value](https://github.com/kiiranh/chef-bach/blob/monitoring/cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb#L98-L102) is set to *3 (numeric unsigned)*

kiiranh:
Code updated to handle dependencies between triggers, updates of zabbix artifacts on changing most items in attributes, added basic triggers.
Following functions from review comments haven't been implemented and are TODOs:
1. Current trigger notifies user only when an item goes to PROBLEM state. Change this so that user is notified when the item gets back to OK state too. This can be done by removing the PROBLEM state condition from action definition.
2. Set the chef-client status check interval to factor in the chef-client run timeout. I could not find how to get the default timeout value, hence the current interval for zabbix to check the chef-client status is set to `((chef-client interval + splay) * 2) minutes`

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,90 @@
...
+default["bcpc"]["hadoop"]["zabbix"]["mail_source"] = "zabbix.zbx_mail.sh.erb"
+default["bcpc"]["hadoop"]["zabbix"]["cookbook"] = nil 
+
+# Graphite queries which specify property to query and alarming trigger, severity(maps to zabbix's api -> trigger -> priority)
+# and owner who the trigger is routed to for resolution
+# Queries are structured as they appear in graphite dashboard
+# Refer Zabbix API(https://www.zabbix.com/documentation/2.2/manual/api/reference/) for the various attributes used
+# The following list is supposed to be a skeleton which should be extended/edited according to the deployment scenario
+# By default items, triggers actions are disabled and won't be monitored. To enable them, set "enable=true" in the json structure
+["bcpc-vm1", "bcpc-vm2", "bcpc-vm3"].each do |host|
```
cbaenziger: If this can't use node search here, can we use a cookbook to set this? As if we build any real cluster (not using bcpc-vm's 1-3), it'll be a pain to always have to override this.

cookbooks/bcpc-hadoop/recipes/hbase_master.rb
```diff
@@ -13,6 +13,66 @@
     'docopy' => true
 }
 
+# Define zabbix triggers for alarming
+# The Hbase triggers depend on the namenode triggers too. So the namenode triggers
+# are also defined here. This takes care of creation of triggers irrespective of
+# how namenode and hbase_master appear in the runlist. Assumption is that hdfs will
+# be installed if hbase is installed
+node.default['bcpc']['hadoop']['graphite']['queries']['namenode'] = {
```
cbaenziger: Dude; this is hbase master not namenode. Please understand the trigger dependency race you were seeing or add support in `graphite_to_zabbix.rb` to only add trigger dependencies once a trigger is created (and log that you are skipping the dependency).

cookbooks/bcpc-hadoop/templates/default/graphite.query_graphite.config.erb
```diff
@@ -1,17 +1,23 @@
...
-    <% end %>
-  <% end %>
-<% end %>
- ]
+  "queries":[
+  <% node["bcpc"]["hadoop"]["graphite"]["queries"].each_with_index do |(host, queries),hidx| %>
+    <% queries.each_with_index do |(key, attrs),qidx| %>
+      <% if attrs['type'] == 'jmx' %>
+        {"graphite_url": "https://<%=node[:bcpc][:management][:vip]%>:<%=node[:bcpc][:graphite][:web_port]%>/render/?from=-2min&target=<%=attrs['type']%>.<%=host%>.*.<%=attrs['query']%>&format=json",
+      <% elsif attrs['type'] == 'chef' %>
```
cbaenziger: Can this be collapsed into the default case since it and servers are the same.

cookbooks/bcpc-hadoop/templates/default/graphite.query_graphite.config.erb
```diff
@@ -1,17 +1,23 @@
...
+      <% elsif attrs['type'] == 'function' %>
+        {"graphite_url": "https://<%=node[:bcpc][:management][:vip]%>:<%=node[:bcpc][:graphite][:web_port]%>/render/?from=-2min&target=<%=attrs['query']%>&format=json",
+      <% else #servers %>
+        {"graphite_url": "https://<%=node[:bcpc][:management][:vip]%>:<%=node[:bcpc][:graphite][:web_port]%>/render/?from=-2min&target=<%=attrs['type']%>.<%=host%>.<%=attrs['query']%>&format=json",
+      <% end %>
+        "zabbix_host" : "<%=host%>",
+      <% if hidx == node[:bcpc][:hadoop][:graphite][:queries].length-1 and qidx == queries.length()-1 %>
+        "zabbix_key" : "<%=key%>"}
+      <% else %>
+        "zabbix_key" : "<%=key%>"},
```
cbaenziger: Erm, might be easier to just make this all a Ruby hash and call the JSON gems' [generate](http://ruby-doc.org/stdlib-2.0.0/libdoc/json/rdoc/JSON.html#module-JSON-label-Generating+JSON).

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,90 @@
...
+      'value_type' => 3,
+      'trigger_cond' => "=0",
+      'trigger_name' => "ChefClientSuccess_#{host}",
+      'trigger_dep' => ["NodeAvailability_#{host}"],
+      'enable' => true,
+      'trigger_desc' => "Chef-client seems to be failing/stopped",
+      'severity' => 1,
+      'route_to' => "admin"
+    },
+    "disk_0_space_avail_#{host}" => {
```
cbaenziger: Disks should be programatically created; Ohai can provide this info.

cookbooks/bcpc-hadoop/recipes/hbase_master.rb
kiiranh: The problem was, I had included namenode related queries in `namenode_master` recipe which was not in the run list on the VM cluster (default mode is Non-HA). The non-HA mode runs `namenode_no_HA` recipe.

cookbooks/bcpc-hadoop/templates/default/graphite.query_graphite.config.erb
kiiranh: Doing that means converting the existing hash of hashes to a new hash of hashes, then converting to json and then pretty'ing it. This uses a bit of logic to identify end of list but is simpler.

cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb
```diff
@@ -29,156 +31,144 @@
...
       Chef::Log.fatal!("Fatal error: could not connect to Zabbix server")
     end
-    #
-    # Looping through the graphite queries attribute to define the required zabbix hosts, items and triggers
-    #
+
+    # Fetch graphite hosts
     graphite_hosts = (get_node_attributes(MGMT_IP_ATTR_SRCH_KEYS,"graphite","bcpc").map {|v| v['mgmt_ip']}).join(",")
+    if graphite_hosts.empty?
+      Chef::Log.fatal!("Fatal error: No graphite hosts found")
```
cbaenziger: Do you intend to continue running after this fatal log message?

kiiranh: Logical thing would be to *not* continue. Same would be true for the `zbx.nil?` check

kiiranh:
Code updated to end chef run if Zabbix API endpoint or graphite hosts are not accessible.

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -0,0 +1,5 @@
+default["bcpc"]["hadoop"]["zabbix"]["history_days"] = 1
+default["bcpc"]["hadoop"]["zabbix"]["trend_days"] = 15
+default["bcpc"]["hadoop"]["zabbix"]["cron_check_time"] = 240
```
cbaenziger: It looks like `cron_check` is being pulled out? (For what it is worth, I believe cron_check was the verify that the Zabbix sender had not died; I'm not sure why it got commented out though -- or was not enabled?)

kiiranh: No code was changed related to `cron_check`. The attributes were just moved from `default.rb`

cookbooks/bcpc-hadoop/recipes/os_basic_queries.rb
```diff
@@ -0,0 +1,75 @@
...
+
+# Graphite queries which specify property to query and alarming trigger, severity(maps to zabbix's api -> trigger -> priority)
+# and owner who the trigger is routed to for resolution
+# Queries are structured as they appear in graphite dashboard
+# Refer Zabbix API(https://www.zabbix.com/documentation/2.2/manual/api/reference/) for the various attributes used
+# The following list is supposed to be a skeleton which should be extended/edited according to the deployment scenario
+# By default items, triggers actions are disabled and won't be monitored. To enable them, set "enable=true" in the json structure
+monitored_nodes_objs.each do |node_obj|
+  host = node_obj.hostname
+  node.default["bcpc"]["hadoop"]["graphite"]["queries"]["#{host}"] = {
```
cbaenziger: Hrm, I'm not sure about storing another hosts' monitoring metadata in this host' node object. This may cause an unalterably large node object. Can we structure this in a way where each node holds its own to-be-monitored attributes even if it is not registering the triggers? (Or perhaps it would get the defaults unless it provides an override?)

cookbooks/bcpc-hadoop/recipes/os_basic_queries.rb
```diff
@@ -0,0 +1,75 @@
...
+        'trigger_name' => "#{host}_NodeAvailability",
+        'enable' => true,
+        'trigger_desc' => "Node seems to be down",
+        'severity' => 5,
+        'route_to' => "admin"
+      },
+      "chef_client_success_#{host}" => {
+        'type' => "chef",
+        'query' => "success",
+        'trigger_val' => "max(#{((node['chef_client']['interval'].to_i + node['chef_client']['splay'].to_i) / 30).ceil}m, 0)",
```
cbaenziger: The arithmetic here is confusing to me; what are you doing?
` 'trigger_val' => "max(#{((node['chef_client']['interval'].to_i + node['chef_client']['splay'].to_i) / 30).ceil}m, 0)",`

kiiranh: If chef_client interval is 300sec and splay is 60s, we check if the chef-client ran in the last
2 * [ (interval+splay)/60 ] = (interval+splay)/30 minutes. 
Basically, in theory chef-client should run least every (interval+splay) sec; we check the metric for twice that interval.

cookbooks/bcpc-hadoop/attributes/monitoring.rb
kiiranh: Put back the `cron_check` code; however keeping it commented since it would take some time to properly understand that function and implement/test it. Can be done in a later update to the work under this PR

cbaenziger:
Nit: if possible, I would really like to see the alarming components end up back in bcpc from bcpc-hadoop. As for OS queries, etc. there is no Hadoop dependency.

cbaenziger:
To try and decouple the queries from the Zabbix hosts' node objects, can os_basic_queries be modified akin to:
````
+triggers = {}
 monitored_nodes_objs.each do |node_obj|
   host = node_obj.hostname
-  node.default["bcpc"]["hadoop"]["graphite"]["queries"]["#{host}"] = { 
+  triggers[host] = node_obj["bcpc"]["hadoop"]["graphite"]["queries"] or { 
[...]
-    node.default["bcpc"]["hadoop"]["graphite"]["queries"]["#{host}"]["#{disk}_space_used_75_#{host}"] = {
+    triggers[host]["#{disk}_space_used_75_#{host}"] = node_obj["bcpc"]["hadoop"]["graphite"]["queries"]["#{disk}_space_used_75_#{host}"]
````

And have graphite_to_zabbix similarly walk the `triggers` hash instead of the host's node object?

Ideally, in the future we could make a more obvious data structure but for now this should provide the desired decoupling without too much complexity or surgery.

http-418:
I am building this right now.

kiiranh:
Thanks @http-418 . Code comments have been incorporated. Moving monitoring code to `bcpc` from `bcpc-hadoop` would be better handled in a separate PR.

http-418:
This looks good to me.   Cluster builds, is no more broken than usual, feature works as advertised.

kiiranh:
Note: Existing Zabbix objects may have to be cleaned before deploying this code, since it does't automatically delete unwanted stuff. A good way to deploy would be to stop chef-client, cleanup objects in Zabbix, deploy this updated code and start chef-client.

PR #189, opened 2015-07-01
Enhancement for supporting kerberos

amithkanand:
This PR adds gems to `build_bins.sh` that are required for `kerberos` implementation. Also, `Vagrantfile` is modified to to read `environment` specific `JSON` file and set `fqdn` host name that is needed to setup `Kerberos KDC`.


.gitignore
```diff
@@ -14,10 +20,19 @@ cookbooks/sysctl/
 cookbooks/ubuntu/
 cookbooks/yum/
 cookbooks/yum-epel/
-environments/*
+cookbooks/7-zip/
```
cbaenziger: Is it possible to do the following instead of calling out pulled down cookbooks:
````
cookbooks/*/
!cookbooks/bcpc
!cookbooks/bcpc-hadoop
!cookbooks/hannibal
[...]
````

File: Vagrantfile [hunk 2]
```diff
@@ -5,14 +5,38 @@
 # Chef server.
 # See http://www.vagrantup.com/ for info on Vagrant.
 
-$local_environment = "Test-Laptop"
+require 'json'
+
+if File.basename(File.expand_path(".")) == "vbox"
```
cbaenziger: What is this trying to do? A comment would be useful if it's for what I think it is.

cbaenziger:
If you are removing support for creating a VM to use as an apt mirror could you make that its own commit too please?

cbaenziger:
Looks good; I'm running a quick test on a cluster to see if anything unexpected crops up.

cbaenziger:
In #117, `vbox/.gitignore` is removed but it's not removed here?

Vagrantfile
```diff
@@ -5,14 +5,44 @@
...
+  puts "More than one environment file found."
+  exit
+end
+
+chef_env = JSON.parse(File.read(json_file.join(",")))
+cluster_environment = chef_env["name"]
+bootstrap_hostname = chef_env["override_attributes"]["bcpc"]["bootstrap"]["hostname"]
+bootstrap_domain = chef_env["override_attributes"]["bcpc"]["domain_name"]
+
+$local_environment = chef_env["name"]
```
cbaenziger: Perhaps `cluster_environment` instead?

cookbooks/bcpc-hadoop/attributes/default.rb
```diff
@@ -39,7 +39,7 @@
 default["bcpc"]["hadoop"]["namenode"]["jmx"]["port"] = 10111
 default["bcpc"]["hadoop"]["namenode"]["rpc"]["port"] = 8020
 default["bcpc"]["hadoop"]["namenode"]["http"]["port"] = 50070 
-default["bcpc"]["hadoop"]["namenode"]["https"]["port"] = 50080
```
cbaenziger: Why does this port change?

cookbooks/bcpc/recipes/cobbler.rb
```diff
@@ -37,7 +37,7 @@
 chef_vault_secret "cobbler" do
   data_bag 'os'
   raw_data({ 'web-password' => web_passwd, 'root-password' => root_passwd, 'root-password-salted' => root_passwd.crypt("$6$" + rand(36**8).to_s(36)) })
-  admins node[:hostname]
+  admins node[:fqdn]
```
cbaenziger: Amit and I discussed why this will not break backwards compatibility and it was pointed out older bootstrap nodes did not know their domain so their fqdn is equal to their hostname.

cookbooks/bcpc-hadoop/attributes/default.rb
amithkanand: This was changed as per [`Hortonworks Documentation`](http://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.2.0/HDP_Man_Install_v22/index.html#Item1.26.3.4.3).

cbaenziger:
I think this horse is dead...

PR #190, opened 2015-07-08
Provide Hadoop ecosystem ulimits via LWRP

cbaenziger:
This pull request replicates what used to be delivered into /etc/security/limits.d in HDP2.0 to be provided by Chef. (For reference those files have now moved to `/usr/hdp/<version>/etc/security/limits.d` but I think we can manage better not playing musical files.)

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/datanode.rb
+++ b/cookbooks/bcpc-hadoop/recipes/datanode.rb
@@ -18,6 +18,21 @@ node.default['bcpc']['hadoop']['copylog']['datanode'] = {
   end
 end
 
+user_ulimit "hdfs" do
+  filehandle_limit 32769
+  process_limit 65536
+end
+
+user_ulimit "mapred" do
+  filehandle_limit 32769
+  process_limit 65536
+end
+
+user_ulimit "yarn" do
+  filehandle_limit 32769
+  process_limit 65536
+end
+
 template "/etc/init.d/hadoop-hdfs-datanode" do
   source "hdp_hadoop-hdfs-datanode-initd.erb"
   mode 0655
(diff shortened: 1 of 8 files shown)


cookbooks/bcpc-hadoop/recipes/datanode.rb
```diff
@@ -18,6 +18,21 @@
   end
 end
 
+user_ulimit "hdfs" do
```
kiiranh: Is `user_ulimit` a built-in resource in Chef, or does it require us to download a user_ulimit cookbook?

cookbooks/bcpc-hadoop/recipes/datanode.rb
```diff
@@ -18,6 +18,21 @@
...
+  filehandle_limit 32769
```
kiiranh: make the numbers configurable or we are always going to have these? If they're not going to be changed often, I wouldn't want an attribute for it.

cookbooks/bcpc-hadoop/recipes/datanode.rb
kiiranh: Should [hadoop-hdfs-datanode service](https://github.com/cbaenziger/chef-bach/blob/ulimit/cookbooks/bcpc-hadoop/recipes/datanode.rb#L236-L240) subscribe to this resource?

cookbooks/bcpc-hadoop/recipes/hbase_master.rb
```diff
@@ -25,6 +25,10 @@
   end
 end
 
+user_ulimit "hbase" do
+  filehandle_limit 32769
```
kiiranh: Why no `process_limit` here as is for `hdfs`, `mapred`, `yarn` ?

cbaenziger:
Thanks for catching my `metadata.rb` miss; it's committed now.

The ulimits have largely not changed in my 5 years of Hadoop-ing that I can recall. Admittedly, I grabbed these ulimits by running `cat /usr/hdp/<version>/etc/security/limits.d/*` on a cluster, however. And the fact that I know ZK could run out of sockets (file descriptors) so I added that which is new from the HDP2.0 process.

The number of processes really should only matter for `yarn` and `mapred` from my view (they scale with the number of containers on a worker node). I can not contrive why one would end up with thousands of HDFS processes; but Hortonworks seems to think that is (or once was) wise.

The `user_ulimit` resource is provided by the `ulimit` cookbook. We have been using it in the [`kafka-bcpc::kafka`](https://github.com/bloomberg/chef-bach/blob/60059dc8c319a79371d14dafb32dad03b5332eb6/cookbooks/kafka-bcpc/recipes/kafka.rb#L5-L8) recipe

kiiranh:
Thanks for the explanation. LGTM.

pu239ppy:
Looks good

PR #191, opened 2015-07-13
Provide configurable namenode memory settings

cbaenziger:
This provides configurable namenode memory configuration (heap size and GC options). The memory ratio logic is based off of PR #175.

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/default.rb
+++ b/cookbooks/bcpc-hadoop/attributes/default.rb
@@ -31,12 +31,15 @@ default["bcpc"]["hadoop"]["jmx_enabled"] = true
 default["bcpc"]["hadoop"]["datanode"]["xmx"]["max_size"] = 4096
 default["bcpc"]["hadoop"]["datanode"]["xmx"]["max_ratio"] = 0.25
 default["bcpc"]["hadoop"]["datanode"]["max"]["xferthreads"] = 16384
+default["bcpc"]["hadoop"]["datanode"]["jmx"]["port"] = 10112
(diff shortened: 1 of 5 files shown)


cbaenziger:
FYI, if testing on an HDP2.0 cluster one will see something akin to the following with two `-Xmx` directives:
````
/usr/lib/jvm/java-7-openjdk-amd64/bin/java -Dproc_namenode -Xmx1000m -Dhadoop.log.dir=/var/log/hadoop-hdfs [...] -Dhadoop.policy.file=hadoop-policy.xml -Djava.net.preferIPv4Stack=true -Djava.net.preferIPv4Stack=true -Xms16384m -Xmx16384m -XX:ParallelGCThreads=14 [...]
````

The first `-Xmx1000m` comes from `/usr/lib/hadoop-hdfs/libexec/hadoop-config.sh`:
````
JAVA_HEAP_MAX=-Xmx1000m 

# check envvars which might override default args
if [ "$HADOOP_HEAPSIZE" != "" ]; then
  #echo "run with heapsize $HADOOP_HEAPSIZE"
  JAVA_HEAP_MAX="-Xmx""$HADOOP_HEAPSIZE""m"
  #echo $JAVA_HEAP_MAX
fi
````

The last `-Xmx` value seems to take precedence.

On HDP2.2, that particular file looks to have moved to `/usr/hdp/current/hadoop-client/libexec/hadoop-config.sh`/`/usr/hdp/2.2.0.0-2041/hadoop/libexec/hadoop-config.sh`.

amithkanand:
LGTM

PR #195, opened 2015-07-17
Make graphite whisper(data) directory configurable

kiiranh:
This PR provides ability to specify/change the location to store graphite whisper data. Whisper files can take up huge amounts of space (in 100's of GBs) based on what metrics are configured to be collected and the number of data points to be collected. With this PR, one can set the data directory to be on a separate disk/partition so that whisper data does not fill up root disk.


kiiranh:
Separated out the graphite attributes to their own file. Unit test on vbox vm cluster.

cbaenziger:
Thanks for splitting the attributes out and making this configurable!

PR #197, opened 2015-07-17
Ensure HDFS client uses host's floating ip (not vip)

amithkanand:
This PR adds `HDFS` parameters to make sure that `hdfs` client uses host's floating ip address (not vip).


amithkanand:
A new PR #198 is created.

PR #198, opened 2015-07-18
HDFS Fixes for HDFS Clients

cbaenziger:
For machines with VIP's it is possible that HDFS clients running on the machine may try to talk out the VIP -- something that will not work for getting packets back to the client. As such we need to ensure the hdfs-site.xml client options are specified. This is especially true for HBase masters. Also, at this time we do not restart HBase when hdfs-site.xml is changed; we should.


amithkanand:
LGTM

PR #199, opened 2015-07-18
For disconnected clusters, do not install ubuntu-tools repo in role[B…

cbaenziger:
For disconnected cluster bootstrap nodes, by convention, we tend to run `role[Basic]` first then run recipes later in the Chef run. One issue with having the support tool in Basic is that the `bcpc::apache-mirror` recipe will not have yet run which configures the end-point for mirrored repos.

For something as big as Canonical there needs to be an independent repo but for all other small (<10GB) repos, we usually can simply drop them into place.

This begs, we should probably have `role[Basic]` and `role[Cluster-Basic]` one being for absolutely all machines and one being only for cluster nodes.


amithkanand:
LGTM.

PR #200, opened 2015-07-21
Spark Fixes

cbaenziger:
This provides an update to how we deploy Spark with the following fixes:

**Update SPARK environment for SPARK-8332 and for LZO use**:
* Fix for [SPARK-8332](https://issues.apache.org/jira/browse/SPARK-8332) -- simply do not include Jackson JARs in the `CLASSPATH`
* Include an `LD_LIBRARY_PATH` directive for Spark to find LZO and not fatally bail out (replaced in later commit with a `spark-defaults.conf`)
* Include `SPARK_LOCAL_IP` and `SPARK_PUBLIC_DNS` to inform the machine to what it should bind and the hostname which it should advertise

**Add in Spark dynamic allocation and logging for eventual Spark history server use**:
* Create a `spark-defaults.conf` with some reasonable defaults
* Create an HDFS `/spark-history` directory for eventual Spark History Server use
* Use `spark.{driver,executor}.extraLibraryPath` for necessary Java libraries instead of `LD_LIBRARY_PATH`
* Add in some GC logging for executors
* Consolidate shuffle files since we use XFS by default
* Increase the number of connections per host to a reasonable number for our default cluster sizes

**Add in cgroups for YARN**:
* Configured per Apache Hadoop [documentation](http://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/NodeManagerCgroups.html)
* Verified by ensuring containers has entries under `/sys/fs/cgroup/cpu/hadoop-yarn/<container ID>`
* Does not set `yarn.nodemanager.linux-container-executor.cgroups.strict-resource-usage` which is useful for benchmarking or deterministically long running processes

**Enables long-running job log aggregation**:
* Set rolling log aggregation to 30 minutes
* Set deletion of old logs to 31 days

**Enables dominant resource fairness**:
* ~~Configures fair-scheduler to use `drf` instead of `fair` scheduling algorithm~~
* Sets YARN to use cgroups to keep absolute percentage of CPU usage at `node['bcpc']['hadoop']['yarn']['nodemanager']['avail_vcpu']['ratio']` while advertising desired number of vCPU's; one can still set either `node['bcpc']['hadoop']['yarn']['nodemanager']['avail_vcpu']['ratio']` and/or `node['bcpc']['hadoop']['yarn']['nodemanager']['avail_vcpu']['count']` and get working behavior calculated using `node['cpu']['total']`.

**__BREAKING__: Changes `rm_port`**:
* To normalize attributes into `nodemanager` and `resourcemanager` sections in the attributes file:
  ````
  default["bcpc"]["hadoop"]["yarn"]["rm_port"] = 8032
  ````
  Changed to **breaking** attribute:
  ````
  default["bcpc"]["hadoop"]["yarn"]["resourcemanager"]["port"] = 8032
  ````


cookbooks/bach_spark/templates/default/spark-defaults.conf.erb
```diff
@@ -0,0 +1,10 @@
+spark.driver.extraLibraryPath=/usr/hdp/current/hadoop-client/lib/native:/usr/hdp/current/hadoop-client/lib/native/Linux-amd64-64
+spark.executor.extraLibraryPath=/usr/hdp/current/hadoop-client/lib/native:/usr/hdp/current/hadoop-client/lib/native/Linux-amd64-64
+spark.executor.extraJavaOptions -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime
+spark.shuffle.consolidateFiles true
+spark.shuffle.io.numConnectionsPerPeer 2
+spark.history.fs.logDirectory hdfs://<% node.chef_environment %>/spark-history
+spark.eventLog.logDirectory hdfs://<% node.chef_environment %>/spark-history
```
amithkanand: Should this be `<%= node.chef_environment %>` instead of `<% node.chef_environment %>`?

cookbooks/bach_spark/templates/default/spark-defaults.conf.erb
```diff
@@ -0,0 +1,10 @@
+spark.driver.extraLibraryPath=/usr/hdp/current/hadoop-client/lib/native:/usr/hdp/current/hadoop-client/lib/native/Linux-amd64-64
+spark.executor.extraLibraryPath=/usr/hdp/current/hadoop-client/lib/native:/usr/hdp/current/hadoop-client/lib/native/Linux-amd64-64
+spark.executor.extraJavaOptions -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime
+spark.shuffle.consolidateFiles true
+spark.shuffle.io.numConnectionsPerPeer 2
```
DrRayWang: How do you decide this connection number?

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -111,9 +111,16 @@
 
   <property>
     <name>yarn.nodemanager.aux-services</name>
-    <value>mapreduce_shuffle</value>
+    <value>mapreduce_shuffle<% if node.run_list.expand(node.chef_environment).recipes.include?("bach_spark::default") %>,spark_shuffle<% end %></value>
   </property>
 
+  <% if node.run_list.expand(node.chef_environment).recipes.include?("bach_spark::default") %>
+  <property>
+    <name>yarn.nodemanager.aux-services.spark.shuffle.class</name>
```
amithkanand: The parameter should be `yarn.nodemanager.aux-services.spark_shuffle.class`.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -111,9 +111,16 @@
...
+    <value>org.apache.spark.network.yarn.YarnShuffleService</value>
```
amithkanand: After fixing the above parameter locally, `node manager` won't start and throws following error:

````
mapred.ShuffleHandler which has a name of 'httpshuffle'. Because these are not the same tools trying to send ServiceData and read Service Meta Data may have issues unless the refer to the name in the config.
2015-07-22 13:54:09,658 INFO org.apache.hadoop.yarn.server.nodemanager.containermanager.AuxServices: Adding auxiliary service httpshuffle, "mapreduce_shuffle"
2015-07-22 13:54:09,740 FATAL org.apache.hadoop.yarn.server.nodemanager.containermanager.AuxServices: Failed to initialize spark_shuffle
java.lang.RuntimeException: java.lang.RuntimeException: java.lang.ClassNotFoundException: Class org.apache.spark.network.yarn.YarnShuffleService not found
        at org.apache.hadoop.conf.Configuration.getClass(Configuration.java:2108)
        at org.apache.hadoop.yarn.server.nodemanager.containermanager.AuxServices.serviceInit(AuxServices.java:121)
        at org.apache.hadoop.service.AbstractService.init(AbstractService.java:163)
        at org.apache.hadoop.service.CompositeService.serviceInit(CompositeService.java:107)
        at org.apache.hadoop.yarn.server.nodemanager.containermanager.ContainerManagerImpl.serviceInit(ContainerManagerImpl.java:237)
        at org.apache.hadoop.service.AbstractService.init(AbstractService.java:163)
        at org.apache.hadoop.service.CompositeService.serviceInit(CompositeService.java:107)
        at org.apache.hadoop.yarn.server.nodemanager.NodeManager.serviceInit(NodeManager.java:253)
        at org.apache.hadoop.service.AbstractService.init(AbstractService.java:163)
        at org.apache.hadoop.yarn.server.nodemanager.NodeManager.initAndStartNodeManager(NodeManager.java:462)
        at org.apache.hadoop.yarn.server.nodemanager.NodeManager.main(NodeManager.java:509)
Caused by: java.lang.RuntimeException: java.lang.ClassNotFoundException: Class org.apache.spark.network.yarn.YarnShuffleService not found
        at org.apache.hadoop.conf.Configuration.getClass(Configuration.java:2076)
        at org.apache.hadoop.conf.Configuration.getClass(Configuration.java:2100)
        ... 10 more
Caused by: java.lang.ClassNotFoundException: Class org.apache.spark.network.yarn.YarnShuffleService not found
        at org.apache.hadoop.conf.Configuration.getClassByName(Configuration.java:1982)
        at org.apache.hadoop.conf.Configuration.getClass(Configuration.java:2074)
        ... 11 more
````

File: cookbooks/bach_spark/templates/default/spark-defaults.conf.erb (hunk 2 above)
Comment (cbaenziger) [inline cookbooks/bach_spark/templates/default/spark-defaults.conf.erb]: Wet finger in wind

**Comment by DrRayWang** [comment]
"recipe[bach_spark::default]" should be put before "recipe[bcpc-hadoop::datanode]" in the runlist of roles/BCPC-Hadoop-Worker.json, because nodemanager run in datanode recipe needs YarnShuffleService class, which is installed in spark package by "recipe[bach_spark::default]".

**Comment by cbaenziger** [comment]
This now also uploads the Spark Assembly JAR per Cloudera's YARN Optimization [recommendation](http://www.cloudera.com/content/cloudera/en/documentation/core/latest/topics/cdh_ig_running_spark_on_yarn.html#concept_nhf_yvf_qr_unique_1).

This also bumps the default version to 1.4.1.

File: cookbooks/bach_spark/recipes/default.rb [hunk 5]
```diff
@@ -11,6 +12,35 @@
...
+end
+
+bash "create-hdfs-spark-history-dir" do
+  code <<-EOH
+  hdfs dfs -mkdir -p #{hdfs_url}/spark-history/
+  hdfs dfs -chmod -R 1777 #{hdfs_url}/spark-history/
+  hdfs dfs -chown -R hdfs:hdfs #{hdfs_url}/spark-history
+  EOH
+  user "hdfs"
+  not_if "sudo -u hdfs hadoop dfs -test -d #{hdfs_url}/spark-history"
```
amithkanand: Do you need to `sudo` as `hdfs` user if `bash` resource is running as `hdfs` user?

cookbooks/bach_spark/recipes/default.rb
```diff
@@ -11,6 +12,35 @@
...
+end
+
+bash "create-hdfs-spark-assembly-jar" do
+  code <<-EOH
+  hdfs dfs -mkdir -p #{hdfs_url}/apps/spark/#{spark_pkg_version}/
+  hdfs dfs -chown -R hdfs:hdfs #{hdfs_url}/apps/spark/#{spark_pkg_version}/
+  hdfs dfs -copyFromLocal  #{spark_bin_dir}/lib/spark-assembly-*.jar #{hdfs_url}/apps/spark/#{spark_pkg_version}/spark-assembly.jar
+  EOH
+  user "hdfs"
+  not_if "sudo -u hdfs hadoop dfs -test -d #{hdfs_url}/apps/spark/#{spark_pkg_version}/"
```
amithkanand: Do you need to `sudo` as `hdfs` user if `bash` resource is running as `hdfs` user?

cookbooks/bach_spark/recipes/default.rb
```diff
@@ -11,6 +12,35 @@
...
+  code <<-EOH
+  hdfs dfs -mkdir -p #{hdfs_url}/spark-history/
+  hdfs dfs -chmod -R 1777 #{hdfs_url}/spark-history/
+  hdfs dfs -chown -R hdfs:hdfs #{hdfs_url}/spark-history
+  EOH
+  user "hdfs"
+  not_if "sudo -u hdfs hadoop dfs -test -d #{hdfs_url}/spark-history"
+end
+
+bash "create-hdfs-spark-assembly-jar" do
```
amithkanand: I got error below while chef-ing the test cluster?
````
Error executing action `run` on resource 'bash[create-hdfs-spark-assembly-jar]'
Mixlib::ShellOut::ShellCommandFailed
------------------------------------
Expected process to exit with [0], but received '1'
---- Begin output of "bash"  "/tmp/chef-script20150831-9698-1nqitbq" ----
STDOUT:
STDERR: copyFromLocal: `hdfs://Test-Laptop/apps/spark/1.4.1/spark-assembly.jar': File exists
---- End output of "bash"  "/tmp/chef-script20150831-9698-1nqitbq" ----
Ran "bash"  "/tmp/chef-script20150831-9698-1nqitbq" returned 1
````

File: cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb [hunk 8]
```diff
@@ -10,97 +10,117 @@
...
-    <value><%="#{float_host(@rm_hosts.first[:hostname])}:#{node["bcpc"]["hadoop"]["yarn"]["rm_port"]}" if not @rm_hosts.empty?%></value>
+    <name>yarn.application.classpath</name>
+    <value>/etc/hadoop/conf,/usr/hdp/current/hadoop-client/*,/usr/hdp/current/hadoop-client/lib/*,/usr/hdp/current/hadoop-hdfs-client/*,/usr/hdp/current/hadoop-hdfs-client/lib/*,/usr/hdp/current/hadoop-yarn-client/*,/usr/hdp/current/hadoop-yarn-client/lib/*</value>
   </property>
 
   <property>
-    <name>yarn.resourcemanager.scheduler.address</name>
-    <value><%="#{float_host(@rm_hosts.first[:hostname])}:8030" if not @rm_hosts.empty?%></value>
+    <name>yarn.nodemanager.aux-services</name>
+    <value>mapreduce_shuffle<% if node.run_list.expand(node.chef_environment).recipes.include?("bach_spark::default") %>,spark_shuffle<% end %></value>
```
amithkanand: Will it be a better idea to move all the `aux-services` into an `array` in a node attribute and define corresponding classes for each `aux-service` in a service specific attribute? Something like:
````
default[:bcpc][:hadoop][:yarn][:aux-services] = ["mapreduce_shuffle", "spark_shuffle"]
````
Then have class specific attributes:
````
default[:bcpc][:hadoop][:yarn][:aux-services][:mapreduce_shuffle][:class] = "org.apache.hadoop.mapred.ShuffleHandler"
default[:bcpc][:hadoop][:yarn][:aux-services][:spark_shuffle][:class] = "org.apache.spark.network.yarn.YarnShuffleService"

````
Then we can loop and generate configuration and we do not have to inspect the chef run_list.

File: roles/BCPC-Hadoop-Worker.json [hunk 9]
```diff
@@ -12,11 +12,11 @@
     "recipe[bcpc-hadoop::hdp_repo]",
     "recipe[bcpc-hadoop::configs]",
     "recipe[pam::default]",
+    "recipe[bach_spark::default]",
```
amithkanand: Why change the order here?

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -10,97 +10,117 @@
...
+   
+    <property> 
+      <name>yarn.client.failover-proxy-provider</name> 
+      <value>org.apache.hadoop.yarn.client.ConfiguredRMFailoverProxyProvider</value> 
+    </property>
+    <% else %>
+
+    <property>
+      <name>yarn.resourcemanager.resource-tracker.address</name>
+      <value><%="#{float_host(@rm_hosts.first[:hostname])}:8031" if not @rm_hosts.empty?%></value>
```
amithkanand: Why not use an attribute for port here? It is done on line 70 for `ResourceManager` port.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -10,97 +10,117 @@
...
+    </property>
+
+    <property>
+      <name>yarn.resourcemanager.address</name>
+      <value><%="#{float_host(@rm_hosts.first[:hostname])}:#{node["bcpc"]["hadoop"]["yarn"]["resourcemanager"]["port"]}" if not @rm_hosts.empty?%></value>
+    </property>
+
+    <property>
+      <name>yarn.resourcemanager.scheduler.address</name>
+      <value><%="#{float_host(@rm_hosts.first[:hostname])}:8030" if not @rm_hosts.empty?%></value>
```
amithkanand: Why not use an attribute for port here? It is done on line 70 for `ResourceManager` port.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -10,97 +10,117 @@
...
+    </property>
+
+    <property>
+      <name>yarn.resourcemanager.admin.address</name>
+      <value><%="#{float_host(@rm_hosts.first[:hostname])}:8033" if not @rm_hosts.empty?%></value>
```
amithkanand: Why not use an attribute for port here? It is done on line 70 for `ResourceManager` port.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -10,97 +10,117 @@
...
+    </property>
+
+    <property>
+      <name>yarn.resourcemanager.webapp.address</name>
+      <value><%="#{float_host(@rm_hosts.first[:hostname])}:8088" if not @rm_hosts.empty?%></value>
```
amithkanand: Why not use an attribute for port here? It is done on line 70 for `ResourceManager` port.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
cbaenziger: I fixed the classpath in `yarn-env.sh` to include the requisite JARs.

cookbooks/bach_spark/recipes/default.rb
cbaenziger: You do need to run the guard attribute as hdfs but that's now been done to remove the sudo's/

cookbooks/bach_spark/recipes/default.rb
cbaenziger: Similarly fixed.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
cbaenziger: As I've only changed indentation here I did not parameterize our various ports; but at some point it would be good to rationalize these indeed.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
cbaenziger: Again, this is still a YARN default but not modified here, only indentation changed. At some point, a more thorough rationalization would be good (particularly for webapp ports to go through HAProxy for user ease with a single VIP-based URL would be nice).

roles/BCPC-Hadoop-Worker.json
cbaenziger: Good call on a fresh cluster, HDFS would be down to upload the JAR's and yeah no hadoop software even yet; I wonder how to keep from racing as the nodemanager needs the Spark shuffle service but we need to install that first -- the HDFS upload of JARs could be done later.

cbaenziger:
Ah yes, I somehow dropped the `yarn.nodemanager`; thanks!

cookbooks/bach_spark/templates/default/spark-defaults.conf.erb
```diff
@@ -0,0 +1,14 @@
...
+spark.history.fs.logDirectory <%= node['spark']['hdfs_url'] %>/spark-history
+spark.eventLog.dir <%= node['spark']['hdfs_url'] %>/spark-history
+spark.eventLog.enabled true
+spark.logConf true
+spark.dynamicAllocation.enabled true
```
DrRayWang: One concern here is that if we do not set "spark.dynamicAllocation.maxExecutors" with "spark.dynamicAllocation.enabled", whether or not some big application will eat up the cluster resource.  In my example, my application by default is executed in several shffleMapStages, each stage is made of different number of tasks by default (I think use as much resource as it can). And each task can be assigned to different executors. If one stage has a lot of tasks that invoke the same huge amount of executors in the cluster, this might uses up the resource of cluster, and make other applications fall into dead lock.  Is this concern valid?

cbaenziger: Hi Ray, the thought here is that the YARN scheduler should be limiting the resources a user can consume more than us limiting it at the Spark layer which is just one of the many higher-level languages which could be in use. (E.g. for traditional Map/Reduce a job could consume a large number of containers for maps on a large input or a large number of containers for a large reduce while otherwise being well bounded in its resource demands.)

cbaenziger:
This now introduces `bach_spark::cluster_install` recipe which requires Hadoop services such as HDFS be available. Further, a node attribute is used for YARN auxiliary services so that Spark can inject itself once installed but before the nodemanager starts.

cookbooks/bach_spark/templates/default/spark-defaults.conf.erb
```diff
@@ -0,0 +1,14 @@
+spark.driver.extraLibraryPath=/usr/hdp/current/hadoop-client/lib/native:/usr/hdp/current/hadoop-client/lib/native/Linux-amd64-64
+spark.executor.extraLibraryPath=/usr/hdp/current/hadoop-client/lib/native:/usr/hdp/current/hadoop-client/lib/native/Linux-amd64-64
+spark.executor.extraJavaOptions -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime
+spark.shuffle.consolidateFiles true
+spark.shuffle.io.numConnectionsPerPeer 2
+spark.history.fs.logDirectory <%= node['spark']['hdfs_url'] %>/spark-history
+spark.eventLog.dir <%= node['spark']['hdfs_url'] %>/spark-history
+spark.eventLog.enabled true
```
DrRayWang: Even eventLog is set, I still hit java.io.FileNotFoundException: File file:/tmp/spark-events does not exist Error. This results in SparkContext did not initialize after waiting for *****ms.

http-418:
I'm going to re-base and build this branch today.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
kiiranh: The indentation change would take effect in the resultant xml too; even though the elements are at the same level, they'll be indented.

kiiranh:
###### Tests
* Built two VM clusters with this branch
** one medium cluster: with 5 nodes (3 worker nodes, each w/ 5G mem and 4 vcores) 
** other small cluster: with 3 nodes (1 worker node w/ 5G mem and 4 vcores)
* On the medium cluster, SparkPi example worked fine with default allocation of memory and cores.
** Command used: /bin/spark-submit --class org.apache.spark.examples.SparkPi lib/spark-examples-1.4.1-hadoop2.6.0.jar 100 
* On the small cluster:
** Because of limited resources, ran the following command: /bin/spark-submit --class org.apache.spark.examples.SparkPi --driver-memory 512m --executor-memory 512m --executor-cores 1 lib/spark-examples-1.4.1-hadoop2.6.0.jar  25
** Used the Unix `stress` program to stress the CPU cores while running the SparkPi example; tried stressing 3 cores and 4 cores and the spark command succeeds.
** While `stressing` the CPU cores, also executed a couple sample PySpark programs which succeeded.
* Verified, when the spark job is running, there is a folder for the container under `/sys/fs/cgroup/cpu/hadoop-yarn/`

P.S. These tests were using the `fair` scheduling algorithm instead of the `drf` algorithm.

kiiranh:
PR still has the `drf` commit.

DrRayWang:
I don't think "Unix stress program" will has the right stress influence we want to see on a spark job. You may try to use, for example, distributed shell to make the cluster run a pig job, or another spark job to give pressure to both CPU and memory resource available on the cluster.

kiiranh:
The only comment not addressed is the one where the xml tags are indented in the template code for `cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb`. The resultant xml will also have this indentation. At this point, code readability is given preference rather than xml readability. A coding practice can be decided and applied later.

kiiranh:
LGTM

cbaenziger:
@kiiranh As to your comment
````
The resultant xml will also have this indentation.
````

I have unindented all XML and instead commented in the erb the loop or conditional. I am thinking going forward it might be best to make the XML ERB code readable and then extend the Chef template provider to run the XML through a pretty-printer (e.g. `xmllint --format`) to make the XML readable. But that's some idea for the future.

I verified the erb comments worked with ` erb -P -x -T '-' cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb | ruby -c`.

kiiranh:
Thanks @cbaenziger. I guess this is good to mainstream

PR #204, opened 2015-07-22
Don't print to STDOUT in Vagrantfile.

http-418:
The existing branch breaks Jenkins builds by including `puts` statements in VagrantFile.  When `vagrant ssh-config` is run, the output isn't a valid SSH config.  e.g.

```
+ scp -rF ssh_config 'bootstrap:~/chef-bcpc/bins' .
ssh_config: line 1: Bad configuration option: base
ssh_config: terminating, 1 bad configuration options
Build step 'Execute shell' marked build as failure
```

This PR replaces `puts` calls with `$stderr.puts`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -17,17 +17,17 @@ else
   base_dir = File.expand_path("./environments")
 end
 
-puts "Base directory is : #{base_dir}"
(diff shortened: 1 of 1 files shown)


pu239ppy:
Looks good

PR #206, opened 2015-07-24
Fix standby namenode restart.

http-418:
This commit prevents a restart of the standby namenode.   If the restart is executed on a fresh cluster, the standby namenode halts with an error.


amithkanand:
Can commit message be little descriptive?

cbaenziger:
Ray's pointed out moving this to a subscription would be better and allow for cluster upgrades to bounce the secondary namenode.

PR #207, opened 2015-07-24
Install required package for Zabbix compile/build

amithkanand:
This PR fixes current build process where `build_bins.sh` fails to compile `Zabbix` from source as it is missing `pkg-config` package that is required for the build.


kiiranh:
LGTM

PR #209, opened 2015-07-27
fix standby namenode restart patch

DrRayWang:
restart hdfs namenode standby by subscribing to creation of directory "/var/log/hadoop-hdfs/gc/" when hdfs HA is on.


kiiranh:
LGTM

PR #211, opened 2015-07-27
Tune memory settings for Hadoop components(based on VM memory size)

amithkanand:
Our current implementation uses default memory allocation of `1024MB` for various hadoop components. Since we currently build VM cluster with either `2048MB` or `4096MB` of memory, using default values for hadoop componentes leaves almost no memory `Yarn` containers to launch and run a job. This PR uses dynamic memory calculations based on total memory available on the node for daemons mentioned below.
- `HBase Regionserver`
- `HBase Master`
- `Yarn NodeManager`
- `Yarn ResourceManager`
  After tuning these components one should be able to successfully run jobs on the cluster.


http-418:
The good news is that this did not break my build.  I was able to build a cluster successfully. 

The bad news is that these changes did not allow containers to run in the default configuration.  Jobs enter "SCHEDULED" state, but they never run.

amithkanand:
I am still testing this. This needs more changes.

pu239ppy:
You need to make adjustments in mapred-site based on minimum allocation as well

bijugs:
@amithkanand , Looks like we may already have these changes in the code. Do we still need this PR?

amithkanand:
Not needed anymore.

PR #214, opened 2015-07-28
Pin windows cookbook to 1.36.6

http-418:
Versions 1.37.0 and later contain metadata.rb that is incompatible
with our release of chef 11.x.

Builds fail in the chef bootstrap process:
```
ERROR: knife encountered an unexpected error
This may be a bug in the 'cookbook upload' knife command or plugin
Please collect the output of this command with the `-VV` option before filing a bug report.
Exception: NoMethodError: undefined method `source_url' for #<Chef::Cookbook::Metadata:0x000000034cf448>
+ echo '############## bootstrap_chef.sh returned 100 ##############'
############## bootstrap_chef.sh returned 100 ##############
+ exit 1
```

This commit pins the `windows` cookbook at the last version with compatible metadata.

### DIFF
--- a/setup_chef_cookbooks.sh
+++ b/setup_chef_cookbooks.sh
@@ -56,7 +56,7 @@ EOF
 cd cookbooks
 
 # allow versions on cookbooks via "cookbook version"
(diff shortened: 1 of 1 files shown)


pu239ppy:
This looks good

PR #215, opened 2015-07-28
Move gpxe ROM into source tree.

http-418:
Our current build process relies on rom-o-matic being up.  In this branch, I've copied the ROM directly into the source tree in order to remove that dependency.


pu239ppy:
This looks good, I don't think adding a ROM is a problem

PR #218, opened 2015-07-30
Change IP to CIDR for dfs.client.local.interfaces. 

amithkanand:
This PR is an addendum to #198 and adds `CIDR` notation to `HDFS` parameter `dfs.client.local.interfaces`. This is required as passing just an `IP` address assumes `CIDR` to be `<IP Address>/0` and breaks the cluster.


pu239ppy:
Change looks good

PR #220, opened 2015-08-03
Generic HDFS/Yarn Issues

amithkanand:
This PR provides following:
* Tune `NodeManager` memory to use 50% of machine resources
* Bind `YARN` services to correct interface/IP (floating) address.
* Correct path for `HDP` in `oozie` recipe to use current version instead of a specific version.


cbaenziger:
BREAKING change: If one has overridden `node["bcpc"]["hadoop"]["yarn"]["scheduler"]` in the past, this will cause that setting to fail; one needs to move the scheduler class string to `["bcpc"]["hadoop"]["yarn"]["scheduler"]["class"]`. If one has set this attribute, one will likely end up will `nil` in their `yarn-site.xml` for `yarn.scheduler.minimum-allocation-mb` and `yarn.resourcemanager.scheduler.class`.

PR #208, opened 2015-07-27
Ensure hdfs and yarn users are in appropriate hdfs and hadoop groups

cbaenziger:
This provides group membership for machines which already have hdfs and yarn users but not th necessary group membership for proper cluster operation.

**HDFS**: Needs to be in group hadoop (for proper updating of /var/log/); and should be in its self-named group for sanity's safe
**YARN**: Needs to be in group hadoop (for proper operation of the Linux Container Executor)

If one does not have this patch, one will see the datanode fail to start with errors in `/var/log/hadoop-hdfs/jsvc.err` like:
````
Initializing secure datanode resources
log4j:ERROR setFile(null,true) call failed.
java.io.FileNotFoundException: /var/log/hadoop-hdfs/hadoop-hdfs-datanode-hostname.log (Permission denied)
[...]
Opened info server at hostname/1.2.3.4:1006
Starting regular datanode initialization
Service exit with a return value of 1
````

PR #223, opened 2015-08-03
user hcat is already created in datanode

pu239ppy:
Removing from hive.rb


cbaenziger:
This should be taken care of in #208; using the `user_exists?` function created in #117 as a guard. Also, on headnodes we do not run the datanode recipe so this would be a bit dangerous there. On non-SSO'd clusters this user action is okay but SSO'd clusters can throw issues.

PR #226, opened 2015-08-06
java configuration fix for kafka cluster

DrRayWang:
Fixed Java configuration on kafka cluster for both zookeeper nodes and kafka server nodes


cbaenziger:
Thanks for catching this omission.

PR #227, opened 2015-08-06
zookeeper home directory configuration fix

DrRayWang:
Correct zookeeper path configuration in recipe - bcpc-hadoop::zookeeper_impl.rb


bijugs:
The changes in this PR is already in the latest code. Labeling it as invalid. @cbaenziger could you please have a second look so that we can close it.

cbaenziger:
@bijugs thanks for marking this. I agree these are all fixed up these days.

PR #229, opened 2015-08-07
Prevent RM from restarting every chef-run. Restart YARN on env change

amithkanand:
This PR fixes following issues:
* Prevents `YARN ResourceManager` from restarting on every chef-run by removing duplicate template resource
* Subscribes `ResourceManager` and `NodeManager` to restart if `yarn-env.sh` changes. 
* Correct the value for `aux` service class in `yarn-site.xml` from `mapreduce.shuffle` to mapreduce_shuffle` to match with what is defined in `yarn.nodemanager.aux-services`

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/hadoop_config.rb
+++ b/cookbooks/bcpc-hadoop/recipes/hadoop_config.rb
@@ -55,7 +55,7 @@ end
   hadoop-env.sh}.each do |t|
  template "/etc/hadoop/conf/#{t}" do
    source "hdp_#{t}.erb"
-   mode 0644
+   mode 0555
    variables(:nn_hosts => node[:bcpc][:hadoop][:nn_hosts],
              :zk_hosts => node[:bcpc][:hadoop][:zookeeper][:servers],
              :jn_hosts => node[:bcpc][:hadoop][:jn_hosts],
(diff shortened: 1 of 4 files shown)


cookbooks/bcpc-hadoop/recipes/hadoop_config.rb
```diff
@@ -55,7 +55,7 @@
   hadoop-env.sh}.each do |t|
  template "/etc/hadoop/conf/#{t}" do
    source "hdp_#{t}.erb"
-   mode 0644
+   mode 0744
```
cbaenziger: Mode `555` would be preferable here.

cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
```diff
@@ -115,7 +115,7 @@
   </property>
 
   <property>
-    <name>yarn.nodemanager.aux-services.mapreduce.shuffle.class</name>
+    <name>yarn.nodemanager.aux-services.mapreduce_shuffle.class</name>
```
cbaenziger: Ah yes, [YARN-1229](https://issues.apache.org/jira/browse/YARN-1229) made `.` illegal.

PR #212, opened 2015-07-27
chef vault haproxy stats secret

DrRayWang:
Include:
* Find all chef vault resources that will deploy on a node
* grant the node access to the vault items
* vault haproxy stats password

PR #231, opened 2015-08-13
Vault keepalived password

DrRayWang:
Chef vault keepalived password and made related changes in recipes and template files

### DIFF
--- a/cookbooks/bcpc/recipes/keepalived.rb
+++ b/cookbooks/bcpc/recipes/keepalived.rb
@@ -20,7 +20,23 @@
 include_recipe "bcpc::default"
 
 make_config('keepalived-router-id', "#{(rand * 1000).to_i%254/2*2+1}")
-make_config('keepalived-password', secure_password)
+
+keepalived_password = get_config('keepalived-password')
+if keepalived_password.nil?
+  keepalived_password = secure_password
+end
+
+bootstrap = get_bootstrap
+results = get_nodes_for("keepalived").map!{ |x| x['fqdn'] }.join(",")
+nodes = results == "" ? node['fqdn'] : results
+
+chef_vault_secret "keepalived" do
+  data_bag 'os'
+  raw_data({ 'password' => keepalived_password })
+  admins "#{ nodes },#{ bootstrap }"
+  search '*:*'
+  action :nothing
+end.run_action(:create_if_missing)
 
 package "keepalived" do
     action :upgrade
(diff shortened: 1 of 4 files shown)


cbaenziger:
This will need a fix for Chef-BACH #244

cookbooks/bcpc/recipes/keepalived.rb
```diff
@@ -20,7 +20,23 @@
...
 
 make_config('keepalived-router-id', "#{(rand * 1000).to_i%254/2*2+1}")
-make_config('keepalived-password', secure_password)
+
+keepalived_password = get_config('keepalived-password')
+if keepalived_password.nil?
+  keepalived_password = secure_password
+end
+
+bootstrap = get_all_nodes.select{|s| s.hostname.include? 'bootstrap'}[0].fqdn
```
cbaenziger: For performance, instead of doing a `get_all_nodes` for the bootstrap 7+ times (once each recipe); perhaps create a cache? Either a memoized helper function or an "ephemeral" node attribute?

cookbooks/bcpc/recipes/keepalived.rb
```diff
@@ -20,7 +20,23 @@
...
+
+nodes = get_nodes_for("keepalived").map!{ |x| x['fqdn'] }.join(",")
```
cbaenziger: Instead of doing a `map!` on `get_nodes_for`, please use the node attribute function [`get_node_attributes`](https://github.com/bloomberg/chef-bach/blob/94b5c20e7ed967e111daae6c05d15cb5642e381a/cookbooks/bcpc/libraries/utils.rb#L176-L180).

cbaenziger:
For all chef-vault PR's please specify their dependencies to help future code spelunkers; particularly reference the changes to `cluster-assign-roles.sh` which allows one to still create a cluster from scratch without repeated Chef converges.

cbaenziger:
Please also secure the permissions for [`keepalived.conf`](https://github.com/bloomberg/chef-bach/blob/b4f3e2483cf10bfc586d448ca7dc2aa4aaee0851/cookbooks/bcpc/recipes/keepalived.rb#L31) to be more restrictive than 644.

cbaenziger:
Please also look into setting the [`sensitive`](https://docs.chef.io/resource_template.html) value for the various template resources.

cookbooks/bcpc/recipes/keepalived.rb
DrRayWang: get_node_attributes returns a array of hash with the requested attributes. It needs more logic to extract what I need from the hash array and reassemble them into a string. Extra-complexity will be added. So simple map() seems not a bad idea.

DrRayWang:
This PR depends on PR#212 (https://github.com/bloomberg/chef-bach/pull/212). PR#212 provide the resource finding on nodes and chef vault item granting procedures running in cluster-assign-roles.sh

amithkanand:
LGTM. Tested on VM cluster.

PR #233, opened 2015-08-14
Lower oozie memory requirements for VM clusters

kiiranh:
PR #121 raised the default memory allocation for oozie server process to 2G heap and 256M perm. This allocation is not viable on VM clusters with about 3-5G total memory. This PR overrides the memory settings and set them to half for VM clusters.

### DIFF
--- a/environments/Test-Laptop.json
+++ b/environments/Test-Laptop.json
@@ -46,6 +46,9 @@
         "distribution" : {
           "version" : "HDP",
           "key" : "hortonworks.key"
+        },
+        "oozie": {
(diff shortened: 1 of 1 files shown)


cbaenziger:
I'd like to see us test if 512MB will still meaningfully run Oozie (i.e. for HDFSDU). If so, reclaiming 1/2GB more on the VM headnodes would be appreciable.

kiiranh:
Initial tests show hdfsdu is running fine on Oozie with 512M heap and 64M perm for oozie. I'll let it run some more to make sure all functions work OK.

kiiranh:
This is good to review/merge.

PR #240, opened 2015-08-19
Use logrotate to rotate chef-server's nginx access.log

pu239ppy:


pu239ppy:
this is related to #242

cbaenziger:
Ah good to see us finally use the `logrotate` cookbook for something. It looks like Chef is not fixing this until Chef 12:
* Chef/Chef#2001
* [Chef-4252](https://tickets.opscode.com/browse/CHEF-4252)

PR #241, opened 2015-08-19
Changes in hbase init.d script for BIGTOP-1956

bijugs:
Each RS on a host need to have a unique ``hbase.tmp.dir" in a multi RS environment so that there are no conflicts when an RS comes up. The change is to pass a unique tmp directory when bringing up each RS instance.

### DIFF
--- a/cookbooks/bcpc-hadoop/templates/default/hdp_hbase-regionserver-initd.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/hdp_hbase-regionserver-initd.erb
@@ -194,18 +194,26 @@ multi_hbase_daemon() {
         exit 1
     fi
 
-    export HBASE_${UPPERCASE_HBASE_DAEMON}_OPTS=" "
-
     for OFFSET in ${OFFSETS} ; do
+        export HBASE_${UPPERCASE_HBASE_DAEMON}_OPTS=" "
         echo -n "$ACTION regionserver daemon $OFFSET: "
         export HBASE_IDENT_STRING="hbase-${OFFSET}"
         LOG_FILE="$HBASE_LOG_DIR/hbase-$HBASE_IDENT_STRING-regionserver-$HOSTNAME.pid"
         PID_FILE="$HBASE_PID_DIR/hbase-$HBASE_IDENT_STRING-regionserver.pid"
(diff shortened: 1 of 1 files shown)


cookbooks/bcpc-hadoop/templates/default/hdp_hbase-regionserver-initd.erb
```diff
@@ -194,18 +194,26 @@ multi_hbase_daemon() {
         exit 1
     fi
 
-    export HBASE_${UPPERCASE_HBASE_DAEMON}_OPTS=" "
-
     for OFFSET in ${OFFSETS} ; do
+        export HBASE_${UPPERCASE_HBASE_DAEMON}_OPTS=" "
```
cbaenziger: Why even use `$HBASE_${UPPERCASE_HBASE_DAEMON}_OPTS` if it is just set to empty and not settable outside this script?

bijugs: This is from the original BIGTOP script. Wanted to make minimal change not knowing the history and also knowing that we will be using HDP provided scripts.

cookbooks/bcpc-hadoop/templates/default/hdp_hbase-regionserver-initd.erb
```diff
@@ -194,18 +194,26 @@ multi_hbase_daemon() {
...
         export HBASE_IDENT_STRING="hbase-${OFFSET}"
         LOG_FILE="$HBASE_LOG_DIR/hbase-$HBASE_IDENT_STRING-regionserver-$HOSTNAME.pid"
         PID_FILE="$HBASE_PID_DIR/hbase-$HBASE_IDENT_STRING-regionserver.pid"
         HBASE_MULTI_ARGS="-D hbase.regionserver.port=`expr ${FIRST_PORT} + $OFFSET` \
                           -D hbase.regionserver.info.port=`expr ${FIRST_INFO_PORT} + ${OFFSET}`"
+
+        if [ "x$JAVA_TMP_DIR" == "x" ] ; then
+            JAVA_TMP_DIR="/tmp/java_tmp_dir"
+        fi
+        HBASE_TMP_DIR=" -Djava.io.tmpdir=${JAVA_TMP_DIR}/${OFFSET}"
```
cbaenziger: Confirmed with Biju that HBase will create the directory if it does not exist.

Issue #205, opened 2015-07-24
Set ULIMIT to a higher value for root user


amithkanand:
In case of secure `datanode`, the process is first started as `root` followed by a child process (actual datanode process). The start of secure `datanode` is handled by `JSVC` program that doesn't pass actual `ULIMIT` value to the child process while launching it. 
We should set `ULIMIT` to `32K` for the `root` user.

cbaenziger:
I would prefer to modify the datanode startup script to run `ulimit -n 32768` before calling the Bigtop JSVC program which launches the daemon.

amithkanand:
Closing as PR #243 is merged.

PR #243, opened 2015-08-20
Set ULIMIT to a higher value for root user

amithkanand:
This PR fixes issue #205 that is required for Secure DN to pickup correct ULIMIT value.


cbaenziger:
This looks fine

PR #245, opened 2015-08-20
Make chef-vaulting backward compatible

DrRayWang:
fix the issue resulting from rechefing the old existed cluster without using C-A-R.sh after pulling in new change to chef vaulting haproxy-stats password.

### DIFF
--- a/cookbooks/bcpc/recipes/haproxy.rb
+++ b/cookbooks/bcpc/recipes/haproxy.rb
@@ -29,10 +29,12 @@ end
 
 bootstrap = get_all_nodes.select{|s| s.hostname.include? 'bootstrap'}[0].fqdn
 
+nodes = get_nodes_for("haproxy").map!{ |x| x['fqdn'] }.join(",")
+
 chef_vault_secret "haproxy-stats" do
   data_bag 'os'
   raw_data({ 'password' => haproxy_stats_password })
-  admins "#{ node['fqdn'] },#{ bootstrap }"
+  admins "#{ nodes },#{ bootstrap }"
   search '*:*'
   action :nothing
 end.run_action(:create_if_missing)


amithkanand:
LGTM. I have tested this and seems to work as expected. This is what I did:

1. De-Admin all head nodes
2. Delete existing haproxy vault
3. Run `chef-clien`t on 1st Head node and expect a failure
4. Change one of the head node to be an admin node
5. Run chef-client on admin head node (This created vault and made all other head nodes admin on vaulted item)
6. De-Admin head node again
7. Run `chef-client` on all other head nodes

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -117,6 +117,16 @@ def get_all_nodes
 	return results.sort
 end
 
+def get_recipe_nodes(recipe_name)
```
cbaenziger: Can we use [`get_nodes_for`](https://github.com/bloomberg/chef-bach/blob/f1ebce5ef72e189f748b04ef231b396ffe969f0a/cookbooks/bcpc-hadoop/libraries/utils.rb#L125-L132) instead?

cookbooks/bcpc/recipes/haproxy.rb
```diff
@@ -29,10 +29,12 @@
 
 bootstrap = get_all_nodes.select{|s| s.hostname.include? 'bootstrap'}[0].fqdn
 
+nodes = get_recipe_nodes("haproxy")
+
 chef_vault_secret "haproxy-stats" do
-  data_bag 'os'
+  data_bag 'os'
```
cbaenziger: Nit: An errant space

amithkanand:
LGTM.

PR #246, opened 2015-08-21
Enable short circuit reads for HBase

amithkanand:
This PR enables `short circuit` reads for `HBase`.


cbaenziger:
Looks reasonable

PR #247, opened 2015-08-21
chef vault mysql pdns password

DrRayWang:
Chef vault mysql-pdns-password with related changes


cbaenziger:
It would be good to ensure `pdns.local.gmysql.erb` is secure from casual viewing on the system.

amithkanand:

PR #248, opened 2015-08-21
chef vault mysql zabbix password

DrRayWang:
Chef vault mysql-zabbix-password and related changes


cbaenziger:
All the various PHP files with passwords should have their permissions set so that casual users on the system can not view the Zabbix passwords.

amithkanand:

PR #249, opened 2015-08-21
chef vault zabbix admin and guest password

DrRayWang:
Chef vault zabbix-admin/guest-password and related changes


amithkanand:

Issue #52, opened 2014-12-17
Zookeeper connections are precious and other ZK administrivia


cbaenziger:
If one is running a large cluster and does not scale `node[:bcpc][:hadoop][:zookeeper][:maxClientCnxns]` then they will have the joys of Zookeeper unavailability and spewing log entries like:

```
2014-12-16 17:11:38,921 [myid:12] - WARN  [NIOServerCxn.Factory:0.0.0.0/0.0.0.0:2181:NIOServerCnxnFactory@193] - Too many connections from /1.2.3.4 - max is 500
```

Ideally, we can create monitoring in Zabbix from the provided JMX metrics:
`org.apache.ZooKeeperService` -> `ReplicatedServer_id`_nodeNumber_ -> `replica.`_nodeNumber_ -> `Attributes`-> `Follower` ->
- `Attributes` ->
  - `PendingRevalidationCount`
  - `AvgRequestLatency`
  - `MaxRequestLatency`
  - `MinRequestLatency`
  - `NumAliveConnections`
  - `OutstandingRequests`
  - `PacketsReceived`
  - `PacketsSent`
- `Connections` ->
  - _client IP_ ->
    - _connection ptr_ ->
      - `OutstandingRequest`
      - `PacketsReceived`
      - `PacketsSent`
      - `MinRequestLatency`
      - `MaxRequestLatency`
      - `AvgRequestLatency`
      - `LastLatency`
  - ...
- `InMemoryDataTree` ->
  - `Attributes` ->
    - `NodeCount`
    - `WatchCount`

The question is how to see rejected connections which I'm not seeing here. Regardless I think a lot of useful cluster monitoring can be done here.

PR #251, opened 2015-08-25
Change to regularly clean ZooKeeper snapshot files

bijugs:
The change will help balance between disk space use and the number of ZK snapshots for data recovery.


cbaenziger:
Thanks for this partial fix of Chef-BACH #52.

PR #255, opened 2015-08-27
Clean-up namenode and datanode memory parameters - provide new, perm …

cbaenziger:
This provides:
* cleans up DN and NN Java opts to not be under the JMX conditional
* provides GC options for DN as was available for the NN
* sets new gen size to 1/8th the size of the heap up to 3GB

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/default.rb
+++ b/cookbooks/bcpc-hadoop/attributes/default.rb
@@ -33,8 +33,9 @@ default["bcpc"]["hadoop"]["datanode"]["xmx"]["max_size"] = 4096
 default["bcpc"]["hadoop"]["datanode"]["xmx"]["max_ratio"] = 0.25
 default["bcpc"]["hadoop"]["datanode"]["max"]["xferthreads"] = 16384
 default["bcpc"]["hadoop"]["datanode"]["jmx"]["port"] = 10112
+default["bcpc"]["hadoop"]["datanode"]["gc_opts"] = "-server -XX:ParallelGCThreads=4 -XX:+UseParNewGC -XX:+UseConcMarkSweepGC -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -Xloggc:/var/log/hadoop-hdfs/gc/gc.log-datanode-$$-$(hostname)-$(date +'%Y%m%d%H%M').log -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime"
 default["bcpc"]["hadoop"]["namenode"]["handler"]["count"] = 100
-default["bcpc"]["hadoop"]["namenode"]["gc_opts"] = "-XX:ParallelGCThreads=14 -XX:+UseParNewGC -XX:+UseConcMarkSweepGC -XX:CMSInitiatingOccupancyFraction=70 -XX:+UseCMSInitiatingOccupancyOnly -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -Xloggc:/var/log/hadoop-hdfs/gc/gc.log-$$-$(hostname)-$(date +'%Y%m%d%H%M').log -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime"
+default["bcpc"]["hadoop"]["namenode"]["gc_opts"] = "-server -XX:ParallelGCThreads=14 -XX:+UseParNewGC -XX:+UseConcMarkSweepGC -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -Xloggc:/var/log/hadoop-hdfs/gc/gc.log-namenode-$$-$(hostname)-$(date +'%Y%m%d%H%M').log -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime"
 default["bcpc"]["hadoop"]["namenode"]["xmx"]["max_size"] = 16384
 default["bcpc"]["hadoop"]["namenode"]["xmx"]["max_ratio"] = 0.25
 default["bcpc"]["hadoop"]["namenode"]["jmx"]["port"] = 10111
--- a/cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb
@@ -48,8 +48,20 @@ export JSVC_HOME=/usr/lib/bigtop-utils
 # On secure datanodes, user to run the datanode as after dropping privileges
 export HADOOP_SECURE_DN_USER=hdfs
 
+<%
+   datanode_heap = [node["bcpc"]["hadoop"]["datanode"]["xmx"]["max_size"], (node["memory"]["total"].to_i * node["bcpc"]["hadoop"]["datanode"]["xmx"]["max_ratio"]/1024).floor].min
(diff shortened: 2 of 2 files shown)


cookbooks/bcpc-hadoop/attributes/default.rb
```diff
@@ -33,8 +33,9 @@
 default["bcpc"]["hadoop"]["datanode"]["xmx"]["max_ratio"] = 0.25
 default["bcpc"]["hadoop"]["datanode"]["max"]["xferthreads"] = 16384
 default["bcpc"]["hadoop"]["datanode"]["jmx"]["port"] = 10112
+default["bcpc"]["hadoop"]["datanode"]["gc_opts"] = "-server -XX:ParallelGCThreads=4 -XX:+UseParNewGC -XX:+UseConcMarkSweepGC -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -Xloggc:/var/log/hadoop-hdfs/gc/gc.log-datanode-$$-$(hostname)-$(date +'%Y%m%d%H%M').log -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime"
```
bijugs: Can we have ``ParallelGCThreads`` set to higher value may be 14 as in NN.

pu239ppy:
Looks good

cookbooks/bcpc-hadoop/attributes/default.rb
cbaenziger: I'd expect 14 to be overkill for both since new gen should be limited to 3GB. (Originally I had copied from our HBase testing configs.) Four for the DN was Hortonwork's recommendation. Also, we likely don't want to go too arbitrarily wide to keep from resource starvation with YARN being told to go quite wide and potentially HBase RSes on the same machine.

cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb
```diff
@@ -48,8 +48,16 @@ export JSVC_HOME=/usr/lib/bigtop-utils
 # On secure datanodes, user to run the datanode as after dropping privileges
 export HADOOP_SECURE_DN_USER=hdfs
 
+<% datanode_heap = [node["bcpc"]["hadoop"]["datanode"]["xmx"]["max_size"], (node["memory"]["total"].to_i * node["bcpc"]["hadoop"]["datanode"]["xmx"]["max_ratio"]/1024).floor].min %>
+<% datanode_newsize = [(0.125*datanode_heap).ceil, 3072].min %>
+<% namenode_heap = [node["bcpc"]["hadoop"]["namenode"]["xmx"]["max_size"], (node["memory"]["total"].to_i * node["bcpc"]["hadoop"]["namenode"]["xmx"]["max_ratio"]/1024).floor].min %>
+<% namenode_newsize = [(0.125*namenode_heap).ceil, 3072].min %>
+
+export HADOOP_DATANODE_OPTS="$HADOOP_DATANODE_OPTS -Xms<%= datanode_heap %>m -Xmx<%= datanode_heap %>m -XX:NewSize=<%= datanode_newsize %>m -XX:MaxNewSize=<%= datanode_newsize %>m -XX:PermSize=128m -XX:MaxPermSize=256m <%= node["bcpc"]["hadoop"]["datanode"]["gc_opts"] %>"
+export HADOOP_NAMENODE_OPTS="$HADOOP_NAMENODE_OPTS -Xms<%= namenode_heap %>m -Xmx<%= namenode_heap %>m -XX:NewSize=<%= namenode_newsize %>m -XX:MaxNewSize=<%= namenode_newsize %>m -XX:PermSize=128m -XX:MaxPermSize=256m <%= node["bcpc"]["hadoop"]["namenode"]["gc_opts"] %>"
```
bijugs: Will allocating a higher value for ```PermSize`` negatively impact clusters with less memory like VM clusters. By default ``PermSize`` seem to default to 32 or 64mb.

cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb
```diff
@@ -48,8 +48,16 @@ export JSVC_HOME=/usr/lib/bigtop-utils
 # On secure datanodes, user to run the datanode as after dropping privileges
 export HADOOP_SECURE_DN_USER=hdfs
 
+<% datanode_heap = [node["bcpc"]["hadoop"]["datanode"]["xmx"]["max_size"], (node["memory"]["total"].to_i * node["bcpc"]["hadoop"]["datanode"]["xmx"]["max_ratio"]/1024).floor].min %>
+<% datanode_newsize = [(0.125*datanode_heap).ceil, 3072].min %>
+<% namenode_heap = [node["bcpc"]["hadoop"]["namenode"]["xmx"]["max_size"], (node["memory"]["total"].to_i * node["bcpc"]["hadoop"]["namenode"]["xmx"]["max_ratio"]/1024).floor].min %>
+<% namenode_newsize = [(0.125*namenode_heap).ceil, 3072].min %>
+
+export HADOOP_DATANODE_OPTS="$HADOOP_DATANODE_OPTS -Xms<%= datanode_heap %>m -Xmx<%= datanode_heap %>m -XX:NewSize=<%= datanode_newsize %>m -XX:MaxNewSize=<%= datanode_newsize %>m -XX:PermSize=128m -XX:MaxPermSize=256m <%= node["bcpc"]["hadoop"]["datanode"]["gc_opts"] %>"
```
bijugs: ``-Xmn`` is an convenience option to set ``NewSize`` =  ``MaxNewSize``

cookbooks/bcpc-hadoop/attributes/default.rb
```diff
@@ -33,8 +33,9 @@
...
 default["bcpc"]["hadoop"]["namenode"]["handler"]["count"] = 100
-default["bcpc"]["hadoop"]["namenode"]["gc_opts"] = "-XX:ParallelGCThreads=14 -XX:+UseParNewGC -XX:+UseConcMarkSweepGC -XX:CMSInitiatingOccupancyFraction=70 -XX:+UseCMSInitiatingOccupancyOnly -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -Xloggc:/var/log/hadoop-hdfs/gc/gc.log-$$-$(hostname)-$(date +'%Y%m%d%H%M').log -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime"
+default["bcpc"]["hadoop"]["namenode"]["gc_opts"] = "-server -XX:ParallelGCThreads=14 -XX:+UseParNewGC -XX:+UseConcMarkSweepGC -XX:CMSInitiatingOccupancyFraction=70 -XX:+UseCMSInitiatingOccupancyOnly -verbose:gc -XX:+PrintHeapAtGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:+PrintGCDateStamps -Xloggc:/var/log/hadoop-hdfs/gc/gc.log-namenode-$$-$(hostname)-$(date +'%Y%m%d%H%M').log -XX:+PrintTenuringDistribution -XX:+UseNUMA -XX:+PrintGCApplicationStoppedTime -XX:+UseCompressedOops -XX:+PrintClassHistogram -XX:+PrintGCApplicationConcurrentTime"
```
bijugs: Do we want to use ``UseCMSInitiatingOccupancyOnly`` or let JVM use its heuristics to determine when to invoke GC. We had to use ```UseCMSInitiatingOccupancyOnly`` for HBase since the old gen usage pattern is well known since they are occupied by blockCache and memstore objects.

cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb
cbaenziger: Good call, I was testing on hardware. I can scale perm size and max perm size from a floor of 32/64MB to the recommended sizes above.

cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb
cbaenziger: Thanks!

cookbooks/bcpc-hadoop/attributes/default.rb
cbaenziger: I would expect the NN to be very similar having most oldgen objects remain resident. However, the NN should quickly converge to stable CMS heuristics unlike HBase since the equivalent of a cache trashing region reload or scan does not really exist that often on HDFS. I am game to try letting CMS do its own tuning for the NN.

PR #259, opened 2015-08-31
Fix virtualbox version check in virtualbox_env.sh

http-418:
The old virtualbox version check fails for VBox 5.0 because 0 is less than the MIN_MINOR of 3.

This check relies on 'bc' to evaluate 4.3 and 5.0 as floats.

### DIFF
--- a/virtualbox_env.sh
+++ b/virtualbox_env.sh
@@ -13,19 +13,13 @@ if [[ -z "$VBM" ]]; then
     fi
 
     function check_version {
-        local MIN_MAJOR=4
-        local MIN_MINOR=3
+        local MIN_VERSION="4.3"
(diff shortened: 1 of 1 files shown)


virtualbox_env.sh
```diff
@@ -13,19 +13,13 @@ if [[ -z "$VBM" ]]; then
     fi
 
     function check_version {
-        local MIN_MAJOR=4
-        local MIN_MINOR=3
+	local MIN_VERSION="4.3"
```
cbaenziger: It looks like there's a tab instead of spaces here.

http-418:
I've fixed the commit to use spaces instead of tabs.

kiiranh:
This has been tested on Ubuntu. For completeness, needs to be tested on OSX.

PR #260, opened 2015-09-01
Vault mysql root password

DrRayWang:
Chef vault mysql root password and related changes


cbaenziger:
It looks like multiple commits got pulled into this PR?

Also, you'll want to secure [`/etc/mysql/debian.cnf'](https://github.com/bloomberg/chef-bach/blob/4fbbccced71628f9cce4bfd2e3e9ef9d0f5c5675/cookbooks/bcpc/recipes/mysql.rb#L65-L69) to be more restrictive than 644.

DrRayWang:
Chef vault mysql check password and mysql galera password, and made related changes in recipes and template files;
Chef vault ssl-private-key and ssl-certificate and related changes including C-A-R.sh changes to make find_resource can find chef-vault resource executed on worker nodes;
Chef vault ssh-private-key and related changes;

pu239ppy:
I get this error while running cluster assign role
Error executing action `run` on resource 'ruby_block[initialize-ssh-keys]'


ChefVault::Exceptions::SecretDecryption
---------------------------------------
os/ssl is not encrypted with your public key.  Contact an administrator of the vault item to encrypt for you!


Cookbook Trace:
---------------
/var/chef/cache/cookbooks/bcpc/libraries/utils.rb:93:in `get_config'
/var/chef/cache/cookbooks/bcpc/recipes/certs.rb:42:in `block (2 levels) in from_file'

http-418:
This branch converges successfully, and the HDFS / HBase checks pass.  Submitting a pi-estimator M/R job fails, because the mapreduce.tar.gz tarball is absent from HDFS.

I'm not sure whether I'm comfortable merging this or not.  The mapreduce tarball is successfully uploaded in `master`, but not here.  However, I don't see any connection between Ray's changes and the failed upload in bcpc-hadoop::resource_manager

ls output from hdfs:
https://gist.github.com/http-418/2a1cc6b6a98d2e7a8c6c

The output from the failed submission:
https://gist.github.com/http-418/6710c536c3fe6e97fc44

http-418:
This problem goes away if the node is reconverged.  I suspect the reason 'master' works but my personal test does not is that automated-install.sh runs c-a-r fewer times than jenkins.

I will open an issue for the converge problem.

PR #265, opened 2015-09-03
Graphite disk usage

pu239ppy:
In addition to deleting whisper files that have not been updated in $retention_time also get rid of logs.  This approach turned out to be easier then log rotate because it eliminated the need to introspect configuration for the number of relay and cache instances and any customization that it would entail

### DIFF
--- a/cookbooks/bcpc/attributes/graphite.rb
+++ b/cookbooks/bcpc/attributes/graphite.rb
@@ -2,8 +2,10 @@ default['bcpc']['graphite_dbname'] = "graphite"
 default['bcpc']['graphite']['relay_port'] = 2013
 default['bcpc']['graphite']['web_port'] = 8888
 default['bcpc']['graphite']['log']['retention'] = 15
+default['bcpc']['graphite']['data']['retention'] = 15
 default['bcpc']['graphite']['timezone'] = "'America/New_York'"
 default['bcpc']['graphite']['local_data_dir'] = "/opt/graphite/storage/whisper"
+default['bcpc']['graphite']['local_log_dir'] = "/opt/graphite/storage/log"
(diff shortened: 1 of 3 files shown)


cbaenziger:
Yes, sadly DRQS 70769016 showed we were generating around 15k entries per second so too many to delete on a rotation.

pu239ppy:
Aside from fixing the problem that caused excessive logging, we still want to remove older files

cookbooks/bcpc/attributes/graphite.rb
```diff
@@ -2,8 +2,10 @@
 default['bcpc']['graphite']['relay_port'] = 2013
 default['bcpc']['graphite']['web_port'] = 8888
 default['bcpc']['graphite']['log']['retention'] = 15
+default['bcpc']['graphite']['data']['retention'] = 15
 default['bcpc']['graphite']['timezone'] = "'America/New_York'"
 default['bcpc']['graphite']['local_data_dir'] = "/opt/graphite/storage/whisper"
+default['bcpc']['graphite']['local_log_dir'] = "/opt/graphite/storage/log"
```
cbaenziger: If you're adding this, could you [update](https://github.com/kiiranh/chef-bach/blob/2e49aec106fea4799bfa0a66f0a5bbd20c7f0005/cookbooks/bcpc/templates/default/graphite.local_settings.py.erb#L73) Graphite to know that is where its logs should go?

pu239ppy: I don't understand what you mean.  Graphite and carbon already write logs to the correct location

cbaenziger:
Thanks for updating `LOG_DIR` in `carbon.conf`.

PR #266, opened 2015-09-03
chef vault mysql hive password

DrRayWang:
Chef vault mysql-hive-password with related changes

### DIFF
--- a/cluster-assign-roles.sh
+++ b/cluster-assign-roles.sh
@@ -193,14 +193,35 @@ function hadoop_install {
     rm -f $status_file
   }
   trap clean_up_status_file EXIT
-  for m in $(printf ${hosts// /\\n} | grep -i "BCPC-Hadoop-Worker" | sort); do
-    [[ "$m" =~ $REGEX ]]
+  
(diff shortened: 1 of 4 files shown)


amithkanand:
This PR doesn't work. Recipe `hive.rb` is not listed in the run_list of any role. Also, code for removing `hive` password should move to `hive_hcatalog` recipe.

http-418:
Ray, this will need to be manually rebased before merge.

In addition, I'd really like to see the calls into chef-vault moved out of compile time.  It can be very frustrating to troubleshoot an issue that occurs before chef-shell is usable.  I see nothing in this changeset that cannot be migrated into lazy helpers and/or lambdas.

DrRayWang:
This PR also breaks for sitting long here. The problem I see is the find_resource.sh running in BCPC-Hadoop-Worker part does not work anymore. The runlist I see when find_resource.sh running is not right. It only includes some basic roles and recipes, such as networking. The full runlist for Worker role is loaded for some reason. Will try to figure out why.

http-418:
I'm going to give this another shot this afternoon.

vt0r:

PR #267, opened 2015-09-03
chef vault mysql oozie password

DrRayWang:
Chef vault mysql-oozie-password with related changes


cbaenziger:
Let's just use Chef-BACH #270 to track securing Oozie's password usage.

PR #268, opened 2015-09-03
change carbon ulimit before startup

pu239ppy:
In a few cases we have noticed carbon processes running out of open files, this results in massive logging.  While the logs are very useful we want to try to eliminate this condition from happening, by making file limits configurable.  4096 seems like a good place to start


amithkanand:
LGTM.

amithkanand:
can you please change the commit message to specify which component is impacted?

PR #278, opened 2015-09-15
jmxtrans run interval set to 15 seconds

pu239ppy:
Adjust the default kmxtrans SECONDS_BETWEEN_RUNS to be 15 as opposed to default 60

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/default.rb
+++ b/cookbooks/bcpc-hadoop/attributes/default.rb
@@ -185,3 +185,7 @@ default['bcpc']['hadoop']['copylog_enable'] = true
 # File rollup interval in secs for log data copied into HDFS through Flume
 #
 default['bcpc']['hadoop']['copylog_rollup_interval'] = 86400
+#
+# Some jmxtrans defaults
+#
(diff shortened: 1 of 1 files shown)


kiiranh:
Would it be better to set `default['jmxtrans']['run_interval']` in the [`jmxtrans wrapper cookbook's attribute file`](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc_jmxtrans/attributes/default.rb) ?

pu239ppy:
I don't have an answer, we can argue this either way.  jmxtrans already has a default of 60, jmxtransbcpc already relies on bcpc-hadoop to set some things if you look at metadata.rb it depends on it.  So why not run interval?  On the other hand you could be right too I suppose and we are overriding default, that we thing we might use all the time anyway

kiiranh:
I hadn't noticed `bcpc_jmxtrans` depends on `bcpc-hadoop`. In that case, I'd defer to someone else to decide whats best.

kiiranh:
After discussion with @pu239ppy and @cbaenziger , the right structuring can be done as part of future refactoring project. For now, this change looks good.

PR #282, opened 2015-09-21
Setup or manage required Hadoop related OS groups

amithkanand:
This PR takes are of adding or managing Hadoop related OS groups.

### DIFF
--- a/cookbooks/bcpc-hadoop/libraries/utils.rb
+++ b/cookbooks/bcpc-hadoop/libraries/utils.rb
@@ -378,3 +378,12 @@ def user_exists?(user_name)
   return user_found
 end
 
+def group_exists?(group_name)
+  chk_grp_cmd = "getent group #{group_name}"
+  Chef::Log.debug("Executing command: #{chk_grp_cmd}")
+  cmd = Mixlib::ShellOut.new(chk_grp_cmd, :timeout => 10).run_command
+  return  cmd.exitstatus == 0 ? true : false 
+end
+
+def get_group_action(group_name)
+  return  group_exists?(group_name) ? :manage : :create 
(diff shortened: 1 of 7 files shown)


cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -378,3 +378,21 @@ def user_exists?(user_name)
   return user_found
 end
 
+def group_exists?(group_name)
+  group_found = false
+  chk_grp_cmd = "getent group #{group_name}"
+  Chef::Log.debug("Executing command: #{chk_grp_cmd}")
+  cmd = Mixlib::ShellOut.new(chk_grp_cmd, :timeout => 10).run_command
+  if cmd.exitstatus == 0
```
pu239ppy: Just a style suggestion
return cmd.exitstatus == 0

pu239ppy:
Looks good to me

cbaenziger:
I believe this could cause issues on systems which end up getting a SSO configuration in their converge. One would use a `:create` action in that case but when the `group` resource runs the group would already exist and fail out since the group information is not local to the machine?

amithkanand:
@cbaenziger: I changed the code to use a `Ruby Block` and that should take care of your concern.

cbaenziger:
This looks good to me; a couple of formatting nits which [rubocop](http://batsov.com/rubocop/) would gripe on but we need this in master for now so I'm happy to merge.

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -378,3 +378,12 @@ def user_exists?(user_name)
...
 
+def group_exists?(group_name)
+  chk_grp_cmd = "getent group #{group_name}"
+  Chef::Log.debug("Executing command: #{chk_grp_cmd}")
+  cmd = Mixlib::ShellOut.new(chk_grp_cmd, :timeout => 10).run_command
+  return  cmd.exitstatus == 0 ? true : false 
+end
+
+def get_group_action(group_name)
+  return  group_exists?(group_name) ? :manage : :create
```
cbaenziger: A trailing end would be useful here.

amithkanand: Trailing `end` is there. Not sure why not visible here.

PR #285, opened 2015-09-23
Add missing 'end' to get_group_action method

amithkanand:
This PR adds missing `end` to `get_group_action` method.


pu239ppy:
Looks good

PR #286, opened 2015-09-23
Start Kerberos services and define static hadoop group mapping

amithkanand:
In current implementation `kerberos` packages are installed and services are configured using upstream cookbook `krb5`. However, the services are not started by upstream and to be able to configure Kerberization for `Hadoop` cluster services must be started. This PR makes sure that the services are started after `kerberos` packages are installed and `KDC` is configured on bootstrap node. 

Also, this PR adds static group to user mapping for `Hadoop` groups.

### DIFF
--- a/cookbooks/bach_krb5/recipes/krb5_server.rb
+++ b/cookbooks/bach_krb5/recipes/krb5_server.rb
@@ -13,14 +13,20 @@ end
 
 include_recipe "krb5::kadmin_init"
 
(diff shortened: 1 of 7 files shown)


pu239ppy:
Aside from the fact that half of this PR should in all fairness go to krb5 cookbook owners it looks good

http-418:
I was able to build this successfully and pass smoke tests with the updated changes.

PR #287, opened 2015-09-28
patch to wait for chef solr update when adding a new node

DrRayWang:
When adding a new node to a cluster, executing C-A-R.sh would get failure because of chef solr still not updated yet for the new node. Making it wait for the chef solr is what the patch does here until it make "chef node search" work. This checking process runs in a loop, jump out only if "chef node search" return not nil value.


amithkanand:
LGTM. Tested on my `VM` cluster.

PR #289, opened 2015-10-01
fix chef-vault cookbook version issue

DrRayWang:
Set chef-vault cookbook used here to be 1.3.0, because version 1.3.1 needs ruby 2.0 support, without it will break the cluster build at cookbook upload stage


pu239ppy:
good catch

Issue #291, opened 2015-10-02
Error while chef-ing worker node.


amithkanand:
After introducing `chef-vault`, cluster build currently throws an error while trying to create the `zabbix-agentd.conf` file. Below is the error message
```` 
================================================================================
Error executing action `create` on resource 'template[/usr/local/etc/zabbix_agentd.conf]'
================================================================================

Chef::Mixin::Template::TemplateError
------------------------------------
os/mysql-zabbix is not encrypted with your public key.  Contact an administrator of the vault item to encrypt for you!
```

DrRayWang:
The reason is :
zabbix-head.rb will chef vault mysql-zabbix-password with chef vault resource.This recipe only runs on Head nodes.
But zabbix-worker.rb run template of this file : cookbooks/bcpc/templates/default/zabbix_agentd.conf.erb, which needs access to mysql-zabbix-password using chef vault ruby API. find_resource can only find chef vault resource, not chef vault ruby API. This results in execution failure for worker nodes when it run zabbix-worker.rb, and needs access to chef-vaulted mysql-zabbix-password.

kiiranh:
Fixed by #292 . Closing

PR #292, opened 2015-10-02
Remove zabbix-work recipe from the run_list

amithkanand:
This PR removes `zabbix-work` recipe from the run list of `BCPC-Hadoop-Worker` and BCPC-Hadoop-Head` roles and fixes issue #291


DrRayWang:
Looks good to me

cbaenziger:
Indeed; we will still run `bcpc::zabbix-work` in `bcpc::zabbix-head` which should be fine as this should still resolve Chef-BACH #291 handily.

Issue #178, opened 2015-06-17
Oozie sharelib directory changed between hdp2.0 to hdp2.2 and is cryptic


kiiranh:
On hdp2.0, the oozie sharelib directory on hdfs was /user/oozie/share/lib. In the hdp2.2 installation, that changed to /user/oozie/share/lib/lib_<some number> (eg: /user/oozie/share/lib/lib_20150605140532). This change of path could possibly break existing users after upgrading to hdp2.2.

kiiranh:
After some analysis, found that the [code that initiates oozie sharelib setup](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/oozie.rb#L104-L116) was change under commits [073ab64](https://github.com/bloomberg/chef-bach/commit/073ab64d1655abcd34fa9bfd980d78b38e7afe60). It now calls `oozie-setup.sh sharelib upgrade` which internally calls [OozieSharelibCLI.java](https://github.com/apache/oozie/blob/master/tools/src/main/java/org/apache/oozie/tools/OozieSharelibCLI.java). And for whatever reason, that class [creates a new directory lib_<current timestamp>](https://github.com/apache/oozie/blob/master/tools/src/main/java/org/apache/oozie/tools/OozieSharelibCLI.java#L151-L153) in `hdfs:///user/oozie/share/lib` instead of just copying over the libs in  `hdfs:///user/oozie/share/lib`.

kiiranh:
So the changes may not have been caused by hdp2.0->hdp2.2, but the chef-bach code change which was changed to use the [upgrade](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/oozie.rb#L107) option. The libraries in the lib_<timestamp> directory could still work, but it would be good to have consistency in oozie sharelib paths.

kiiranh:
Also, the [temp file](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/oozie.rb#L110) created during share lib setup on hdfs isn't cleaned up correctly. This could skip sharelib upgrade in the next chef-client run.

cbaenziger:
For some more background on Oozie shared-libs:
[...]

PR #295, opened 2015-10-06
Fix condition checks regarding oozie sharelib dir/lib update

kiiranh:
I observed on a test VM cluster that a new oozie sharelib directory (hdfs:///user/oozie/share/lib/lib_<timestamp>) was created on every chef-client run. This PR fixes a few conditional errors in the code which manages the sharelib directory.

Fixes Issue #178

### DIFF
--- a/cookbooks/bcpc-hadoop/libraries/utils.rb
+++ b/cookbooks/bcpc-hadoop/libraries/utils.rb
@@ -382,11 +382,11 @@ def group_exists?(group_name)
   chk_grp_cmd = "getent group #{group_name}"
   Chef::Log.debug("Executing command: #{chk_grp_cmd}")
   cmd = Mixlib::ShellOut.new(chk_grp_cmd, :timeout => 10).run_command
-  return  cmd.exitstatus == 0 ? true : false 
+  return cmd.exitstatus == 0 ? true : false 
 end
 
 def get_group_action(group_name)
-  return  group_exists?(group_name) ? :manage : :create 
+  return group_exists?(group_name) ? :manage : :create 
 end
 
 def has_vip?
@@ -394,4 +394,48 @@ def has_vip?
     "ip addr show", :timeout => 10
   ).run_command
   cmd.stderr.empty? && cmd.stdout.include?(node[:bcpc][:management][:vip])
+end
+
+# Internal: Check if oozie server is running on the given host.
+#
+# host - Endpoint (FQDN/IP) on which Oozie server is available.
+#
+# Examples
+#
+#   oozie_running?("f-bcpc-vm2.bcpc.example.com")
+#   # => true
+#
+# Returns true if oozie server is operational with 'NORMAL' status, false otherwise.
+def oozie_running?(host)
+    oozie_url = "sudo -u oozie oozie admin -oozie http://#{host}:11000/oozie -status"
+    cmd = Mixlib::ShellOut.new(
+      oozie_url, :timeout => 20
+    ).run_command
+    Chef::Log.debug("Oozie status: #{cmd.stdout}")
+    cmd.exitstatus == 0 && cmd.stdout.include?('NORMAL')
+end
+
+# Internal: Have the specified Oozie host update its ShareLib to the latest lib_<timestamp> 
+#           sharelib directory on hdfs:/user/oozie/share/lib/, without having to restart 
+#           that Oozie server. Oozie server, by default, uses the latest one when it (re)starts.
+#
+# host - Endpoint (FQDN/IP) on which Oozie server is available.
+#
+# Returns nothing. 
+def update_oozie_sharelib(host)
+  if oozie_running?(host) 
+    update_sharelib = "sudo -u oozie oozie admin -oozie http://#{host}:11000/oozie -sharelibupdate"
+    cmd = Mixlib::ShellOut.new(
+      update_sharelib, :timeout => 20
+    ).run_command
+    if cmd.exitstatus == 0
+      Chef::Log.info("Sharelibupdate: Updated sharelib on #{host}")
+    else
+      Chef::Log.info("Sharelibupdate: sharelibupdate command failed on #{host}")
+      Chef::Log.info("  stdout: #{cmd.stdout}")
+      Chef::Log.info("  stderr: #{cmd.stderr}")
+    end 
+  else
+    Chef::Log.info("Sharelibupdate: Oozie server not running on #{host}")
+  end
 end
--- a/cookbooks/bcpc-hadoop/recipes/oozie.rb
+++ b/cookbooks/bcpc-hadoop/recipes/oozie.rb
@@ -5,7 +5,6 @@ dpkg_autostart "oozie" do
   allow false
 end
 
-#%w{libmysql-java zip unzip extjs hadoop-lzo oozie oozie-client}.each do |pkg|
 %w{zip unzip extjs hadoop-lzo oozie oozie-client}.each do |pkg|
   package pkg do
     action :upgrade
@@ -20,7 +19,8 @@ end
 OOZIE_LIB_PATH='/usr/hdp/current/oozie'
 OOZIE_CLIENT_PATH='/usr/hdp/current/oozie-client'
 OOZIE_SERVER_PATH='/usr/hdp/current/oozie-server/oozie-server'
-HDFS_URL="hdfs://#{node.chef_environment}/"
+OOZIE_SHARELIB_TARBALL_PATH='/usr/hdp/2.2.0.0-2041/oozie/oozie-sharelib.tar.gz'
+HDFS_URL=node[:bcpc][:hadoop][:hdfs_url]
 
 directory "#{OOZIE_LIB_PATH}/libext" do
   owner "oozie"
@@ -37,16 +37,6 @@ directory "/var/run/oozie" do
   action :create
   recursive true
 end
-
-#bash "copy hadoop libs" do
-#  src_dir="/usr/hdp/2.2.0.0-2041/hadoop/lib"
-#  dst_dir="#{OOZIE_LIB_PATH}/libext/"
-#  code "for f in `ls *.jar`; do ln -s #{src_dir}/$f #{dst_dir}/$f; done"
-#  cwd src_dir
-#  user "oozie"
-#  # check if all source files are in destination directory
-#  not_if { (::Dir.entries("#{src_dir}") - ::Dir.entries("#{dst_dir}")).empty? }
-#end
 
 %w{/usr/share/HDP-oozie/ext-2.2.zip
    /usr/share/java/mysql-connector-java.jar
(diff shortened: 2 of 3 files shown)


cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -106,23 +106,14 @@
...
   user "oozie"
-  not_if "sudo -u hdfs hdfs dfs -test -d #{HDFS_URL}/user/oozie/share/lib", :user => "hdfs"
-  only_if "echo 'test'|sudo -u hdfs hdfs dfs -copyFromLocal - /tmp/oozie-test"
-  notifies :run,"bash[delete-oozie-temp-file]",:immediately
+  not_if "hdfs dfs -test -d #{HDFS_URL}/user/oozie/share/lib", :user => "hdfs"
+  only_if "echo 'test'| hdfs dfs -copyFromLocal - /tmp/oozie-test && hdfs dfs -rm -skipTrash /tmp/oozie-test", :user => "hdfs"
   #not_if { require 'time'
   # hdfs_mtime=`hdfs dfs -stat #{share_dir_url}`.strip
   # Time.parse("#{hdfs_mtime} UTC") >
   # File.mtime("#{OOZIE_CLIENT_PATH}/oozie-sharelib.tar.gz") }
```
amithkanand: Can we remove these commented lines?

cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -106,23 +106,14 @@
   #code "#{OOZIE_LIB_PATH}/bin/oozie-setup.sh sharelib update -fs #{HDFS_URL}"
```
amithkanand: Can we remove these commented line?

cookbooks/bcpc-hadoop/recipes/oozie.rb
kiiranh: Per @cbaenziger, these need to be uncommented again so sharelib is updated on hdfs. The current change will only make sure sharelib dir is created once on hdfs, it won't update if the on-disk sharelib has changed

kiiranh:
@amithkanand Comments incorporated. Now, sharelib is updated on hdfs only if the sharelib tarball has been changed/updated.

cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -65,61 +56,73 @@
...
+    node.set[:bcpc][:hadoop][:oozie][:sharelib_checksum] = 
+      Digest::MD5.hexdigest(File.read(OOZIE_SHARELIB_TARBALL_PATH))
+  end
+  action :nothing
+  notifies :run, "ruby_block[notify_sharelib_update]", :immediately
+end
+
+update_oozie_sharelib = lambda do |host|
+  begin
+    oozie_status=`sudo -u oozie oozie admin -oozie http://#{host}:11000/oozie -status 2>&1`
```
cbaenziger: Please use [mixlib-shellout](https://github.com/chef/mixlib-shellout) instead of ```

cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -65,61 +56,73 @@
...
+    if /NORMAL/ =~ oozie_status && $?.to_i
```
cbaenziger: Here `$?.to_i` will always evaluate to true; regardless of if `$?` is 0 or 1. I suspect the intention is `$? != 0`.

cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -65,61 +56,73 @@
...
+      update_status=`sudo -u oozie oozie admin -oozie http://#{host}:11000/oozie -sharelibupdate 2>&1`
+      Chef::Log.warn("Sharelibupdate: Updated sharelib on #{host}") if $?.to_i
```
cbaenziger: Again `$? != 0`?

cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -65,61 +56,73 @@
...
+    end
+  rescue Exception => e
+    Chef::Log.error(e.message)
+  end
+end
+
+ruby_block "notify_sharelib_update" do
+  block do
+    node[:bcpc][:hadoop][:oozie_hosts].each do |oozie_host|
+      update_oozie_sharelib.call(oozie_host[:hostname])
```
cbaenziger: As you pointed out; should be `float_host()`

cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -155,15 +158,14 @@
...
+        CREATE DATABASE oozie;
+        GRANT #{privs} ON oozie.* TO 'oozie'@'%' IDENTIFIED BY '#{get_config('mysql-oozie-password')}';
+        GRANT #{privs} ON oozie.* TO 'oozie'@'localhost' IDENTIFIED BY '#{get_config('mysql-oozie-password')}';
+        FLUSH PRIVILEGES;
       EOF
       IO.popen("mysql -uroot -p#{get_config('mysql-root-password')}", "r+") do |db|
         db.write code
       end
-      #system "sudo -u oozie #{OOZIE_CLIENT_PATH}/bin/ooziedb.sh create -sqlfile #{OOZIE_CLIENT_PATH}/oozie.sql -run Validate DB Connection"
       system "sudo -u oozie /usr/hdp/current/oozie-server/bin/ooziedb.sh create -sqlfile /usr/hdp/current/oozie-server/oozie.sql -run Validate DB Connection"
```
cbaenziger: Could you continue to use `#{OOZIE_CLIENT_PATH}`?

cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -181,9 +183,9 @@
 ruby_block "Oozie Down" do
   i = 0
   block do
-    status=`sudo -u oozie oozie admin -oozie http://"#{float_host(node[:fqdn])}":11000/oozie -status 2>&1` 
+    status=`sudo -u oozie oozie admin -oozie #{OOZIE_URL} -status 2>&1` 
     while not /NORMAL/ =~ status and $?.to_i
-      status=`sudo -u oozie oozie admin -oozie http://"#{float_host(node[:fqdn])}":11000/oozie -status 2>&1` 
+      status=`sudo -u oozie oozie admin -oozie #{OOZIE_URL} -status 2>&1`
```
cbaenziger: Maybe create a function or some sort of definition for testing that Oozie's alive? This is the second instance of this status check code in the recipe.

roles/BCPC-Hadoop-Head-MapReduce.json
```diff
@@ -1,6 +1,13 @@
 {
   "name": "BCPC-Hadoop-Head-MapReduce",
   "default_attributes": {
+    "bcpc": { 
+      "hadoop": {
+        "oozie": {
+          "sharelib_checksum": "updated_automatically"
```
cbaenziger: I would prefer for `oozie.rb` to handle this attribute being unset (or having a cookbook default attribute) rather than putting a required piece of info in a role. If someone included the Oozie recipe they would also need to set this in some one-off way.

cookbooks/bcpc-hadoop/recipes/oozie.rb
http-418: $? has the methods "success?" and "exitstatus" for this purpose

(as with perl, taking $? directly as an integer can have surprising results, because the exit code is stored in the high bits of a 16 bit integer)

cookbooks/bcpc-hadoop/recipes/oozie.rb
kiiranh: Can't use `#{OOZIE_CLIENT_PATH}` since the file `ooziedb.sh` moved from that path to `/usr/hdp/current/oozie-server/bin/ooziedb.sh`:
```bash
bcpc-vm2:~$ find /usr/hdp -type f -name ooziedb*
/usr/hdp/2.2.0.0-2041/oozie/bin/ooziedb.sh
bcpc-vm2:~$ ls -l /usr/hdp/current/oozie-server
lrwxrwxrwx 1 root root 27 Nov 24 15:14 /usr/hdp/current/oozie-server -> /usr/hdp/2.2.0.0-2041/oozie
```

That is why the command with `#{OOZIE_CLIENT_PATH}` was commented out. I merely deleted the commented out code in this commit.

kiiranh:
Code changes have been tested with kerberos enabled as well. 

**Note**:
Current code only [sets](https://github.com/kiiranh/chef-bach/blob/oozie_sharelib_fix/cookbooks/bcpc-hadoop/recipes/oozie.rb#L115-L116) the sharelib file checksum in the node object but does not perform a `node.save`. It expects that the attribute will be persisted when the node object is saved at the end of the chef-client run. Caveat here is: it could take extra chef-client runs for this code to converge if the initial chef-client runs fail to save the node object (e.g.: due to some exception later on in the run_list). You might end up with a couple extra sharelib directories in hdfs because of this. Those will be cleaned up per the sharelib cleanup schedule mentioned [here](http://blog.cloudera.com/blog/2014/05/how-to-use-the-sharelib-in-apache-oozie-cdh-5/)

cbaenziger:
Please provide function documentation for the `utils.rb` code. Per the `README.md` [Tomdoc](http://tomdoc.org/) style would be good. (See "Method Documentation" therein.)

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -395,3 +395,33 @@ def has_vip?
...
+      update_sharelib = "sudo -u oozie oozie admin -oozie http://#{host}:11000/oozie -sharelibupdate"
+      cmd = Mixlib::ShellOut.new(
+        update_sharelib, :timeout => 10
+      ).run_command
+      Chef::Log.info("Sharelibupdate: Updated sharelib on #{host}") if cmd.exitstatus == 0 
+    else
+      Chef::Log.info("Sharelibupdate: Failed to update sharelib on #{host}")
+    end
+  rescue Exception => e
+    Chef::Log.error(e.message)
```
cbaenziger: I would like us not to swallow exceptions unless we have a good reason for handling them elsewhere.

kiiranh:
Incorporated review comments:
- Added documentation for the proposed library functions
- Increased timeout for shellout commands from 10 to 20 to factor in the retries (default 4 times, after intervals 1, 2, 4, 8 seconds) that the Oozie client makes to try to connect to Oozie server.
- Removed `rescue` blocks that swallowed exceptions because the failures we expect (oozie server not reachable/running, user oozie doesn't exist, user oozie doesn't have permission to execute the commands) will cause a non-zero exit status for shellout and will not throw any exceptions. So if and when an exception does occur we want it to be noticed.

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -395,3 +395,45 @@ def has_vip?
...
+def update_oozie_sharelib(host)
+  if oozie_running?(host) 
+    update_sharelib = "sudo -u oozie oozie admin -oozie http://#{host}:11000/oozie -sharelibupdate"
+    cmd = Mixlib::ShellOut.new(
+      update_sharelib, :timeout => 20
+    ).run_command
+    if cmd.exitstatus == 0
+      Chef::Log.info("Sharelibupdate: Updated sharelib on #{host}")
+    else
+      Chef::Log.info("Sharelibupdate: sharelibupdate command failed on #{host}")
```
pu239ppy: Consider returning STDERR and STDOUT as part of your error message

kiiranh: @pu239ppy Done

pu239ppy:
Looks good

PR #297, opened 2015-10-07
enable gssapi authentication for sshd

pu239ppy:
Enable the GSSAPI Authentication flag

### DIFF
--- a/cookbooks/bcpc/templates/default/sshd_config.erb
+++ b/cookbooks/bcpc/templates/default/sshd_config.erb
@@ -61,8 +61,13 @@ ChallengeResponseAuthentication no
 #KerberosTicketCleanup yes
 
 # GSSAPI options
+<% if node[:bcpc][:kerberos][:enable] == true %>
+GSSAPIAuthentication yes
+GSSAPICleanupCredentials yes
+<% else %>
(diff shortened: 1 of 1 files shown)


cookbooks/bcpc/templates/default/sshd_config.erb
```diff
@@ -61,8 +61,13 @@ ChallengeResponseAuthentication no
 #KerberosTicketCleanup yes
 
 # GSSAPI options
+<% if node[:bcpc][:kerberos][:enable] == true %>
+GSSAPIAuthentication yes
+GSSAPICleanupCredentials yes
+<% else %>
 #GSSAPIAuthentication no
```
cbaenziger: Mind removing these comments if not necessary?

cbaenziger:
Admittedly, I would rather see us stop twiddling ssh configs and instead use something like the [openssh cookbook](https://github.com/chef-cookbooks/openssh).

cookbooks/bcpc/templates/default/sshd_config.erb
pu239ppy: The comments are part of stock sshd_config, I would rather keep it in its entirety just to make sure we keep whatever defaults are there

Issue #294, opened 2015-10-06
get_config issue about returning nothing result in silent fail


DrRayWang:
Currently, `get_config()` return nil if password is not found in the data bag. In this case, it will fail password setting scenario like `get_config('password','mysql-graphite','os')` in [this](https://github.com/DrRayWang/chef-bach/blob/afff934491925cc970ed493b0f6a016764b270ba/cookbooks/bcpc/recipes/graphite.rb#L174-L179): 
```ruby
ruby_block "graphite-database-creation" do
    block do
        if not system "mysql -uroot -p#{get_config('mysql-root-password')} -e 'SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = \"#{node['bcpc']['graphite_dbname']}\"'|grep \"#{node['bcpc']['graphite_dbname']}\"" then
            %x[ mysql -uroot -p#{get_config('mysql-root-password')} -e "CREATE DATABASE #{node['bcpc']['graphite_dbname']};"
                mysql -uroot -p#{get_config('mysql-root-password')} -e "GRANT ALL ON #{node['bcpc']['graphite_dbname']}.* TO '#{get_config('mysql-graphite-user')}'@'%' IDENTIFIED BY '#{get_config('password','mysql-graphite','os')}';"
                mysql -uroot -p#{get_config('mysql-root-password')} -e "GRANT ALL ON #{node['bcpc']['graphite_dbname']}.* TO '#{get_config('mysql-graphite-user')}'@'localhost' IDENTIFIED BY '#{get_config('password','mysql-graphite','os')}';"
```
We have another method `get_config!()` to raise error when `get_config()` return nil. Hence we should modify current `get_config!()` and use it for all the cases that get password from chef vault to shout out password missing case, like above code example.

kiiranh:
The related PRs have been merged. This can be closed.

PR #298, opened 2015-10-07
Get config fix

DrRayWang:
Fixed the [issue 294](https://github.com/bloomberg/chef-bach/issues/294), and made related code change in code case to apply get_config!().

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -115,8 +115,8 @@ def delete_config(key)
   end
 end
 
-def get_config!(key)
-  value = get_config(key)
+def get_config!(key,item=node.chef_environment,bag="configs")
+  value = get_config(key,item,bag)
   raise "Failed to find value for #{key}!" if value.nil?
   return value
 end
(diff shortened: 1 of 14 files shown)


amithkanand:
Recipe changes?

DrRayWang:
Yes all the recipes which use old get_config to access chef vault item are changed to using new get_config!()

http-418:
This failed to build for me.  Couldn't chef nodes:
https://gist.github.com/http-418/82bebe00669100128c58

DrRayWang:
Andrew, I tested it again, I didn't get your errors. Could you check your test environment setting, and try it again.

pu239ppy:
Worked for me, however you should remove FLUSH PRIVILEGES it is unnecesary, since you are using GRANT and not directly modifying privilege tables by INSERTing or DELETEing

DrRayWang:
we can work on "remove FLUSH PRIVILEGES" in a different PR.

kiiranh:
You could save a code-test-review cycle if you make the change in this PR!

cookbooks/bcpc/libraries/utils.rb
```diff
@@ -115,8 +115,8 @@ def delete_config(key)
   end
 end
 
-def get_config!(key)
-  value = get_config(key)
+def get_config!(key,item=node.chef_environment,bag="configs")
+  value = get_config(key,item,bag)
   raise "Failed to find value for #{key}!" if value.nil?
```
kiiranh: It would have been good to print `item` along with `key` in the Error message

pu239ppy:
Built a cluster with this pull request tested with C-A-R

PR #299, opened 2015-10-07
Move sqoop material into its own cookbook

http-418:
This PR moves the sqoop installation from bcpc::data-node to its own cookbook, bach_sqoop.

It also adds support for installing JDBC jar files into Sqoop's lib directory from arbitrary URLs.  The list of URLs and checksums is communicated via node attributes.

### DIFF
--- a/cookbooks/bach_sqoop/README.md
+++ b/cookbooks/bach_sqoop/README.md
@@ -16,15 +16,11 @@ Only a default recipe is included.  The default recipe installs Sqoop, creates i
 Attributes
 ==========
 
-<ol>
-  <li>node[:bach][:sqoop][:install_dir]<br/>
-    <p>The Sqoop install directory.  Default: <tt>/usr/lib/sqoop</tt></p>
-  </li>
-  <li>node[:bach][:sqoop][:install_dir]<br/>
-    <p>The Sqoop library directory, where JDBC jars will be installed. Default: <tt>/usr/lib/sqoop/lib</tt></p>
-  </li>
-  <li>node[:bach][:sqoop][:jdbc_jars]<br/>
-    <p>This attribute is a hash.  The keys are the filenames for JDBC jars.  Each value is another hash, containing an 'url' key and a 'checksum' key.</p>
-	<p>See <tt>attributes/default.rb</tt> for an example value.</p>
-  </li>
-</ol>
+1. node[:bach][:sqoop][:install_dir]
+The Sqoop install directory.  Default: <tt>/usr/lib/sqoop</tt>
+
+2. node[:bach][:sqoop][:install_dir]
+The Sqoop library directory, where JDBC jars will be installed. Default: <tt>/usr/lib/sqoop/lib</tt>
+
+3. node[:bach][:sqoop][:jdbc_jars]
+This attribute is a hash.  The keys are the filenames for JDBC jars.  Each value is another hash, containing an 'url' key and a 'checksum' key. See <tt>attributes/default.rb</tt> for an example value.
--- a/cookbooks/bach_sqoop/README.md
+++ b/cookbooks/bach_sqoop/README.md
@@ -1,0 +1,1 @@
+Cookbook to install Apache Sqoop on data nodes.
--- a/cookbooks/bach_sqoop/README.md
+++ b/cookbooks/bach_sqoop/README.md
@@ -1,1 +1,30 @@
-Cookbook to install Apache Sqoop on data nodes.
+Overview
+========
+
+This cookbook installs Apache Sqoop and needed JDBC jars. The Sqoop packages come from Hortonworks via 'apt'.  JDBC jars are downloaded directly.
+
+Requirements
+============
+
+This cookbook has been tested with Chef 11 and 12.
+
+Recipes
+=======
+
+Only a default recipe is included.  The default recipe installs Sqoop, creates its library directory, and installs any JDBC jars to the library directory.
+
+Attributes
+==========
+
+<ol>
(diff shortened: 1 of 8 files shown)


cookbooks/bcpc-hadoop/metadata.rb
```diff
@@ -7,6 +7,7 @@
 version          "0.1.0"
 
 depends "bcpc", ">= 0.5.0"
+depends "bach_sqoop"
```
kiiranh: I suppose the dependency should be inverted: 
- `bach_sqoop` should depend on `bcpc-hadoop` instead
- Instead of including `bach_sqoop` in `bcpc-hadoop::datanode`, we should add `bach_sqoop` to `[BCPC-Hadoop-Worker role](https://github.com/bloomberg/chef-bach/blob/master/roles/BCPC-Hadoop-Worker.json#L19)`

cookbooks/bach_sqoop/metadata.rb
```diff
@@ -0,0 +1,11 @@
+# encoding: utf-8
+
+name             'bach_sqoop'
+maintainer       'Bloomberg Finance L.P.'
+description      'Cookbook to setup Apache Spark'
```
cbaenziger: I think you have a typo here?

cookbooks/bach_sqoop/attributes/default.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+
+default[:bach][:sqoop][:install_dir] = '/usr/lib/sqoop'
+
+default[:bach][:sqoop][:lib_dir] =
+  default[:bach][:sqoop][:install_dir] + '/lib'
+
+# Place JDBC jars in this hash.
+# They will be installed to the sqoop lib_dir.
+default[:bach][:sqoop][:jdbc_jars] =
```
cbaenziger: Do we want to point to specific files or use Maven more fully? I'm not sure how complex the  requirements are for various databases; e.g. MS SQL, Greenplum, Oracle, etc.?

cookbooks/bach_sqoop/README.md
```diff
@@ -0,0 +1 @@
+Cookbook to install Apache Sqoop on data nodes.
```
cbaenziger: Could you please provide a more full README? Describing the attributes and recipes available?

http-418:
I have updated the README with more useful information.

I'm still thinking about the dependency question.

cookbooks/bach_sqoop/attributes/default.rb
http-418: For JDBC specifically, I think they should all offer self-contained jars to make this easy.  (They're already putting up a much higher barrier than java developers expect by putting the jars behind a click-through license)

For something unrelated to JDBC, we would have to write POMs or cache an upstream artifact repository, then use maven to install the dependency tree.

http-418:
I really don't like the inverted dependency: if sqoop requires bcpc-hadoop, why not just be a part of bcpc-hadoop?

Instead, I moved the zookeeper configuration directory into the defaults file for sqoop.  bcpc-hadoop requires bach_sqoop, and sets the default to the same as its own zookeeper conf dir.

cookbooks/bach_sqoop/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Recipes
+=======
+
+Only a default recipe is included.  The default recipe installs Sqoop, creates its library directory, and installs any JDBC jars to the library directory.
+
+Attributes
+==========
+
+<ol>
```
cbaenziger: Mark-up supports [bulleted lists](https://guides.github.com/features/mastering-markdown/) on its own, no need to HTML.

cookbooks/bach_sqoop/templates/default/sq_sqoop-env.sh.erb
```diff
@@ -1,4 +1,6 @@
- Licensed to the Apache Software Foundation (ASF) under one or more
+# -*- mode: shell-script -*-
```
cbaenziger: Erm, it looks like some vi bits leaked in?

cookbooks/bcpc-hadoop/metadata.rb
cbaenziger: I would rather not see` node[:bcpc][:hadoop]` attributes used here. Please use `node[:bach][:sqoop]` and use a wrapper in bcpc-hadoop to set the `node[:bach][:sqoop]` attributes needed (like Hannibal does).

http-418: Clay, this was already fixed in the latest (squashed) commit.

cookbooks/bach_sqoop/templates/default/sq_sqoop-env.sh.erb
http-418: Is that bad? Many tools will parse the modeline, and this file has a non-standard filename for a shell script.

cookbooks/bach_sqoop/README.md
http-418: I've changed the ordered list to a markdown list.

cookbooks/bach_sqoop/templates/default/sq_sqoop-env.sh.erb
cbaenziger: Ah good call; I'm happy to set a precedent for it.

cbaenziger:
Kiran were you able to test this successfully?

http-418:
Rather than provide JDBC jars and sqoop ourselves, we will be documenting how clients can provide their JDBC jars to oozie jobs that run sqoop.

PR #300, opened 2015-10-08
Graphite to zabbix optimization

kiiranh:
This PR improves the graphite-to-zabbix function:
- Fix formatting issue for metric value sent from graphite to zabbix
- **Create Zabbix objects only on the head node holding vip** (So only one head node will configure Zabbix)
- Increase disk space check limit
- Numerical formatting fixes for graphite metrics
- Set space used precision to 2 decimal places based on Zabbix latest data UI
- **Reduce Zabbix API calls for creation of Hosts, Items, Triggers and Actions** (This significantly reduces the time it took to configure all the objects in Zabbix)
- Add trigger for bootstrap chef-client runs

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/monitoring.rb
+++ b/cookbooks/bcpc-hadoop/attributes/monitoring.rb
@@ -4,6 +4,10 @@ default["bcpc"]["hadoop"]["zabbix"]["cron_check_time"] = 240
 default["bcpc"]["hadoop"]["zabbix"]["mail_source"] = "zabbix.zbx_mail.sh.erb"
 default["bcpc"]["hadoop"]["zabbix"]["cookbook"] = nil 
 
+# Interval within which chef-client is expected to run
+default["bcpc"]["hadoop"]["zabbix"]["chef_client_check_interval"] =
+  "#{((node['chef_client']['interval'].to_i + node['chef_client']['splay'].to_i) * 2 / 60).ceil}m"
+
 # Override Graphite/Zabbix queries/triggers here
 default["bcpc"]["hadoop"]["graphite"]["basic_queries"] = {} # Basic OS/Node Queries
 default["bcpc"]["hadoop"]["graphite"]["service_queries"] = {} # Service specific queries
--- a/cookbooks/bcpc-hadoop/recipes/graphite_queries.rb
+++ b/cookbooks/bcpc-hadoop/recipes/graphite_queries.rb
@@ -1,18 +1,37 @@
 # Get a list of cluster nodes to monitor
 monitored_nodes_objs = get_all_nodes.select { |n| not n.hostname =~ /bootstrap/ }.compact
 
-# Graphite queries which specify property to query and alarming trigger, severity(maps to zabbix's api -> trigger -> priority)
-# and owner who the trigger is routed to for resolution
-# Queries are structured as they appear in graphite dashboard
-# Refer Zabbix API(https://www.zabbix.com/documentation/2.2/manual/api/reference/) for the various attributes used
-# The following list is supposed to be a skeleton which should be extended/edited according to the deployment scenario
-# By default items, triggers actions are disabled and won't be monitored. To enable them, set "enable=true" in the json structure
-triggers = {}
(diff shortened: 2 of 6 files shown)


amithkanand:
As discussed, @kiiranh can you please take care of code formatting?

kiiranh:
@amithkanand Done. Please let me know if you see any issues.

amithkanand:

kiiranh:
@amithkanand Commits squashed. Tested this on the latest master. Code is working as expected.

cookbooks/bcpc-hadoop/attributes/monitoring.rb
```diff
@@ -4,6 +4,10 @@
 default["bcpc"]["hadoop"]["zabbix"]["mail_source"] = "zabbix.zbx_mail.sh.erb"
 default["bcpc"]["hadoop"]["zabbix"]["cookbook"] = nil 
 
+# Interval within which chef-client is expected to run
+default["bcpc"]["hadoop"]["zabbix"]["chef_client_check_interval"] =
+  "#{((node['chef_client']['interval'].to_i + node['chef_client']['splay'].to_i) * 2 / 60).ceil}m"
```
cbaenziger: Probably overboard for now:
In the future, I could see different machines with different run-times (e.g. a worker may run quickly but maintenance tasks may blow up run time every once and a while). As such it may make sense for each Chef run to talk to Graphite to get its last execution time or some simple statistics to compute how long it _should_ take (we could even dynamically kick the trigger out before running a long-running task). Otherwise, we could query our node object for the last `:ohai_time` on the node object for when it last started a Chef run (could keep an array for summary statistics).

For now just a Chef attribute should be awesome though.

cookbooks/bcpc-hadoop/recipes/graphite_queries.rb
```diff
@@ -49,20 +70,33 @@
...
-  # Format the mount names like diamond(https://github.com/python-diamond/Diamond/blob/master/src/collectors/diskspace/diskspace.py#L196-L197) and convert size to GB
-  disk_size_hash = disk_size_hash.map{ |mount,kb_size| tmp = mount.gsub('/', '_').gsub('.', '_').gsub('\\', ''); [tmp == '_' ? 'root':tmp, (kb_size.to_i / 1024 / 1024).ceil] }
-  
+  disk_size_hash = node_obj.filesystem
+    .map { |fs,props| [props['mount'],props['kb_size']] }.compact.uniq
+  disk_size_hash = disk_size_hash.reject { |mount,kb_size| kb_size.nil? }
+    .reject {
+      |mount,kb_size| mount.start_with?('/dev') || mount.start_with?('/sys') ||
+      mount.start_with?('/proc') || mount.start_with?('/run') ||
+      mount.start_with?('/boot') || mount.start_with?('/tmp')
```
cbaenziger: Please don't mask out `/tmp`; it's pretty important to watch.

cookbooks/bcpc-hadoop/recipes/graphite_queries.rb
```diff
@@ -72,17 +106,17 @@
...
         'query' => "diskspace.#{disk}.byte_used",
         'trigger_val' => "max(61,0)",
-        'value_type' => 3,
-        'trigger_cond' => ">#{(size * 0.80).floor}G",
+        'value_type' => 0,
+        'trigger_cond' => ">#{(size * 0.90).round(2)}G",
         'trigger_name' => "#{host}_#{disk}_SpaceUsed",
         'enable' => true,
-        'trigger_desc' => "More than 80% of disk space used",
+        'trigger_desc' => "More than 90% of disk space used",
```
cbaenziger: Again the percentage is quite arbitrary and the total remaining bytes as well. The big question is how much does data usually grow over time and can we support some amount of time of growth (e.g. down to ~5 days of space left) is probably more important with some fall backs of minimum expected slop space (e.g. enough to run maintenance Map/Reduce tasks and such).

Again, for now this should be great.

amithkanand:
LGTM.

PR #301, opened 2015-10-09
Changes to enable HBase replication

bijugs:
Changes in this PR will enable users to set-up HBase replication. 

In order to enable replication users need to 

- Set ``default["bcpc"]["hadoop"]["hbase"]["repl"]["enabled"]`` to ``true``
- Set ``default["bcpc"]["hadoop"]["hbase"]["repl"]["peer_id"]`` to a string which will be used to create the replication peer in HBase and the string should not include "-" (hyphens). By default the string will be set to the chef environment name.
- Set ``default["bcpc"]["hadoop"]["hbase"]["repl"]["target"]`` to a string value of "target hbase ZooKeeper quorum hosts:ZK port number:/hbase root directory". For e.g. ``zkhost1.example.com,zkhost2.example.com,zkhost3.example.com:2181:/hbase``.

For more details please refer to [HBase book](http://hbase.apache.org/book.html#_cluster_replication).

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/hbase_repl.rb
+++ b/cookbooks/bcpc-hadoop/recipes/hbase_repl.rb
@@ -1,0 +1,19 @@
+#
+# To enable/disable replication peers
+#
+if (node["bcpc"]["hadoop"]["hbase"]["repl"]["enabled"])
+  bash "create_hbase_repl_peer" do
+    code <<-EOH
+      echo "add_peer '#{node["bcpc"]["hadoop"]["hbase"]["repl"]["peer_id"]}','#{node["bcpc"]["hadoop"]["hbase"]["repl"]["target"]}'"|hbase shell
+      EOH
+    not_if "echo list_peers|hbase shell|grep #{node["bcpc"]["hadoop"]["hbase"]["repl"]["peer_id"]}"
+  end
+else
+  bash "stop_repl_remove_repl_peer" do
+    code <<-EOH
+      echo "disable_peer '#{node["bcpc"]["hadoop"]["hbase"]["repl"]["peer_id"]}'"|hbase shell
+      echo "remove_peer '#{node["bcpc"]["hadoop"]["hbase"]["repl"]["peer_id"]}'"|hbase shell
+      EOH
+    only_if "echo list_peers|hbase shell|grep #{node["bcpc"]["hadoop"]["hbase"]["repl"]["peer_id"]}"
+  end
+end
--- a/cookbooks/bcpc-hadoop/templates/default/hb_hbase-site.xml.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/hb_hbase-site.xml.erb
@@ -166,4 +166,9 @@
     <value>org.apache.hadoop.hbase.ipc.SecureRpcEngine</value>    
   </property>   
   <% end %>
+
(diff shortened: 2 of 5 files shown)


cookbooks/bcpc-hadoop/templates/default/hb_hbase-site.xml.erb
```diff
@@ -166,4 +166,9 @@
     <value>org.apache.hadoop.hbase.ipc.SecureRpcEngine</value>    
   </property>   
   <% end %>
+
+  <property>
+    <name>hbase.replication</name>
```
amithkanand: Do we want to put this property only when `default["bcpc"]["hadoop"]["hbase"]["repl"]["enabled"]` is set to `true`.

roles/BCPC-Hadoop-Head-HBase.json
```diff
@@ -4,6 +4,7 @@
   "run_list": [
     "role[Basic]",
     "role[BCPC-Hadoop-Head]",
+    "recipe[bcpc-hadoop::hbase_repl]",
```
amithkanand: Do we want to add this to the run list on VM cluster? There are usually no peer clusters on VMs?

cookbooks/bcpc-hadoop/recipes/hbase_repl.rb
```diff
@@ -0,0 +1,19 @@
+#
+# To enable/disable replication peers
+#
+if (node["bcpc"]["hadoop"]["hbase"]["repl"]["enabled"])
+  bash "create_hbase_repl_peer" do
```
amithkanand: Will it a better idea to use [HBase-Stargate](https://github.com/greglu/hbase-stargate)?

cookbooks/bcpc-hadoop/templates/default/hb_hbase-site.xml.erb
bijugs: The default value for ``base.replication`` in ``hbase`` is ``true``. This is to make it explicit in ``hbase-site`` so that users are aware and doesn't change the default behavior of the cluster.

roles/BCPC-Hadoop-Head-HBase.json
bijugs: The idea is that ``chef-bach`` can be used to create two clusters and users can implement ``hbase`` replication. As acknowledged since the default value is set to ``false`` for the attribute to enable replication, it should not impact any user who is creating a single cluster using ``chef-bach``

cookbooks/bcpc-hadoop/recipes/hbase_repl.rb
bijugs: For the need to enable HBase replication, this simple solution would suffice since it is executed only on the HBase master node. Also this uses the components which are available by default on the ``chef-bach`` stack and maintained by HBase Apache project reducing any complexity/maintenance overhead. Tools like [HBase-Stargate](https://github.com/greglu/hbase-stargate) are good but the tool is not up to date, doesn't support  all the commands and hasn't been maintained for at least 4 years.

amithkanand:
I will do a cluster build with this commit included to make sure everything goes smoothly.

pu239ppy:
Built cluster and verified HBase functionality

Issue #303, opened 2015-10-13
Graphite is broken on latest code


kiiranh:
As of yesterday, the graphite UI and JSON interfaces aren't working with latest chef-bach. I see db related errors for graphite e.g.: `DatabaseError: no such table: auth_user`; that and other tables do exist in mysql. Also, now can get the graphite-mysql password from chef-vault and use it to access the graphite database in mysql. There were two commits recently which chef-vaulted the graphite-mysql password: [chef vault mysql graphite password](https://github.com/bloomberg/chef-bach/commit/7ae4bfca6efc5d8363ffebcff0ce88a951dab27e) and [fix graphite mysql password issue](https://github.com/bloomberg/chef-bach/commit/afff934491925cc970ed493b0f6a016764b270ba). Without those two commits, graphite functions seem to work just fine. I'm guessing the chef-vaulting change should have made more code changes.

kiiranh:
The actual problem was change in permissions on `/opt/graphite/webapp/graphite/local_settings.py`. PR #305 fixes the issue.

PR #305, opened 2015-10-13
Enable graphite-webapp to read local_settings.py to get its configuration

kiiranh:
In ['chef vault mysql graphite password'](https://github.com/bloomberg/chef-bach/commit/7ae4bfca6efc5d8363ffebcff0ce88a951dab27e), the permissions on `/opt/graphite/webapp/graphite/local_settings.py` were changed from `00644` to [`00640`](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc/recipes/graphite.rb#L163). This was done so only the owner of the file can read the config and protect the graphite-mysql password from general browsing. The file was owned by `root:root`. Graphite webapp is run as user `www-data` which reads the local_settings.py file to get its configuration, which after this commit it couldn't do.

This PR makes `www-data` owner of `/opt/graphite/webapp/graphite/local_settings.py`; group is still `root` and changes permissions to `00440` so only root can change the file. This fixes issue #303

### DIFF
--- a/cookbooks/bcpc/recipes/graphite.rb
+++ b/cookbooks/bcpc/recipes/graphite.rb
@@ -158,9 +158,9 @@ end
 
 template "/opt/graphite/webapp/graphite/local_settings.py" do
     source "graphite.local_settings.py.erb"
-    owner "root"
+    owner "www-data"
     group "root"
-    mode 00640
+    mode 00440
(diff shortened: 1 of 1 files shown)


cbaenziger:
Looks as I'd expect

amithkanand:
LGTM.

PR #306, opened 2015-10-20
fixed pytz package name, ensured curl checks if archive exists

pu239ppy:
Fixed a few others that were out there

### DIFF
--- a/build_bins.sh
+++ b/build_bins.sh
@@ -1,4 +1,5 @@
 #!/bin/bash 
+# vim: tabstop=2:shiftwidth=2:softtabstop=2
 
 set -e
 set -x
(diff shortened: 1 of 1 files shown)


build_bins.sh
```diff
@@ -174,7 +177,9 @@ FILES="python-pytz_2015.6_all.deb $FILES"
 
 # build Django 
 if ! [[ -f python-django ]]; then
```
anandkamith: Shouldn't this be ` if ! [[ -f python-django*.deb ]]; then`?

amithkanand:

amithkanand:
LGTM.

PR #309, opened 2015-10-21
Add YARN HA Related parameters to yarn-site.xml

amithkanand:
This PR adds additional `YARN` HA related parameters.


kiiranh:
LGTM

kiiranh:
After discussion with @amithkanand, these are already deployed on clusters and tested.

Issue #310, opened 2015-10-22
Pull Request #200 does not force a spark package update


pu239ppy:
- PR #200 does not force hosts to update spark package to 1.4.1, need to either uninstall existing package to force an reinstall or change the package resource to call :upgrade
- In addition current package naming convention would force an uninstall of existing package, ideally we want to keep multiple binary packages

PR #311, opened 2015-10-22
Spark package name include version and build spark metapackage

pu239ppy:
Purpose of this pull request is to mitigate items mentioned in #310 
- Including spark version in spark package name will allow us to keep multiple versions of spark on the same host
- In addition we are also creating a spark meta package just **spark** which includes a specific spark version as a dependency
- meta package creation and installation is regulated by a Boolean flag so that we do not trample spark binary package on systems where it was already deployed

--Edit--
I kept the meta package option around, however having the version in the package name means that we do not need a metapackage

````text
total 4
drwxrwxr-x 11 root root 4096 Feb 14 06:35 1.4.1
lrwxrwxrwx  1 root root   16 Feb 14 06:35 current -> /usr/spark/1.4.1
````
after making changes to "cookbooks/bach_spark/attributes/default.rb"

...

default[:spark][:package][:version] = "1.6.0"
...

````text
drwxrwxr-x 11 root root 4096 Feb 14 06:35 1.4.1
drwxrwxr-x 11 root root 4096 Feb 14 17:26 1.6.0
lrwxrwxrwx  1 root root   16 Feb 14 17:26 current -> /usr/spark/1.6.0
````

### DIFF
--- a/cookbooks/bach_spark/attributes/default.rb
+++ b/cookbooks/bach_spark/attributes/default.rb
@@ -2,6 +2,7 @@ default[:spark][:download][:url] = "http://d3kbcqa49mib13.cloudfront.net"
 default[:spark][:download][:file][:name] = "spark-1.4.1-bin-hadoop2.6"
 default[:spark][:download][:file][:type] = "tgz"
 default[:spark][:download][:dir] = "/home/vagrant/chef-bcpc/bins"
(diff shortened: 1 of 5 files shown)


cbaenziger:
Yea! Thanks Lev. This has been working well in my testing.

PR #312, opened 2015-10-22
Fix infinite loop in build_bins Ubuntu download

http-418:
Recent changes to build_bins.sh result in an infinite loop, as it checks for a non-existent filename over and over.

This commit moves the filename into a variable to avoid this issue now and in the future.


pu239ppy:
It built

PR #313, opened 2015-10-22
Retain zip files, delete build directories instead

pu239ppy:
- download files as project_master.zip as opposed to master.zip
- keep the zip files after fpm is run
- delete unzipped directories

**Breaking**
Changes archive handling logic in build bins -- so should rerun `build_bins.sh` on a disconnected cluster.


kiiranh:
LGTM

Issue #53, opened 2014-12-17
Chef is grep'ing for HBase services


cbaenziger:
It appears for our [hbase-master service](https://github.com/bloomberg/chef-bach/blob/b7bdfeefe42b52c58a5a15e77ded2d3298da46c2/cookbooks/bcpc-hadoop/recipes/hbase_master.rb#L33-L38) is falling back to walking the process list for the regular expression `/hbase-master/` which means should a user be running a processes like:

````
user1 14724 13749  0 Dec16 pts/4    00:00:00 tail -f hbase-hbase-master-cluster1-r1n8.log
````

Then the Chef run will frustratingly not start the service should it be down; running Chef with debug logging yields:
````
[2014-12-17T17:57:52-05:00] INFO: Processing service[hbase-master] action enable (bcpc-hadoop::hbase_master line 68)
[2014-12-17T17:57:52-05:00] DEBUG: Platform ubuntu version 12.04 found
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] falling back to process table inspection
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] attempting to match 'hbase-master' (/hbase-master/) against process list
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] running: true
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] runlevel 0, action stop, priority 20
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] runlevel 1, action stop, priority 20
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] runlevel 2, action start, priority 20
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] runlevel 3, action start, priority 20
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] runlevel 4, action start, priority 20
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] runlevel 5, action start, priority 20
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] runlevel 6, action stop, priority 20
[2014-12-17T17:57:52-05:00] DEBUG: service[hbase-master] already enabled - nothing to do
[2014-12-17T17:57:52-05:00] INFO: Processing service[hbase-master] action start (bcpc-hadoop::hbase_master line 68)
[2014-12-17T17:57:52-05:00] DEBUG: Platform ubuntu version 12.04 found
[...]

PR #314, opened 2015-10-22
Zabbix operational fixes

kiiranh:
This PR provides the following:

- If one has to do a cluster maintenance/upgrade, there will possibly be unwanted alarms generated by Zabbix. Now we have a flag to disable just the action(alarming) bits in Zabbix while still being able to monitor the system state through Zabbix Items and Triggers. 
- Reduces severity for node and disk going down from "disaster" to a lower severity level.
- Run only one instance of `run_zabbix_sender.sh` at a time
- Add an attribute to specify amount of data sent from graphite to zabbix
- Facilitate delay in executing operations in Zabbix Actions


cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb
```diff
@@ -229,10 +229,16 @@
         end # End of "if (trigger_id = existing_triggers[trigger_name]).nil?"
 
         # Create/Update Actions
+        if node[:bcpc][:hadoop][:zabbix][:enable_alarming] == false
```
cbaenziger: Just a nit but to point it out, you could restructure this to avoid the `== false` and do the expected instead of the negation first; e.g.:
````
if node[:bcpc][:hadoop][:zabbix][:enable_alarming]
  action_status = status
else
  action_status = 1
end
````
Or more concise:
`action_status = node[:bcpc][:hadoop][:zabbix][:enable_alarming] ? status : 1`

File: cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb [hunk 2]
```diff
@@ -306,5 +309,5 @@
   minute "*"
   hour   "*"
   user   "nobody"
-  command  "/usr/local/bin/run_zabbix_sender.sh"
+  command  "pgrep 'zabbix_sender' > /dev/null || /usr/local/bin/run_zabbix_sender.sh"
```
cbaenziger: Do you want to direct `pgrep` to look only for processes run by this user or anything more specific to avoid unexpected interactions? We've hit issues in the past not being terribly specific with Chef services for example, as in Chef-BACH #53.

kiiranh:
@cbaenziger Thanks for calling that out. Fixed as recommended.

cbaenziger:
Looks good!

PR #315, opened 2015-10-23
Change to support version 0.3 of Oozie shell action

bijugs:
Version 0.3 of Oozie shell action XSD supports ``global`` in workflow.xml. This in turn reduces complexity in the workflow xml definition and improves readability. 

Once we move to HDP 2.3, this override in site.xml can be removed since these will be available in the oozie-default.xml.


pu239ppy:
Looks good

PR #317, opened 2015-10-27
Feature to generate kerberos principals/keytabs for non-standard arbitrary host names

amithkanand:
This PR adds functionality to generate `kerberos principals` and `keytabs` for non-standard hosts on the cluster. An example of non-standard host is global `VIP (192.168.100.5)`.


cookbooks/bach_krb5/recipes/keytabs.rb
```diff
@@ -59,5 +104,24 @@
...
+  node[:bcpc][:hadoop][:kerberos][:additional][:hosts].each do |h|
+    node[:bcpc][:hadoop][:kerberos][:data].each do |ke, va|
+
+      service_principal = va['principal']
+      host_fqdn = h
+      keytab_file = "#{h.split(".")[0]}_#{va['principal']}"
+
+      # Create the keytab file
+      execute "creating-keytab-for-#{ke}" do
+        command "kadmin.local -q 'xst -k #{keytab_dir}/#{host_fqdn}/#{keytab_file} -norandkey #{service_principal}/#{host_fqdn}@#{realm} HTTP/#{host_fqdn}@#{realm}'"
```
pu239ppy: if a keytab exists the block will not run, however if it does not do you even care if the KVNO is incremented?

amithkanand:
No longer needed. PR #585  takes care of this.

Issue #307, opened 2015-10-21
Logs are missing for HiveServer2 and Hive Metastore


amithkanand:
In current implementation, `Hive` processes `HiveServer2(HS2)` and `Hive Metastore(MS)` do not generate log files at default log location `/var/log/hive/`.  There were two set of templates that were being used to generate `init` scripts for `HS2` and `MS` and one set of templates were removed as part of [this](https://github.com/amithkanand/chef-bach/commit/469bdca885a2beaf2459c000727e682e0560c43b) commit.  The set of templates that was kept for generating `init` scripts doesn't start the `HS2` and `MS` process with correct set of values to generate log files.

PR #320, opened 2015-10-30
Enable logging/rotation for HiveServer2 and Hive Metastore

amithkanand:
This PR fixes `Hive init` scripts and `Hive log4j.properties` file to generate/roate `HiveServer2` and `Hive MetaStore` logs. This PR resolves issue #307 
Also, this PR adds operation logging for `HiveServer2`.


cbaenziger:
Looks good! (It looks like HDP2.3.2 does not provide Hive/HCatalog start-up scripts from what I can find, so we will be using these a bit longer.)

Issue #308, opened 2015-10-21
Enable HDFS audit logging


kiiranh:
Currently, we do not use HDFS's security/audit logging. This log will be useful in analysis when HDFS goes down due a non-obvious cause.

To enable HDFS audit logging; based on our [log4j.properties](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/templates/default/hdp_log4j.properties.erb), we need to add the following Java variables to the environment variable [HADOOP_NAMENODE_OPTS](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb#L61):
`-Dhadoop.security.logger=INFO,RFAS -Dhdfs.audit.logger=INFO,RFAAUDIT`

Eg: 
```bash
cat /etc/hadoop/conf/hadoop-env.sh
...
export HADOOP_NAMENODE_OPTS="$HADOOP_NAMENODE_OPTS -Xms752m -Xmx752m -XX:ParallelGCThreads=14 -XX:+UseParNewGC -XX:+UseConcMarkSweepGC ... -Dcom.sun.management.jmxremote.port=10111 -Dhadoop.security.logger=INFO,RFAS -Dhdfs.audit.logger=INFO,RFAAUDIT”
...
```
Restart the NameNode services and check the `/var/log/hadoop-hdfs/hdfs-audit.log`.

kiiranh:
Should we tune log [rolling and backup](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/templates/default/hdp_log4j.properties.erb#L122-L135) for hdfs audit and other logs?

kiiranh:
PR merged. Closing this.

PR #321, opened 2015-10-30
Enable HDFS Audit logging

kiiranh:
This PR fixes Issue #308 .

### DIFF
--- a/cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/hdp_hadoop-env.sh.erb
@@ -65,3 +65,5 @@ JMX_OPTS="-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.ssl=fals
 export HADOOP_NAMENODE_OPTS="$HADOOP_NAMENODE_OPTS $JMX_OPTS -Dcom.sun.management.jmxremote.port=<%= @nn_jmx_port %>"
 export HADOOP_DATANODE_OPTS="$HADOOP_DATANODE_OPTS $JMX_OPTS -Dcom.sun.management.jmxremote.port=<%= @dn_jmx_port %>"
 <% end %>
+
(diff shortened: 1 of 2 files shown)


cbaenziger:
I would rather see you modify the [`hdp_log4j.properties.erb`](https://github.com/bloomberg/chef-bach/blob/78316e384c4093ddab81c60d9bac06c990368387/cookbooks/bcpc-hadoop/templates/default/hdp_log4j.properties.erb#L101) file instead of `hdp_hadoop-env.sh.erb`.

kiiranh:
It seems the hdfs-audit logs all hdfs operations (INFO log level), which would be too verbose on an active cluster. I'm trying to find out if these are going to be useful in tracing major issues like HDFS failure of JN/NN formatting.

kiiranh:
Doing it in [hdp_log4j.properties.erb](https://github.com/bloomberg/chef-bach/blob/78316e384c4093ddab81c60d9bac06c990368387/cookbooks/bcpc-hadoop/templates/default/hdp_log4j.properties.erb#L101) means the setting will be used by all hadoop components e.g.: yarn since the template is used in [hadoop_config.rb](https://github.com/bloomberg/chef-bach/blob/a12a1f2c6b9a4bacdd718d26a7e730ad53369c08/cookbooks/bcpc-hadoop/recipes/hadoop_config.rb#L16-L36) included by the recipes for other hadoop components too. At this point we want this enabled only for HDFS.

cbaenziger:
Per recommendations from Hortonworks, we should also set the following in `hdfs-site.xml`:
```
<property>
  <name>dfs.namenode.audit.log.async</name>
  <value>true</value>
</property>
````

kiiranh:
@cbaenziger Made suggested changes. Please note that the `hdfs-audit.log` created after this change will still not track `hdfs dfsadmin` functions, but just the following HDFS functions:
* create
* delete
* getfileinfo
* listStatus
* mkdirs
* open
* rename
* setPermission
* setTimes

So if someone were to execute a `hdfs dfsadmin` function we would know what changed on hdfs (depending on whether the dfsadmin function changes anything), but not who initiated it.

http-418:
This converges and works for me.

kiiranh:
Do we set the Logging level to `WARNING` by default? or leave it as `INFO`?

Issue #323, opened 2015-11-02
build_bins.sh rebuilds graphite related packages every time it runs


pu239ppy:
It does not appear as -f detects file globs so [ -f python-pytz* ] or -f anything* will always fail.
There is no need to rebuild the cluster when testing this code, in fact the issue is that build_bins.sh will try to rebuild packages on subsequent runs.

http-418:
I have been doing this with chef.

Unfortunately, I have not yet integrated the https_intracluster stuff.

https://github.com/http-418/chef-bach/blob/next/cookbooks/bach_repository/recipes/graphite.rb

pu239ppy:
Pull request merged

PR #324, opened 2015-11-03
Fixed an issue where build bins would build certain packages even if they existed

pu239ppy:
This is to address issue #323 
double bracket conditions do not like file globs where as single brackets are work just like test.  Replaced globs with full names where known, and replaced double brackets with singles where globs are still required

### DIFF
--- a/build_bins.sh
+++ b/build_bins.sh
@@ -183,7 +183,7 @@ if ! [[ -f python/pyrabbit-1.0.1.tar.gz ]]; then
 fi
 FILES="pyrabbit-1.0.1.tar.gz $FILES"
 
-if ! [[ -f python-pytz*.deb ]]; then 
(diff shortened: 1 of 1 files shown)


build_bins.sh
```diff
@@ -204,9 +204,9 @@ FILES="python-django_1.5.4_all.deb $FILES"
 
 
 # Build graphite packages
-if ! [[ -f python-carbon_*.deb && \
-        -f python-whisper_*.deb && \
-        -f python-graphite-web_*.deb ]]; then
+if ! [ -f python-carbon_*.deb ] && \
+       [ -f python-whisper_*.deb ] && \
+       [ -f python-graphite-web_*.deb ]; then
```
cbaenziger: I am debating if we care about:
````
bash: [: too many arguments
````

cbaenziger:
Looks good; merged.

PR #326, opened 2015-11-05
pyparsing is needed by the graphite-web

pu239ppy:
Added python-pyparsing to package install section in graphite.rb


kiiranh:
LGTM

PR #332, opened 2015-11-16
Update bcpc-hadoop::maven_config for maven cookbook v2.0.0

http-418:
Maven cookbook v2.0.0 drops maven 2.x support, so all the `node[:maven][:3]` attributes are located at `node[:maven]` instead.

This patch updates bcpc-hadoop::maven_config for changes in v2.0.0

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/maven_config.rb
+++ b/cookbooks/bcpc-hadoop/recipes/maven_config.rb
@@ -2,14 +2,15 @@ require 'pathname'
 
 node.override['maven']['install_java'] = false
 
-internet_download_url = node['maven']['3']['url']
-maven_file = Pathname.new(node['maven']['3']['url']).basename
+internet_download_url = node['maven']['url']
+maven_file = Pathname.new(node['maven']['url']).basename
 
-node.override['maven']['3']['url'] = "#{get_binary_server_url}/#{maven_file}"
+node.override['maven']['url'] = "#{get_binary_server_url}/#{maven_file}"
 
 # download Maven only if not already stashed in the bins directory
 remote_file "/home/vagrant/chef-bcpc/bins/#{maven_file}" do
   source internet_download_url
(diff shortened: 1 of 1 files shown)


cbaenziger:
This is a braking change as an existing cluster needs the `maven` cookbook updated.

cbaenziger:
No, time and code march on...

cookbooks/bcpc-hadoop/recipes/maven_config.rb
```diff
@@ -2,14 +2,15 @@
...
 
-node.override['maven']['3']['url'] = "#{get_binary_server_url}/#{maven_file}"
+node.override['maven']['url'] = "#{get_binary_server_url}/#{maven_file}"
 
 # download Maven only if not already stashed in the bins directory
 remote_file "/home/vagrant/chef-bcpc/bins/#{maven_file}" do
   source internet_download_url
-  action :create_if_missing
-  mode '0555'
+  action :create
```
cbaenziger: On a disconnected cluster would this cause grumpiness? If so perhaps the URL variable should not be `internet_download_url` but maven tarball URL or some such (implying one should set it to a local URL)?

cbaenziger: For [reference](https://docs.chef.io/resource_remote_file.html):
````
:create 	Default. Create a file. If a file already exists (but does not match), update that file to match.
:create_if_missing 	Create a file only if the file does not exist. When the file exists, nothing happens.
````

http-418: I added a checksum using the attribute provided by the maven cookbook.

If the file exists, and the checksum fits, it won't reach out to the internet.
If the file doesn't match the checksum, grumpiness is desired behavior.

cbaenziger:
Cool; thanks!

PR #337, opened 2015-11-18
Fix up bcpc-hadoop metadata for new maven cookbook

http-418:
I updated bcpc-hadoop with fixes for v2.0.0 of the Maven cookbook in a previous PR.  This PR fixes the metadata.rb to reflect the new requirements.


cbaenziger:
Mind squashing commits Andrew? Looks good otherwise; thanks for the quote/author fix up too!

cbaenziger:
This is a breaking change as the Maven cookbook needs to rev on already existing clusters.

PR #340, opened 2015-11-20
adjusted yarn client fail over timers to fail over between RMs faster

pu239ppy:
Currently in environments with two or more ResourceManagers a yarn client will attempt to connect to a list of hosts corresponding to a list of ResourceManager IDs in 
````xml
  <property>
    <name>yarn.resourcemanager.ha.rm-ids</name>
    <value>rmMyCluster1,rmMyCluster2</value>
  </property>
````
The client attempts a connection in port 8032 and if the ResourceManager is currently in standby will receive a ConnectionRefused as standby ResourceManagers do not listen on the RPC port.  The proposed request will limit the amount of time it takes for the yarn client to wait between retries and thus speed up the selection process. This does not affect the failover and leader election logic of the ResourceManagers.

Below is the explanation for this property
````
When HA is enabled, the sleep base (in milliseconds) to be used for calculating the exponential delay between failovers. When set, this overrides the yarn.resourcemanager.connect.* settings. When not set, yarn.resourcemanager.connect.retry-interval.ms is used instead.
````

The right behavior will eventually be introduced in https://issues.apache.org/jira/browse/YARN-4496


cbaenziger:
Lev did you test on VM's just to make sure the latency will not break VM build tests hopefully?

pu239ppy:
Clay why would this break VM build tests?  This only affects how quickly yarn client switches to the next ResourceManager in the list when the the current one produces a connection refused

pu239ppy:
For VM environment set the sleep time to 30 seconds

PR #222, opened 2015-08-03
Routed Rack Configuration

cbaenziger:
For folks using layer 3 networking between racks of hosts, this PR supports using BFD to register VIP endpoints and setting up the various data structures to be multi-subnet aware.

PR #343, opened 2015-12-01
Changes for Zabbix to work with Routed Rack

kiiranh:
The network changes that are proposed as part of PR #222 mean Zabbix won't work as expected. This is because:

- VIP won't be attached to the `node[:bcpc][:management][:interface]` and so `has_vip?` check will fail (Thus none of the Zabbix servers will be active w.r.t configuring zabbix and sending metrics to zabbix server)
- In Routed Rack, zabbix_sender's traffic will appear to be coming from the Global VIP and not the host local IP. Thus Zabbix items will reject the metric information since they expect that info to come only from [list of head/graphite nodes' local IPs](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb#L175)

This PR proposes changes which will enable Zabbix to work with Routed Racks too.

### DIFF
--- a/cookbooks/bcpc-hadoop/libraries/utils.rb
+++ b/cookbooks/bcpc-hadoop/libraries/utils.rb
@@ -391,7 +391,7 @@ end
 
 def has_vip?
   cmd = Mixlib::ShellOut.new(
-    "ip addr show dev #{node[:bcpc][:management][:interface]}", :timeout => 10
+    "ip addr show", :timeout => 10
   ).run_command
   cmd.stderr.empty? && cmd.stdout.include?(node[:bcpc][:management][:vip])
 end
--- a/cookbooks/bcpc-hadoop/templates/default/zabbix.run_zabbix_sender.sh.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/zabbix.run_zabbix_sender.sh.erb
@@ -1,5 +1,5 @@
 #!/bin/bash
-if ip addr show dev <%=node[:bcpc][:management][:interface]%> | grep "<%=node[:bcpc][:management][:vip]%>" > /dev/null
+if ip addr show | grep "<%=node[:bcpc][:management][:vip]%>" > /dev/null
 then
   python /usr/local/bin/query_graphite.py | /usr/local/bin/zabbix_sender -z <%= node[:bcpc][:management][:vip] %> -p <%= node[:bcpc][:zabbix][:server_port] %> -T -i - > /dev/null
 fi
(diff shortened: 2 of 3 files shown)


cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -391,7 +391,7 @@ def get_group_action(group_name)
 
 def has_vip?
   cmd = Mixlib::ShellOut.new(
-    "ip addr show dev #{node[:bcpc][:management][:interface]}", :timeout => 10
+    "ip addr", :timeout => 10
```
cbaenziger: I would suggest to use `ip addr list` instead of `ip addr` to be more explicit that we want a listing of all addresses.

cookbooks/bcpc-hadoop/templates/default/zabbix.run_zabbix_sender.sh.erb
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-if ip addr show dev <%=node[:bcpc][:management][:interface]%> | grep "<%=node[:bcpc][:management][:vip]%>" > /dev/null
+if ip addr | grep "<%=node[:bcpc][:management][:vip]%>" > /dev/null
```
cbaenziger: Again perhaps `ip addr list` instead of `ip addr` to be more explicit.

cookbooks/bcpc-hadoop/libraries/utils.rb
kiiranh: @cbaenziger Thanks for the quick review. To confirm, you meant we should have `ip addr show` in place of just `ip addr`? Because `ip addr help` shows there is no `list` option:

```bash
kiiranh@bcpc-vm1~$ ip addr help
Usage: ip addr {add|change|replace} IFADDR dev STRING [ LIFETIME ]
                                                      [ CONFFLAG-LIST ]
       ip addr del IFADDR dev STRING
       ip addr {show|flush} [ dev STRING ] [ scope SCOPE-ID ]
                            [ to PREFIX ] [ FLAG-LIST ] [ label PATTERN ]
IFADDR := PREFIX | ADDR peer PREFIX
          [ broadcast ADDR ] [ anycast ADDR ]
          [ label STRING ] [ scope SCOPE-ID ]
SCOPE-ID := [ host | link | global | NUMBER ]
FLAG-LIST := [ FLAG-LIST ] FLAG
FLAG  := [ permanent | dynamic | secondary | primary |
           tentative | deprecated | dadfailed | temporary |
           CONFFLAG-LIST ]
CONFFLAG-LIST := [ CONFFLAG-LIST ] CONFFLAG
CONFFLAG  := [ home | nodad ]
LIFETIME := [ valid_lft LFT ] [ preferred_lft LFT ]
LFT := forever | SECONDS

```

PR #346, opened 2015-12-07
Zabbix leader election

kiiranh:
This PR provides Zabbix leader election using the shared MySQL database already being deployed on the cluster. Zabbix server processes will continue running on all the head nodes, but only one Head node will act as a Zabbix master at a given time. Here, Zabbix master refers to the node which assumes [responsibility of configuring Zabbix items/triggers/actions](https://github.com/bloomberg/chef-bach/blob/2d8720f76141f5562b6f0ef58560b7e1ba52599d/cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb#L34-L308) and running the [cron job for zabbix_sender](https://github.com/bloomberg/chef-bach/blob/2d8720f76141f5562b6f0ef58560b7e1ba52599d/cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb#L310-L315) which pushes graphite metric data to Zabbix server. On the non-zabbix-master, the cron job will be deleted and the `bcpc-hadoop::zabbix_monitor` chef resource won't run.

### DIFF
--- a/cookbooks/bcpc/recipes/zabbix-head.rb
+++ b/cookbooks/bcpc/recipes/zabbix-head.rb
@@ -20,8 +20,7 @@
 include_recipe "bcpc::mysql"
 include_recipe "bcpc::apache2"
 
-mysql_zabbix_user = 'zabbix'
-make_config('mysql-zabbix-user', mysql_zabbix_user)
+make_config('mysql-zabbix-user','zabbix')
 
 # backward compatibility
 mysql_zabbix_password = get_config("mysql-zabbix-password")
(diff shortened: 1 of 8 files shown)


cookbooks/bcpc/recipes/zabbix-head.rb
```diff
@@ -20,7 +20,8 @@
 include_recipe "bcpc::mysql"
 include_recipe "bcpc::apache2"
 
-make_config('mysql-zabbix-user', "zabbix")
+mysql_zabbix_user = 'zabbix'
```
cbaenziger: I would not recommend directly setting this variable. First, `make_config()` the user and `get_config!()` the user anywhere you need the string; you could cache the result of `get_config!()` in `mysql_zabbix_user` though.

pu239ppy:
Looks good

PR #348, opened 2015-12-09
Use an updated ubuntu iso which pxe boots faster

kiiranh:
This reduces the pxe boot time for a VM from 20-25 minutes to less than 10min.


http-418:
I checked this last night.  It took the Ubuntu installation step from 43 minutes to *12*.  I'm real happy to merge this.

PR #349, opened 2015-12-11
Update Zabbix and ZabbixAPI to version 2.4.5

kiiranh:
This installs the latest Zabbix (v.2.4.5) server, agent and api.

### DIFF
--- a/build_bins.sh
+++ b/build_bins.sh
@@ -6,8 +6,8 @@ set -x
 
 # Define the version of Zabbix server and zabbixapi gem to be downloaded
 # Refer https://github.com/bloomberg/chef-bcpc/issues/343
-ZABBIXAPI_VERSION=2.2.2
-ZABBIX_VERSION=2.2.9
+ZABBIXAPI_VERSION=2.4.5
+ZABBIX_VERSION=2.4.5
 
 # Define the appropriate version of each binary to grab/build
 VER_KIBANA=d1495fbf6e9c20c707ecd4a77444e1d486a1e7d6


kiiranh:
On an existing cluster, one will have to place the following files on the `bootstrap` node and run `build_bins.sh`:
```bash
vagrant@bcpc-bootstrap:~/chef-bcpc$ ls -lrt bins/ | grep -i zabbix
-rwxr-xr-x 1 root    root       918888 Dec 10 22:43 zabbix-agent.tar.gz
-rwxr-xr-x 1 root    root      7136644 Dec 10 22:43 zabbix-server.tar.gz
vagrant@bcpc-bootstrap:~/chef-bcpc$ ls -lrt bins/gems/ | grep -i zabbix
-rwxr-xr-x 1 root root  21504 Dec 10 22:41 zabbixapi-2.4.5.gem
lrwxrwxrwx 1 root root     19 Dec 10 22:41 zabbixapi.gem -> zabbixapi-2.4.5.gem
```

To actually upgrade Zabbix, following process will have to be followed:
1. Stop chef-client on the head nodes
2. Stop Zabbix services: `zabbix-server` and `zabbix-agent`
3. [Uninstall Zabbix]( https://www.zabbix.com/forum/showthread.php?t=5450)
4. Update the `bootstrap` node as mentioned above
5. Run chef-client on the head nodes one at a time. This should setup Zabbix on the head nodes.

amithkanand:
LGTM.

PR #352, opened 2015-12-16
disabled logging for cache and query related events

pu239ppy:
Most notably will not add entries to the update log


mlongob:
LGTM - This seems to be the recommended way to configure logging as many people are complaining about how noisy these updates can be.

amithkanand:
LGTM

PR #356, opened 2015-12-22
METRIC_CLIENT_IDLE_TIMEOUT was in the wrong place

pu239ppy:
This setting needs to be under [relay]


cookbooks/bcpc/templates/default/carbon.conf.erb
```diff
@@ -43,17 +43,16 @@ USE_RATIO_RESET=False
 MIN_RESET_STAT_FLOW=1000
 MIN_RESET_RATIO=0.9
 MIN_RESET_INTERVAL=121
+METRIC_CLIENT_IDLE_TIMEOUT = <%="#{node[:bcpc][:graphite][:carbon][:relay][:idle_timeout]}"%>
```
kiiranh: Nit: The `"#{}"` is not needed in an erb file. We can directly use `node[...`

kiiranh:
LGTM

PR #363, opened 2015-12-29
Improve graphite-to-zabbix flow

kiiranh:
This PR provides the following:

- Changes the default data type for a Zabbix item from numeric float to numeric unsigned. This was done because numeric float cannot hold values greater than [double(16,4)](https://www.zabbix.com/documentation/2.2/manual/config/items/item) and the size(in bytes) for disks could be in Tera Bytes which is incompatible for this data type. Metric values from graphite are of the form `[0-9]+.0` and so we don't loose anything while converting the value to an int (in python) and it should be compatible for the type and range of values we expect to see in the metrics we use.
- Add a trigger dependency between `"#{host}_#{disk}_SpaceUsed"` and `"#{host}_NodeAvailability"`
- Add logging to `query_graphite.py` so one has visibility into when the script ran, for how long and if there were any errors.

### DIFF
--- a/cookbooks/bcpc-hadoop/templates/default/graphite.query_graphite.py.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/graphite.query_graphite.py.erb
@@ -3,9 +3,26 @@ import urllib
 import urllib2
 import json
 import sys
+import logging
+from logging.handlers import RotatingFileHandler
+
+LOGGER = None
+
+def create_log(path):
+  """  
+  Creates a rotating log
+  """
+  global LOGGER
+  LOGGER = logging.getLogger('query_graphite')
+  LOGGER.setLevel(logging.<%= @log_level %>)
+  handler = RotatingFileHandler(path, maxBytes=<%= @max_bytes %>, backupCount=<%= @backup_count %>)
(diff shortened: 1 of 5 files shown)


mlongob:
LGTM

cookbooks/bcpc-hadoop/templates/default/graphite.query_graphite.py.erb
```diff
@@ -0,0 +1,89 @@
...
+
+LOGGER = None
+
+def create_log(path):
+  """  
+  Creates a rotating log
+  """
+  global LOGGER
+  LOGGER = logging.getLogger('query_graphite')
+  LOGGER.setLevel(logging.<%= node[:bcpc][:hadoop][:zabbix][:query_graphite][:logging_level] %>)
```
http-418: Kiran, do you suppose you could pull these 'node' object references into local variables on the template resource?

e.g.
```ruby
template '/usr/local/bin/thing.py' do
  variables({:log_level => node[:bcpc][:hadoop][:zabbix][:query_graphite][:logging_level]})
end
```

kiiranh:
@http-418 Thanks for the suggestion. Incorporated changes and rebased.

PR #366, opened 2016-01-07
Correc the Yarn site property resource.percentage-physical-cpu-limit 

bijugs:
The supported yarn-site property name in is yarn.nodemanager.resource.percentage-physical-cpu-limit.

### DIFF
--- a/cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb
@@ -238,7 +238,7 @@
   </property>
 
   <property>
-    <name>resource.percentage-physical-cpu-limit</name>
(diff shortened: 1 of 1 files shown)


http-418:
Looks good to me.

pu239ppy:
Added breaking as it it causes kernel panics on stock 12.04

cbaenziger:
BREAKING: Until a newer kernel is installed (e.g. as per PR #377 ) this can cause your NM machines to kernel panic the mechanism is not fully understood.

It looks like YARN may trigger the kernel panics per Canonical due to walking `/proc`. Past examples of YARN and kernel interactions are (but seem unrelated):
* YARN-2809 - Implement workaround for linux kernel panic when removing cgroup
* YARN-4048 - Linux kernel panic under strict CPU limits
Further, it was found the following was not the issue per a hotfix attempt:
[LP #1534413](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1534413)

Issue #354, opened 2015-12-18
Add hbase jars to Oozie


bijugs:
Currently Oozie doesn't get shipped with HBase jars. This I believe is to keep check on the Oozie war size. This enhancement is to add HBase jars to Oozie. 

Following are the steps to add HBase jars to Oozie: The version used here is HDP 2.2

**Prepare Oozie war**
- cp /usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie.war /tmp/
- mkdir /tmp/oozie-hbase-prep
- cd /tmp/oozie-hbase-prep
- jar -xf ../oozie.war
- cd /usr/hdp/2.2.0.0-2041/hbase/lib/
- cp hbase-common-0.98.4.2.2.0.0-2041-hadoop2.jar hbase-client-0.98.4.2.2.0.0-2041-hadoop2.jar - hbase-server-0.98.4.2.2.0.0-2041-hadoop2.jar hbase-protocol-0.98.4.2.2.0.0-2041-hadoop2.jar hbase-hadoop2-compat-0.98.4.2.2.0.0-2041-hadoop2.jar htrace-core-*.jar netty-3.6.6.Final.jar /tmp/oozie-hbase-prep/WEB-INF/lib/.
- cd /tmp/oozie-hbase-prep
- jar -cMf ../oozie.war *

**Deploying updated war to Oozie**
- service oozie stop
- Edit /etc/oozie/conf/oozie-site.xml and edit the property oozie.credentials.credentialclasses to include hbase credential handler class "hbase=org.apache.oozie.action.hadoop.HbaseCredentials"
- cp /usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie.war /tmp/oozie.war1
- rm -fr /usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie.war
- cp /tmp/oozie.war /usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie.war
- rm -rf /usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie
- service oozie start

bijugs:
This is a sample recipe. This doesn't include the step to modify oozie-site.xml. Obviously it need bit more enhancement wrt to be idempotent. Need to be run on the node where oozie server runs.

```
#
# Recipe to include hbase jar files into Oozie
#
%w{
hbase
}.each do |p|
  package p do
    action :install
  end
end

[...]

PR #369, opened 2016-01-08
Changes to include hbase jars into Oozie war and deploy to Oozie server

bijugs:
This is for enhancement #354 . This will help users to make Hbase related requests through Oozie.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
+++ b/cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
@@ -1,0 +1,98 @@
+#
+# Recipe to include hbase jar files into Oozie
+#
+%w{
+hbase
+}.each do |p|
+  package p do
+    action :install
+  end
+end
+
+file "/tmp/oozie.war" do
+  owner 'root'
+  group 'root'
+  mode 0744
+  content ::File.open("/usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie.war").read
+  action :create
+end
+
+directory '/tmp/oozie-hbase-prep' do
+  owner 'root'
+  group 'root'
+  mode '0755'
+  action :create
+end
+
+bash 'extract-oozie-war' do
+  user "root"
+  group "root"
+  code <<-EOH
+    cd /tmp/oozie-hbase-prep
+    jar -xf ../oozie.war
+  EOH
+  action :run
+end
+
+bash 'copy-hbase-jars' do
+  user "root"
+  group "root"
+  code <<-EOH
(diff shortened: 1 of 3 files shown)


pu239ppy:
I am going to attempt to build this on a test VM and replicate the workflow

pu239ppy:
Looks good

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
```diff
@@ -0,0 +1,98 @@
+#
+# Recipe to include hbase jar files into Oozie
+#
+%w{
+hbase
```
cbaenziger: Can we use Maven or something lighter weight, instead of installing the HBase package?

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
```diff
@@ -0,0 +1,98 @@
...
+  group 'root'
+  mode '0755'
+  action :create
+end
+
+bash 'extract-oozie-war' do
+  user "root"
+  group "root"
+  code <<-EOH
+    cd /tmp/oozie-hbase-prep
```
cbaenziger: This can use the `:cwd` attribute of the [bash](https://docs.chef.io/resource_bash.html#properties) resource.

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
```diff
@@ -0,0 +1,98 @@
...
+    jar -xf ../oozie.war
+  EOH
+  action :run
+end
+
+bash 'copy-hbase-jars' do
```
cbaenziger: Every time the machine restarts this will cause us to rebuild the Oozie WAR since this uses `/tmp`. Could you use a less raceable directory choice? (E.g. [Dir.mktmpdir](http://www.rubydoc.info/stdlib/tmpdir/Dir) or `Chef::Config[:file_cache_path]`.)

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
bijugs: Would recommend to use the HBase jars shipped with the version of HDP being installed on the cluster.

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
bijugs: The Oozie war will not be rebuilt if the current war already includes hbase jars due to the ``not_if`` condition.

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
```diff
@@ -0,0 +1,98 @@
...
+#
+%w{
+hbase
+}.each do |p|
+  package p do
+    action :install
+  end
+end
+
+file "/tmp/oozie.war" do
```
kiiranh: Why do we need to copy the war to a temp directory? We could just unzip the jar from its original location into the temp directory.

PR #373, opened 2016-01-12
Add hbase property hbase.coprocessor.abortonerror

bijugs:
The property ``hbase.coprocessor.abortonerror`` will help recover HBase cluster gracefully if there was an issue with a co-processor that got deployed to the cluster. The default value is ``true`` and to recover it can be set to ``false`` so that the HBase master and RS will come-up so that the issue can be fixed.


amithkanand:
LGTM.

PR #376, opened 2016-01-13
enable Zookeeper NodeCount via JMX

pu239ppy:
Report NodeCount in graphite console


cbaenziger:
@pu239ppy Could you add a Zabbix alarm for when this differs from `QuorumSize`?

PR #377, opened 2016-01-13
Provide a way to install a newer kernel

kiiranh:
This PR:

- Adds an attribute to specify a kernel to install
- Installs the latest trusty kernel by default
- Installs and configures linux kernel crash debugging tools


pu239ppy:
Looks good

cookbooks/bcpc/templates/default/cobbler.bcpc_ubuntu_host.preseed.erb
```diff
@@ -94,4 +94,7 @@ d-i     debian-installer/exit/poweroff  boolean false
 d-i     pkgsel/include string openssh-server lldpd
 
 d-i     preseed/late_command string <%= @node[:bcpc][:bootstrap][:preseed][:late_command] %> && \
+        in-target apt-get -y install linux-crashdump kdump-tools && \
+        in-target sed -i 's/USE_KDUMP=0/USE_KDUMP=1/' /etc/default/kdump-tools; \
```
cbaenziger: Sorry to be late to the party but this should be a `&&` to fail the `late_command` if any step fails. Right now this could succeed and then we would still pull the machine out of Cobbler despite an earlier failure.

PR #378, opened 2016-01-13
Fix preseed late_command logic

kiiranh:
This fixes the issue mentioned in PR #377.


cbaenziger:
Thank you; this should work well.

PR #375, opened 2016-01-13
Reduce root disk io (BREAKING CHANGE)

pu239ppy:
Reserve a disk on head node to be used by graphite and carbon, this is to address #374 
**BREAKING**
In order for graphite to use the /disk/x disk graphite.rb must run after disks.rb, hence the run list change for BCPC-Hadoop-Head.json

PR #380, opened 2016-01-15
Dynamic disk reservations with multiple roles

pu239ppy:
Follow up and a list of improvements to #375 

**Changes**
- No longer need to specify a disk for each reservation the element index in the reservation list will be used to reserve the disk in available disks
- Multiple roles may be now specified to enable this logic
- disk setup has been moved out of the environment into attributes
- disks recipe modified
- graphite recipe does not need to be aware of extra directory and or symlink work, it is handled in a wrapper

**Usage**
This recipe reserves disks by modifying how ````node.set[:bcpc][:hadoop][:mounts]```` is calculated, if no reservations are defined under ````default[:bcpc][:hadoop][:disks][:reservation_requests]```` the math will remain the same since the subtraction of an empty list from a list returns the same list

**Testing**
````ruby
# enumerate available disks and how they will be used
default[:bcpc][:hadoop][:disks] = {
  :available_disks => ["sdb" , "sdc", "sdd", "sde"],
  # keep at least that many disks for the :disk_exclude_role
  :role_min_disk => 2,
  # we are reserving disks for the following
  :reservation_requests => ["graphite_disk"],
  # reservations will be saved here
  :disk_reserve_roles => ["BCPC-Hadoop-Head"]
}
````
````
ubuntu@bcpc-vm1:/opt$ ls -l
graphite -> /disk/0/graphite_disk
````

**Breaking**
Deploying this change on an existing system will attempt to relocation graphite to /disk0, as graphite may already exist under /opt this change may leave a graphite installation in an inconsistent state.  To avoid this set override ````default[:bcpc][:hadoop][:disks][:reservation_requests]```` to an empty array is the environment file
````
"bcpc: {
  "hadoop": {
...
      "disks": {
        "reservation_requests": []
      },
...
  }
}
````
if this will be used then `bcpc-hadoop::graphite_disk_wrapper.rb` need to be added to the runlist

The environment will need to overwrite `:available_disks`
````ruby
default[:bcpc][:hadoop][:disks] = {
  :available_disks => ["sdb" , "sdc", "sdd", "sde"]
}```

### CODE REVIEW

File: cluster-assign-roles.sh [hunk 1]
```diff
@@ -85,7 +85,7 @@ function install_stub {
     local role="${BASH_REMATCH[1]}"
     local ip="${BASH_REMATCH[2]}"
     local fqdn="${BASH_REMATCH[3]}"
-    sudo knife node show $fqdn $KNIFE_ADMIN 2>/dev/null >/dev/null ||  install_machines "role[Basic],recipe[bcpc::default],recipe[bcpc::networking]${BANG}${ip}${BANG}${fqdn}" &
+    sudo knife node show $fqdn $KNIFE_ADMIN 2>/dev/null >/dev/null ||  install_machines "role[Basic],recipe[bcpc::default],recipe[bcpc::chef_vault_install],recipe[bcpc::networking]${BANG}${ip}${BANG}${fqdn}" &
```
kiiranh: Do we need `recipe[bcpc::chef_vault_install]` in the run_list here?

cookbooks/bcpc/recipes/zabbix-head.rb
```diff
@@ -225,3 +231,28 @@
...
+  end
+  action :nothing
+end
+
+service "apache2" do
+  action :restart
+  only_if { node.run_state['restart_apache2_needed']  == true }
+end
+
+service "zabbix-server" do
```
kiiranh: We already have a `service "zabbix-server"` resource defined [earlier](https://github.com/pu239ppy/chef-bach/blob/multiple_roles_for_disk_exclusion/cookbooks/bcpc/recipes/zabbix-head.rb#L160-L163) in the same recipe and I'd assume that this definition overwrites the previous one. We should have just one resource for enable/start/restart zabbix-server to keep things simple.

cookbooks/bcpc/recipes/zabbix-head.rb
```diff
@@ -202,6 +207,7 @@
   EOH
   not_if "test -r /etc/apache2/sites-enabled/zabbix-web"
   notifies :restart, "service[apache2]", :delayed
+  notifies :run, "ruby_block[run_state_apache2_restart]", :immediate
```
kiiranh: This obviates the previous line which should make sure zabbix-server is restarted towards the end of the chef-client run. If we want a resource to trigger an immediate restart of zabbix-server, it would be good to use an immediate restart notification instead of keeping this additional state.

cookbooks/bcpc/recipes/zabbix-head.rb
```diff
@@ -225,3 +231,28 @@
...
+end
+
+ruby_block "run_state_zabbix-server_restart" do
+  block do
+    node.run_state['restart_zabbix-server_needed'] = true
+  end
+  action :nothing
+end
+
+service "apache2" do
```
kiiranh: I think this would overwrite the resource definition `service "apache2"` in the apache2 cookbook's recipe. Not sure if its a good idea to re-define a resource here again.

cookbooks/bcpc-hadoop/recipes/graphite_disk_wrapper.rb
```diff
@@ -0,0 +1,18 @@
+# assign prepare a direcotry structure for graphite to install itself to
+reservation_requests = node[:bcpc][:hadoop][:disks][:reservation_requests]
+
+if reservation_requests.include?("graphite_disk") then
+  disk_index = reservation_requests.index("graphite_disk")
+
+  directory "/disk/#{disk_index}/graphite_disk" do
```
kiiranh: Nit: Just out of curiosity, why the name `graphite_disk` instead of just `graphite`? With this code the effect will be `/opt/graphite -> /disk/0/graphite_disk`. Would it be better to have `/opt/graphite -> /disk/0/graphite`?

kiiranh:
A lot of code was merged since this PR was opened. A rebase would be good.

cookbooks/bcpc-hadoop/recipes/graphite_disk_wrapper.rb
pu239ppy: More descriptive since we are supposedly reserving a whole disk disk, though in the future we may perhaps have allowances to colocate technologies on same disk by having a list of lists.

cookbooks/bcpc-hadoop/recipes/disks.rb
```diff
@@ -31,17 +36,22 @@
       action [:enable, :mount]
     end
   end
-  if node.roles.include? "BCPC-Hadoop-Head" and node[:bcpc][:hadoop][:head_disks] != nil then
-    mnt_idxs = node[:bcpc][:hadoop][:head_disks].map do |dsk| 
-      node[:bcpc][:hadoop][:disks].index(dsk)
+
+  # is our role included in the list
+  if not (node[:bcpc][:hadoop][:disks][:disk_exclude_roles] & node.roles).empty? then
```
amithkanand: Shouldn't you use `&&` for logical `AND` as mentioned [here](http://www.tutorialspoint.com/ruby/ruby_operators.htm).

pu239ppy: This is a set intersection
````
irb(main):006:0> [1,2,3] & [3,4,5]
=> [3]
````

File: cookbooks/bcpc-hadoop/attributes/disks.rb [hunk 7]
```diff
@@ -0,0 +1,10 @@
+# enumerate available disks and how they will be used
+default[:bcpc][:hadoop][:disks] = {
+  :available_disks => ["sdb" , "sdc", "sdd", "sde"],
+  # keep at least that many disks for the :disk_exclude_role
+  :role_min_disk => 2,
+  # we are reserving disks for the following
+  :reservation_requests => ["graphite_disk"],
+  # reservations will be saved here
+  :disk_exclude_roles => ["BCPC-Hadoop-Head"]
```
ekund: Why is the variable named disk_exclude_roles?  It sounds like the roles listed will be excluded from the reservation request.  It would be better to have a name like remaining_disks_roles.

cookbooks/bcpc-hadoop/recipes/graphite_disk_wrapper.rb
```diff
@@ -0,0 +1,18 @@
+# assign prepare a direcotry structure for graphite to install itself to
```
ekund: Nit: should say "directory"

cookbooks/bcpc-hadoop/attributes/disks.rb
pu239ppy: I am naming this disk_reserve_roles

cookbooks/bcpc-hadoop/attributes/disks.rb
```diff
@@ -0,0 +1,10 @@
+# enumerate available disks and how they will be used
+default[:bcpc][:hadoop][:disks] = {
+  :available_disks => ["sdb" , "sdc", "sdd", "sde"],
+  # keep at least that many disks for the :disk_exclude_role
```
ekund: Nit: the comment mentions :disk_exclude_role but it's been renamed.

cookbooks/bcpc-hadoop/recipes/disks.rb
```diff
@@ -1,16 +1,21 @@
+# vim: tabstop=2:shiftwidth=2:softtabstop=2
```
cbaenziger: Erm?

pu239ppy: Sets up spacing for vim users, emacs users are welcome to do so as well.  Good for enforcing code standard.

PR #381, opened 2016-01-15
Add HBase jars to Oozie server

bijugs:
This is for enhancement #354 . This will help users to make Hbase related requests through Oozie. These changes were earlier included in PR #369 . But VM cluster build failed with the latest master. This PR is with the latest master.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
+++ b/cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
@@ -1,0 +1,96 @@
+#
+# Recipe to include hbase jar files into Oozie
+#
+package "hbase" do
+  action :upgrade
+end
+
+bash "/tmp/oozie.war" do
+  user "root"
+  group "root"
+  code <<-EOH
+    cp /usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie.war /tmp/oozie.war
+  EOH
+  action :run
+end
+
+directory '/tmp/oozie-hbase-prep' do
+  owner 'root'
+  group 'root'
+  mode '0755'
+  action :create
+end
+
+bash 'extract-oozie-war' do
+  user "root"
+  group "root"
+  cwd "/tmp/oozie-hbase-prep"
+  code <<-EOH
+    jar -xf ../oozie.war
+  EOH
+  action :run
+end
+
+bash 'copy-hbase-jars' do
+  user "root"
+  group "root"
+  code <<-EOH
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/hbase-common-0.98.4.2.2.0.0-2041-hadoop2.jar /tmp/oozie-hbase-prep/WEB-INF/lib/hbase-common-0.98.4.2.2.0.0-2041-hadoop2.jar
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/hbase-client-0.98.4.2.2.0.0-2041-hadoop2.jar /tmp/oozie-hbase-prep/WEB-INF/lib/hbase-client-0.98.4.2.2.0.0-2041-hadoop2.jar
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/hbase-server-0.98.4.2.2.0.0-2041-hadoop2.jar /tmp/oozie-hbase-prep/WEB-INF/lib/hbase-server-0.98.4.2.2.0.0-2041-hadoop2.jar
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/hbase-protocol-0.98.4.2.2.0.0-2041-hadoop2.jar /tmp/oozie-hbase-prep/WEB-INF/lib/hbase-protocol-0.98.4.2.2.0.0-2041-hadoop2.jar
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/hbase-hadoop2-compat-0.98.4.2.2.0.0-2041-hadoop2.jar /tmp/oozie-hbase-prep/WEB-INF/lib/hbase-hadoop2-compat-0.98.4.2.2.0.0-2041-hadoop2.jar
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/htrace-core-2.04.jar /tmp/oozie-hbase-prep/WEB-INF/lib/htrace-core-2.04.jar
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/htrace-core-3.0.4.jar /tmp/oozie-hbase-prep/WEB-INF/lib/htrace-core-3.0.4.jar
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/netty-3.6.6.Final.jar /tmp/oozie-hbase-prep/WEB-INF/lib/netty-3.6.6.Final.jar
+  EOH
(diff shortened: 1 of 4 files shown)


cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+end
+
+bash "/tmp/oozie.war" do
+  user "root"
+  group "root"
+  code <<-EOH
+    cp /usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie.war /tmp/oozie.war
+  EOH
+  action :run
```
pu239ppy: I'd put a guard here for file availability

pu239ppy:
I am still getting the same error
Error executing action `create_if_missing` on resource 'chef_vault_secret[mysql-root]'


ChefVault::Exceptions::SecretDecryption
---------------------------------------
os/mysql-root is not encrypted with your public key.  Contact an administrator of the vault item to encrypt for you!


Cookbook Trace:
---------------
/var/chef/cache/cookbooks/chef-vault/libraries/chef_vault_secret_provider.rb:67:in `load_current_resource'
/var/chef/cache/cookbooks/bcpc/recipes/mysql.rb:39:in `from_file'


Resource Declaration:
---------------------
# In /var/chef/cache/cookbooks/bcpc/recipes/mysql.rb

 33: chef_vault_secret "mysql-root" do
 34:   data_bag 'os'
 35:   raw_data({ 'password' => mysql_root_password })
 36:   admins "#{ nodes },#{ bootstrap }"
 37:   search '*:*'
 38:   action :nothing
 39: end.run_action(:create_if_missing)
 40: 
 41: make_config('mysql-galera-user', "sst")
 42: 
 43: # backward compatibility
 44: mysql_galera_password = get_config("mysql-galera-password")
 45: if mysql_galera_password.nil?
 46:   mysql_galera_password = secure_password
 47: end
 48:

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
bijugs: We want the chef-client run to fail if the oozie.war is not found. Since this recipe need to run after the ```oozie`` recipe this step should not fail. If it fails we have a bigger issue which need to be looked at.

bijugs:
@pu239ppy . Sorry for providing the incorrect branch name for you to build the new cluster.

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
```diff
@@ -0,0 +1,100 @@
...
+    jar -xf ../oozie.war
+  EOH
+  action :run
+end
+
+bash 'copy-hbase-jars' do
+  user "root"
+  group "root"
+  code <<-EOH
+    cp /usr/hdp/2.2.0.0-2041/hbase/lib/hbase-common-0.98.4.2.2.0.0-2041-hadoop2.jar /tmp/oozie-hbase-prep/WEB-INF/lib/hbase-common-0.98.4.2.2.0.0-2041-hadoop2.jar
```
mlongob: Can you please not use explicit version numbers in the recipe? That goes for both the hdp version and the hbase jar versions. This will break the second we merge the 2.3 changes

bijugs: @mlongob . This is one-off for HDP 2.2. We will need to modify this recipe for HDP 2.3 since we have ``chef-bach`` changes and HDP related changes including this path.

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
```diff
@@ -0,0 +1,100 @@
...
+# Recipe to include hbase jar files into Oozie
+#
+%w{
+hbase
+}.each do |p|
+  package p do
+    action :install
+  end
+end
+
```
pu239ppy: there is only one package to install why the each construct?

pu239ppy:
One more thing, we already have facilities to rebuild war in oozie.rb.  All you have to do is symlink the files into libext directory. See
https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/oozie.rb#L53-L73

pu239ppy:
Some testing notes we may need to adjust heap size with
````export HBASE_HEAPSIZE=256````

File: cookbooks/bcpc-hadoop/recipes/oozie_pack.rb (hunk 2 above)
Comment (pu239ppy) [inline cookbooks/bcpc-hadoop/recipes/oozie_pack.rb]: It should not be too difficult just interpolate the value
````ruby
/usr/hdp/#{node.default['hdp_ver']}/hbase/lib/hbase-common-0.98.4.#{node.default['hdp_ver']}-hadoop2.jar /tmp/oozie-hbase-prep/WEB-INF/lib/hbase-common-0.98.4.#{node.default['hdp_ver']}-hadoop2.jar
````
This will hopefully ensure that it does the right thing on 2.3.  You can use the same thing for destination directory, in fact define it in an attribute.

File: cookbooks/bcpc-hadoop/recipes/oozie_pack.rb (hunk 3 above)
Comment (bijugs) [inline cookbooks/bcpc-hadoop/recipes/oozie_pack.rb]: @pu239ppy . This can be changed to not use ``.each`` construct. It was initially coded to look at ``hbase-client`` ``hbase-master`` components separately.

**Comment by bijugs** [comment]
@pu239ppy . Thanks for sharing the Oozie credential section to use to run a hbase action through Oozie once the jars are included into Oozie as per this PR. Regarding using (setting Oozie set-up.sh)[https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/oozie.rb#L53-L73], the script is used to set-up for a new install and there may be some assumptions made in the script which comes out of HDP distro and it may change. It would be better if the HBase jar inclusion into the Oozie to be not dependent on these assumptions if any.

**Comment by pu239ppy** [comment]
The method in oozie.rb is of our own production, granted it may use something enabled by HDP, but every time we wanted to add something to oozie war it might be easier to symlink a file rather then remake the whole war.  It is also more generic and other products may be able to take advantage of this process if they require inclusion into the war file.  That is unless we suspect this will undergo a very drastic change.

File: cookbooks/bcpc-hadoop/recipes/oozie_pack.rb [hunk 4]
```diff
@@ -0,0 +1,96 @@
+#
+# Recipe to include hbase jar files into Oozie
+#
+package "hbase" do
+  action :upgrade
+end
+
+bash "/tmp/oozie.war" do
```
cbaenziger: Please do not use `/tmp` but the chef cache instead.

cookbooks/bcpc-hadoop/recipes/oozie_pack.rb
```diff
@@ -0,0 +1,96 @@
...
+  user "root"
+  group "root"
+  code <<-EOH
+    cp /usr/hdp/2.2.0.0-2041/oozie/oozie-server/webapps/oozie.war /tmp/oozie.war
```
cbaenziger: This needs to be updated to active release

bijugs:
Handled in PR #681.

PR #385, opened 2016-01-21
Use a restricted plugin set with ohai on bootstrap

http-418:
By only loading the plugins needed in the configuration file, we can avoid having ohai run out of memory while parsing knife.rb on low-memory bootstrap nodes.


http-418:
This ended up rolled up in https://github.com/bloomberg/chef-bach/pull/636 because I was tired of waiting for literally minutes to perform basic chef actions.

Issue #384, opened 2016-01-21
Do not zap disks during cluster-assign-roles.sh run


kiiranh:
During `cluster-assign-roles.sh` run, we [zap disks](https://github.com/bloomberg/chef-bach/blob/7c11f0bd92a71a040a19993e6f5c4aa15b0381be/chefit.sh#L39). This is OK for a new node setup. But when a existing node is re-installed, this script destroys the partition table and corrupts the disk (e.g.: if using XFS) thus making the existing data unusable. This script is not relevant to chef-bach (where we would like to preserve the data on disk) and should be removed. Any disk related operations should be handled in the `preseed` file or a separate cookbook/file.

kiiranh:
I guess this can be closed?

cbaenziger:
For future reference, a nice find of probably what was actually the root cause of #149 but not found at that time.

PR #386, opened 2016-01-22
Don't format disks before chef install

http-418:
This fixes issue #384.

I believe the zap-disks script is what causes disks to require an `xfsrepair` after re-installing hosts. 

`xfsrepair` is always a destructive operation, because XFS only requires repair after serious corruption, e.g. a missing superblock.  Routine problems, of the kind fixed by fsck, are handled by mount(2)


kiiranh:
LGTM

amithkanand:
LGTM.

PR #288, opened 2015-09-29
added :data key to default[:bcpc][:hadoop][:kerberos] 

pu239ppy:
moved all kerberos principal information into :data key to fix error thrown by https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/configs.rb#L89

PR #390, opened 2016-01-28
Fix for Initial HDFS maintenance recipes attributes

cbaenziger:
This uses the correct kerberos attribute which changed in PR #288.


cbaenziger:
Tested happily

kiiranh:
Thanks for the prompt fix @cbaenziger

PR #391, opened 2016-02-01
Set lower default Zabbix trigger severities

kiiranh:
Setting the default severities for pre-defined Zabbix triggers to lower values. To be set to higher values if and when required by overriding in the environment.


cbaenziger:
This was caused by a misunderstanding. It seems your original severity definitions are fine.

PR #392, opened 2016-02-02
Restart services when core-site.xml changes

amithkanand:
This PR takes care of restarting services when `core-site.xml` and `hdfs-site.xml` changes. The impacted services are:
* `NameNode`
* `Oozie`
* `HBase`
* `Hive`


http-418:
I'm taking a look.

http-418:
This is definitely working.

PR #398, opened 2016-02-11
control carbon-cache max updates and prevent it from over utilizing disk

pu239ppy:
When MAX_UPDATES_PER_SECOND is too high carbon-cache can deny disk access to other processes such as ZooKeeper. When Zookeeper becomes unstable, so does the entire cluster, Name Node and Resource Manager and HBase and others all depend on being able to reach ZooKeeper. On many clusters we've noticed that carbon-cache is driving root disk utilization to 100%, we need to tell carbon cache to keep more in memory and trickle data to disk slower.


kiiranh:
LGTM

PR #400, opened 2016-02-11
Change to use 0.9.0 version of ark cookbook

bijugs:
Latest version of ``ark`` cookbook (1.0.0) uses ``seven_zip`` instead of ``7-sip``. But ``maven`` cookbook requires ``ark`` 0.9.0. This change is to pin ``ark`` version to 0.9.0 so that these components are compatible.


http-418:
Testing now

PR #401, opened 2016-02-12
Make group owner of the hdfs group directory

amithkanand:
If an `HDFS` group directory is owned by hdfs then group account can't create `hdfs snapshots` on it. To make it work the directory must be owned by `group`.


cbaenziger:
I don't think this will quite work. There's no [requirement](https://github.com/bloomberg/chef-bach/blob/49e0434e6a98f2e7a48d745bdc35a4b68e1f4c9b/cookbooks/bcpc-hadoop/recipes/hdfs_group_directories.rb#L36) `#{group}` is also a user.

PR #402, opened 2016-02-16
Derive 'node_number' field from MAC addresses

http-418:
The node_number was previously derived from a node's management IP.
In a routed rack configuration, this could cause the node_number to
wrap around and overlap inside a single cluster.

This commit derives the node_number from the MAC address on the
management interface.  The MAC is converted to an integer.  The node
number is this integer modulo Integer.MAX_VALUE.

This gives us a ~31 bit positive integer that is guaranteed to be
unique inside any single-vendor customer.  It is very likely to be
unique across any two NICs, but not certainly so, since only the
last 7 bits of a 24 bit OUI are present.

Zookeeper and keepalived take use (node_number % 8) to derive an 8 bit
value from the 31 bit node number.  In a sufficiently large cluster
this may overlap.


http-418:
This should not be a breaking change, since the current node_number in code has precedence "default" while the old node_number implementation used "set".

Existing clusters will have an already-existing node number in the node definition.

pu239ppy:
- The numbering scheme is also used to assign IDs to resource manager hosts in /etc/hadoop/conf/yarn-site.xml and /etc/hadoop/conf/hdfs-site.xml, for uniformity's sake you may want to suggest we make a change there as well

- How do you see this number wrapping around?
- If we still agree to go ahead with this, perhaps to ease the transition a switch can be used to control how the node number is generated, this way a massive reconfiguration should not be a problem.  However it would be interesting to see if this can be handled gracefully.  I suspect neither hdfs nor yarn would be problematic.

cbaenziger:
BREAKING: This may require some coordination since the node numbers will change in an existing cluster. The following technologies use this:
* Zookeeper (should not matter if different across different nodes)
* YARN RM (only the node number is in the Zookeeper znode for the active RM)
* HDFS Namenode (the node number and hostname is in the Zookeeper znode for the active NN)

One will need to be cautious about rolling out the change. I suspect the following will be needed:
# Start with all heads except the active RM and NN (ensure at least one in-active RM and NN)
# Then, Chef all clients to provide the mapping for the recently updated nodes
# Next, ensure the previously chef'd in-active RM and NN are made active
# Chef the previously-active RM and NN's
# Then, re-Chef all clients to ensure consistent mappings

http-418:
I have reduced the precedence on the "new" node number attribute to default.

Existing clusters all have node numbers hard-coded in the node object with "node.set".  If we want to migrate those clusters, we can simply remove the override from the node object.

New clusters and hosts who have had their node objects cleared will derive node numbers and ZK IDs from the MAC address.

cookbooks/bcpc/attributes/default.rb
```diff
@@ -24,6 +24,37 @@
...
+#
+# On the provisioning node and during early bootstrap, we won't have
+# any of these values, in which case we just don't set the
+# node_number.
+#
+default_interface = begin
+                      interface_name = node[:network][:default_interface]
+                      node[:network][:interfaces][interface_name]
+                    rescue
+                      nil
```
cbaenziger: Should we not `raise`? Originally `chef-bcpc` would try its darnedest to never fail -- this would result in swiss cheese clusters which did not have key attributes set and debugging why a cluster was not stable or did not come up at all was quite difficult.

http-418: Yeah.  We should raise here.  The original code referenced attributes that could actually be missing from upstream, so nil was consistent behavior.

Now that those are un-referenced, this should _never_ fail.

cbaenziger:
For reference the VRRP Virtual Router ID is 8 bit per section 5 of [RFC 3768](https://www.ietf.org/rfc/rfc3768.txt) can you also truncate the keepalived router ID to 8 bits?

http-418:
Clay, the router ID is generated with "rand(1000)," it doesn't use the node_number.

The VRRP router priorities are determined by node_number, but they are already mod 255.

cbaenziger:
Wow; that `rand(1000)` craziness needs to stop; please use the 8-bit node_number. Also, please `raise` if you find another node has the same 8-bit node_number. This new 8-bit node_number will be more opaque to an operator should it collide now than when it was an IP address last octet or offset.

cbaenziger:
Andrew are you planning to squash commits into node_number changes and syntax fix-ups or did you want to leave this in its current incarnation?

If you do plan to leave as is, I am confused about this comment:
````
Zookeeper and keepalived take use (node_number % 8) to derive an 8 bit
value from the 31 bit node number.  In a sufficiently large cluster
this may overlap.
````

**Comment by cbaenziger** [comment]
In testing, this seems to work so far for adding to an existing cluster. I ran the following for validation:
* before running new code
  * `for i in 1 2 3; do EDITOR=cat knife node edit -a bcpc-vm${i}.bcpc.example.com > /tmp/pre-bcpc-vm${i}.bcpc.example.com.json; done`
* updated and ran new code, via
  * `git pull <new code URL>`
  * `knife cookbook upload -a`
  * `./cluster-assign-roles.sh Test-Laptop hadoop`
* then re-dumped the node objects with `for i in 1 2 3; do EDITOR=cat knife node edit -a bcpc-vm${i}.bcpc.example.com > /tmp/pre-bcpc-vm${i}.bcpc.example.com.json; done`
* finally differed the pre- and post- node objects, ensuring `node[:bcpc][:node_number]` was the same

File: cookbooks/bcpc/attributes/default.rb [hunk 2]
```diff
@@ -3,101 +3,163 @@
...
+# files, keepalived node priorities, and other values.  It must be
+# unique within a cluster.
+#
+# The node number is generated from the integer value of the
+# default interface mac_address, modulo Java's Integer.MAX_VALUE.
+#
+# For 8-bit node numbers, (e.g. zookeeper, keepalived) there is a
+# helper method defined in libraries/node_numbers.rb
+#
+unless node['bcpc']['node_number']
```
cbaenziger: On general principal, doing this in an attribute file and not in a function is pretty concerning to me for maintainability -- it is quite a bit akin to a side-affect not actual intention from my viewpoint.

cbaenziger:
To test this with nodes that did not have a `node_number`, I cleared the attribute via a `knife node edit <fqdn>` command. However, this was not repopulated via a `chef-client -o 'role[Basic]'` while a normal Chef run of any of the head node roles would fail since there was a node object without a node number.

While running `chef-shell -z` I could `puts node[:bcpc][:node_number]` and see it was being set correctly but since I was running chef-client with the `-o` option the node object was not being written back to the server. (When running a full Chef run to failure it did update the node with the `node_number`.)

cbaenziger:
For reference, Andrew has asserted he was able to build a test cluster from scratch with this addressing my concerns, so I verbally gave my accent to a self-merge for this PR.

PR #403, opened 2016-02-18
Install Unix debuggin tools

bijugs:
Changes to install ``traceroute`` and ``iotop`` on all the nodes which will help in debugging issues.


kiiranh:
LGTM

PR #404, opened 2016-02-19
maven works again

pu239ppy:
in #400 we gave pinned ark to 0.9.0 to a specific version that does not require seven_zip in order to satify maven requirements, but it looks like maven has now caught up and requires  a maven ~> 1, which in turn wants seven_zip.  I am also keeping 7-zip on purpose in case we miss any requirements.


cbaenziger:
BREAKING: One needs to manually update the cookbooks on a cluster to include `seven_zip` and a current `ark`.

PR #405, opened 2016-02-24
Performance improvements for metrics and alarming

kiiranh:
This PR proposes changes to improve the performance of the monitoring stack, specifically the graphite-to-zabbix functionality. Details are listed below:

##### Mysql:
- Some tweaks for better I/O performance since this is an important part of the stack. Hannibal will use mysql as a backing store and mysql performance will be a factor when we actively use it
- Allocate 2% of the system memory to innodb buffer pool

##### Zabbix:
- Enable password-less login for the Zabbix “guest” user
- Provide read access to the <chef.environment> HostGroup to the “Guests” user group, so anyone can look at the cluster status, view events and latest data for the items defined for the HostGroup
- Reduce load by setting the default time to store history to 1 day and trends to 7 days
- Change the frequency of Zabbix housekeeper from 1 hour to 4 hours
- Increase cache sizes

##### Diamond:
- Disable the collectors that are not used right now
- Update the polling interval from 10 sec to 15 sec to match with the highest resolution for graphite metric retention

##### Graphite-to-Zabbix:
- Reduce the number of queries made to graphite to fetch metric data. Before, we made a http query per item to graphite. Now, by default we glob most of the queries so data is fetched for multiple items per host or multiple hosts in a single query.
- Update the query_graphite.py python script to log time taken per query, provide multithreading support, and logic to unmarshall the data from the graphite glob query.
- Add queries to monitor chef-client failure, HBase RegionServer GC time and Zookeeper quorum

##### BugFixes:
- Remove “Safemode” from the list of JMX metrics collected from the Namenode because when the namenode is not in safe mode the value for this is “” and the resulting graphite metric line is: “namenode  timestamp”. Carbon-relay is unable to parse the ‘value’ from it and complains in the log.
- Zabbix was updated to 2.4.5 under PR #349. However, we missed the API change in zabbix’s action.[create/update] method and hence actions were not created properly. The zabbix action.[create/update] method has been updated in bcpc-hadoop::graphite_to_zabbix.rb.
- bcpc-hadoop::utils::get_head_nodes had incorrect search string for finding head nodes
- Fix for multiple redundant entries ("source /etc/network/interfaces.d/*.cfg") in /etc/network/interfaces because of which networking fails to start(after reboot) as it runs out of file descriptors when the count of these lines is greater than the number of descriptors (credits: cbaenziger)

##### Breaking:
- The position on `bcpc::diamond` has changed in role `BCPC-Hadoop-Head`; on existing cluster on will have to run `cluster-assign-roles.sh` to update the run_list on the nodes.
- The retention period has changed for chef graphite metrics. On existing clusters, one will have to resize the whisper database for chef* or delete the databases and let chef create new ones with the updated retentions
- Zabbix: before deploying the new code, on head nodes
    - Stop chef-client
    - Stop the graphite-to-zabbix script (sudo -u zabbix crontab -r) 
    - On Zabbix UI with admin login, delete all the hosts and actions (will be recreated with new configuration)
    - Run chef-client to update zabbix configuration and create items on UI and have the new query_graphite.py script deployed

##### Future work:
- The current structure to define graphite queries and zabbix triggers is static. For every host, you specify an item and the trigger to be created for that item. There is no way to specify multiple triggers for a single item. Multiple triggers on single item help in alarming for different states of the same item. A good way to address this would be to use Zabbix templates to specify the Item, Trigger, Action configuration and either use Zabbix LLD to discover and auto-create the items or in the Ruby code create hosts and attach the relevant templates. This way, one can make a change to the Zabbix template XML which is more flexible it specifying relationships among items, trigger, actions and other templates. Templates can be created for different roles and/or applications and attached to a host depending on its role.
- Possibly replace the default carbon-relay with [carbon-relay-ng](https://github.com/graphite-ng/carbon-relay-ng) or [carbon-c-relay](https://github.com/grobian/carbon-c-relay) 
  - We could also use the relay to replace the 'pull metric from graphite and send to zabbix' model to forward selected metrics to Zabbix directly. For this, one can write rules in the relay to reformat(and/or aggregate) the metrics of interest in the format that zabbix expects and forward those to Zabbix directly (or through another custom relay which just sends to Zabbix)


pu239ppy:
Go ahead and squash

kiiranh:
@pu239ppy Squashed the commits; every commit is now an independent one.

cbaenziger:
Could the Zabbix bits please all live in BCPC and not split across BCPC-Hadoop and BCPC?

kiiranh:
@cbaenziger I agree the zabbix and graphite bits should all be under bcpc cookbook. However, making that change under this already huge PR will take more time for coding and testing. I created Issue #410 to address this. I've also squashed the commits to further reduce their count.

PR #412, opened 2016-03-04
Do not sync time between bootstrap and cluster nodes

amithkanand:
This PR fixes the issue where `chefit.sh` script takes time from bootstrap node and sets it on the cluster nodes. This has caused issues when bootstrap node doesn't have correct time.


kiiranh:
LGTM

Issue #372, opened 2016-01-11
hive-site.xml on worker nodes


pu239ppy:
hive-site.xml on worker nodes should not expose metastore password information

Issue #325, opened 2015-11-04
chef code keep creating data base 'hive_table_stats' even it's already existed


DrRayWang:
Mysql log : 

```
...
2015-11-04 12:14:46 27938 [ERROR] Slave SQL: Error 'Can't create database 'hive_table_stats'; database exists' on query. Default database: ''. Query: 'CREATE DATABASE hive_table_stats', Error_code: 1007
...
```

this error keeps appearing in the log.

Clay said :
As to the hive_table_stats:

```
* ruby_block[hive_table_stats_db] action runWarning: Using a password on the command line interface can be insecure.
ERROR 1054 (42S22) at line 1: Unknown column 'hive_table_stats' in 'where clause'
Warning: Using a password on the command line interface can be insecure.
ERROR 1007 (HY000) at line 1: Can't create database 'hive_table_stats'; database exists
```

13:59:26 That's what causes that error; it's a bug but a relatively benign one

PR #414, opened 2016-03-04
Hive configuration cleanup

pu239ppy:
**Changes**
- Templates now get variables explicitly
- If the host being chefed is not a hive server mysql passwords are blank addresses #372 
- addresses database creation issue #325 where chef will create hive_table_stats DB even if it already exists

**Testing**
- Standard wiki test non-kerberized
````
insert into table t1 values(1, "aaa111");
insert into table t1 values(2, "bbb222");
insert into table t1 values(3, "ccc333");

hive> select * from t1;
OK
1       aaa111
2       bbb222
3       ccc333
Time taken: 0.142 seconds, Fetched: 3 row(s)
````
- Kerberized
passed as well with a different table
````
Time taken: 19.27 seconds
hive> select * from t2;
OK
3       ccc333
2       bbb222
1       aaa111
Time taken: 0.123 seconds, Fetched: 3 row(s)
````

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/hive_config.rb
+++ b/cookbooks/bcpc-hadoop/recipes/hive_config.rb
@@ -6,11 +6,12 @@
 
 
 #Create hive password
-make_config('mysql-hive-password', secure_password)
+hive_password = make_config('mysql-hive-password', secure_password)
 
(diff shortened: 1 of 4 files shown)


ekund:
LGTM

PR #416, opened 2016-03-07
ZooKeeper configuration changes

bijugs:
Changes  in this PR includes
- Change ``export ZOO_LOG_DIR`` to use node attribute so that it is configurable. Currently it is hardcoded.
- Add ``dataLogDir`` ZooKeeper config and ideally this need to point to a dedicated disk for performance.


kiiranh:
LGTM

PR #418, opened 2016-03-07
Change path for sqoop options to be hdp current.

cbaenziger:
Chris Martin made this change and it has worked well.
* Change path for sqoop options to be hdp current.
* Using env variables used and giving defaults if they are not defined.


cbaenziger:
Apparently this got [merged](https://github.com/bloomberg/chef-bach/blob/aa6b333717cb203c0ced7ae3c867372a955ef701/cookbooks/bcpc-hadoop/templates/default/sq_sqoop-env.sh.erb) in, in some form, as part of HDP 2.3.4.

Issue #419, opened 2016-03-07
HBase processes running with multiple jaas files


bijugs:
Since `HBASE_OPTS`, `HBASE_MASTER_OPTS` and `HBASE_REGIONSERVER_OPTS` are set with different `jaas` files in [hbase-env.sh](https://github.com/bloomberg/chef-bach/blob/a4dc8a967ec988d170a1f833881567d218cea191/cookbooks/bcpc-hadoop/templates/default/hb_hbase-env.sh.erb#L117-L119), the HBase master process and region process are running with multiple `jaas` files. The following is from `ps` output for `hbase-master`.

```
*-Djava.security.auth.login.config=/etc/hbase/conf/hbase-client.jaas*
*-Djava.security.auth.login.config=/etc/hbase/conf/hbase-server.jaas*
```

We should use `hbase-server.jaas` for all server processes.

PR #421, opened 2016-03-08
Changes to use one jaas file template for HBase server process

bijugs:
This is to fix issue #419 so that we use only one ``jaas`` file for all the base server processes.


pu239ppy:
Looks as if the master and region server were identical.  Have you tested this just in case?  Also it needs to be rebased

bijugs:
Outdated. Closing.

PR #422, opened 2016-03-08
Avoid openjdk installation and use Oracle Java

amithkanand:
Since we have started using Oracle Java, this PR removes installation of OpenJDK.


bijugs:
Any reason we want to avoid OpenJDK. It will be better to have an option to choose between JDK flavors so that users can choose. One compelling reason for OpenJDK support is the feature additions and frequent bug fixes along with the compelling reason of being OpenSource.

amithkanand:
We are removing it from our script. The upstream cookbook that we use does have an option to install OpenJDK and we need to set attributes accordingly. I don't see any need to have two JDK's installed simultaneously on the machine.

pu239ppy:
As long as we have an option to easily switch between the two if a need arises.

cbaenziger:
Sorry to rain on the parade, but did any one see if setting `node[:bcpc][:binary_server_url]` will still work for this? I'm a bit concerned about races calling `get_binary_server_url` in an attribute file.

PR #423, opened 2016-03-08
prevent hive startup scripts form redefining HIVE_SERVER2_OPTS and HIVE_METASTORE_OPTS

pu239ppy:
Modified startup scripts


amithkanand:
Why are we removing this what if some values are defined in `hive-env.sh`?

pu239ppy:
Only making sure that those values are used in startup as opposed to being
redefined

VAL="$VAL and more" vs VAL="and more"

amithkanand:
Ah..my bad. I misread the change. This looks good to me.

PR #434, opened 2016-03-14
Make Yarn application logs visiable through RM UI after app completes

amithkanand:
As of today, after a `Yarn` application is FINISHED successfully, the logs for completed application are not visible via RM UI. One would see error akin to
````
Failed while trying to construct the redirect URL to the log server. Log server url may not be configured
````
This PR adds the correct parameter to make `Yarn` application log visible

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
+++ b/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
@@ -54,6 +54,9 @@ bash "kill yarn-resourcemanager" do
   returns [0, 1]
 end
 
+hdfs_write = "echo 'test' | hdfs dfs -copyFromLocal - /user/hdfs/chef-mapred-test"
+hdfs_remove = "hdfs dfs -rm -skipTrash /user/hdfs/chef-mapred-test"
+
 bash "setup-mapreduce-app" do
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb [hunk 1]
```diff
@@ -303,4 +303,10 @@
     <name>yarn.scheduler.fair.preemption</name>
     <value><%= node["bcpc"]["hadoop"]["yarn"]["scheduler"]["fair"]["preemption"] %></value>
   </property>
+
+  <property>
+    <name>yarn.log.server.url</name>
+    <value>http://<%="#{float_host(@hs_hosts.map{|i| i[:hostname] }.sort.first)}:19888" if not @hs_hosts.empty?%>/jobhistory/logs</value>
```
bijugs: For some reason ``@hs_hosts`` passed is empty, what will be the impact?

pu239ppy: Agree with @bijugs better off just making this a conditional.  Or don;t ask, if this is empty you have a whole new set of problems.

PR #436, opened 2016-03-17
Remove hadoop metrics 1 properties

bijugs:
Changes to remove hadoop metrics 1 properties from the code base and use metrics 2 properties.


pu239ppy:
In the PR description, state the purpose of this.

PR #438, opened 2016-03-17
Avoid creating keytabs for Bootstrap host

amithkanand:
This PR adds logic to avoid generation of keytabs for bootstrap host.


amithkanand:
This is not needed anymore.

PR #440, opened 2016-03-18
Move apt after ubuntu recipe in role::Basic

mkoni:
This issue happens when try to kerberize chef-bach cluster. 
re-chefing bootstrap node after adding
"recipe[bach_krb5::krb5_server]", "recipe[bach_krb5::keytabs]"
 to BCPC-Bootstrap.json
failed on
sudo apt-get -q -y  install libkrb5-dev=1.10+dfsg~beta1-2ubuntu0.7

It seems there was problem with apt-get package authentication because apt recipe ran before ubuntu recipe, but really apt depends on ubuntu recipe, and hence ubuntu should run first

the proposed change is to run ubuntu recipe before apt recipe.

### DIFF
--- a/roles/Basic.json
+++ b/roles/Basic.json
@@ -6,8 +6,8 @@
     "run_list": [
       "recipe[bach_common::motd]",
       "recipe[bcpc::graphite_handler]",
-      "recipe[apt]",
       "recipe[ubuntu]",
+      "recipe[apt]",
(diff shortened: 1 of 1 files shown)


cbaenziger:
Mo, is this needed in the end?

mkoni:
Clay, I am closing this PR since the change did not help.

PR #327, opened 2015-11-09
HDP2.3 Support

cbaenziger:
This PR supports Hortonworks Data Platform 2.3

PR #444, opened 2016-03-23
Create hbase staging dir required for SecureBulkLoadEndPoint

amithkanand:
This PR adds creation of `hbase` staging directory that is required for the coprocessor `org.apache.hadoop.hbase.security.access.SecureBulkLoadEndpoint` in a `Kerberized` cluster. RegionServers do not start and throw an error akin to given below in case the directory is missing.
````
2016-03-22 20:11:15,458 INFO org.apache.hadoop.hbase.regionserver.HRegionServer: STOPPED: The coprocessor org.apache.hadoop.hbase.security.access.SecureBulkLoadEndpoint threw java.lang.IllegalStateException: Failed to get FileSystem instance
````

I have already tested this in my test environment. Also, this was added as a comment on PR #327


amithkanand:
Discussed with @cbaenziger and he is going to create a PR for this.

PR #445, opened 2016-03-23
Fix for HBase staging dir

cbaenziger:
This is omitted code from the HDP2.3 (PR #327) which supports HBase bulk loads by setting an explicit staging directory and making it and the HBase root directory attributes.


amithkanand:
LGTM.

cbaenziger:
This fixes the following issue reported in the HBase master log:
````
2016-03-24 16:45:16,942 DEBUG org.apache.hadoop.hbase.util.FSUtils: Unable to create version file at hdfs://Test-Laptop/hbase, retrying org.apache.hadoop.security.AccessControlException: Permission denied: user=hbase, access=WRITE, inode="/hbase/.tmp/hbase.version":hdfs:supergroup:drwxr-xr-x
        at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.check(FSPermissionChecker.java:319)
        at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.check(FSPermissionChecker.java:292)
[...]
````

PR #446, opened 2016-03-24
fortan lib support for https://github.com/fommil/netlib-java

ekund:
Added a new recipe to bcpc-hadoop cookbook that installs the two necessary packages.  This will allow worker nodes to have libblas and liblapack installed on them.  This is for support of https://github.com/fommil/netlib-java.

Testing:
```
ubuntu@bcpc-vm3:~$ ls /etc/alternatives/libbla*
/etc/alternatives/libblas.so.3gf
ubuntu@bcpc-vm3:~$ ls /etc/alternatives/liblapack*
/etc/alternatives/liblapack.so.3gf
```

### DIFF
--- a/cookbooks/bach_spark/recipes/default.rb
+++ b/cookbooks/bach_spark/recipes/default.rb
@@ -1,11 +1,11 @@
-node.default[:bcpc][:hadoop][:yarn][:aux_services][:spark_shuffle][:class] = "org.apache.spark.network.yarn.YarnShuffleService"
+node.default[:bcpc][:hadoop][:yarn][:aux_services][:spark_shuffle][:class] =
+  'org.apache.spark.network.yarn.YarnShuffleService'
 
 spark_pkg_version = node[:spark][:package][:version]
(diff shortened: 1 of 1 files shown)


cbaenziger:
Since this is to support [Spark MLlib](https://spark.apache.org/docs/1.1.0/mllib-guide.html#dependencies) I would prefer to see the Spark recipe pull this in rather than adding another step to roles. (Roles are a pain to merge in existing clusters and I do not see a reason to not use this.)

cbaenziger:
Please try to follow [`rubocop`](https://github.com/cbaenziger/chef-bach#user-content-contributing) advice as well.

ekund:
I was thinking that if this is a generic library, we may not want it added under spark, as it will be available for other components as well.  It isn't actually directly related to spark at all.

ekund:
Ok, changes made as per comments by @cbaenziger .  Also made some cosmetic changes due to complains from [`rubocop`](https://github.com/cbaenziger/chef-bach#user-content-contributing).

PR #448, opened 2016-03-28
Disable yarn's nodemanager virtual memory check

amithkanand:
This PR disables `Yarn NodeManager` virtual memory check.


ekund:
What is the purpose of doing this?

cbaenziger:
This prevents YARN containers from being killed when they request more that twice the physical memory requested.

PR #452, opened 2016-03-29
Install spark with Hadoop jars

amithkanand:
This PR makes sure that the `Spark` is installed with `Hadoop` jars bundled in `Spark assembly` jar. This is required for Spark to work with `Hive`. Without `Hadoop/Hive` jars `Spark` is not able to create a `HiveContext`.


amithkanand:
This was tested on my VM cluster.

PR #455, opened 2016-03-30
Provide GitHub workflow for README Contributing section

cbaenziger:
This adds a pointer to our current GitHub workflow (with color matched labels) as stored in our `pages` branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,13 +72,22 @@ also manually create the databag & item (as per the example in
 ``data_bags/configs/Example.json``) and manually upload it if you'd rather not
 bother with the whole ``admin`` thing for the first run.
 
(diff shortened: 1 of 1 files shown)


cbaenziger:
Let me drop in the deployment types we support while I'm dorking in the README. Just a few minutes for that...

ekund:
Can you squash the commits?

cbaenziger:
Ah I thought the commits were pretty reasonable? What'd you want to group all three under?

ekund:
I'd group them under "Improved Documentation of SDLC".  I thought in general we try to keep 1 commit per PR.

cbaenziger:
Will squash under that banner. (One commit per feature is the desire. That way we can tell what went in where most easily. No need for a work log, but don't want to hide things in commits either.)

Issue #457, opened 2016-03-30
HBase RegionServer fails to start in secure mode


amithkanand:
After kerberizing the cluster the `HBase` regionserver fails to start and throws following error:

```
2016-03-30 08:05:42,630 ERROR org.apache.hadoop.hbase.coprocessor.CoprocessorHost: The coprocessor org.apache.hadoop.hbase.security.access.SecureBulkLoadEndpoint threw java.lang.IllegalStateException: Failed to get FileSystem instance
java.lang.IllegalStateException: Failed to get FileSystem instance
        at org.apache.hadoop.hbase.security.access.SecureBulkLoadEndpoint.start(SecureBulkLoadEndpoint.java:153)
        at org.apache.hadoop.hbase.coprocessor.CoprocessorHost$Environment.startup(CoprocessorHost.java:411)
        at org.apache.hadoop.hbase.coprocessor.CoprocessorHost.loadInstance(CoprocessorHost.java:253)
        at org.apache.hadoop.hbase.coprocessor.CoprocessorHost.loadSystemCoprocessors(CoprocessorHost.java:156)
        at org.apache.hadoop.hbase.regionserver.RegionCoprocessorHost.<init>(RegionCoprocessorHost.java:231)
        at org.apache.hadoop.hbase.regionserver.HRegion.<init>(HRegion.java:723)
        at org.apache.hadoop.hbase.regionserver.HRegion.<init>(HRegion.java:631)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
        at org.apache.hadoop.hbase.regionserver.HRegion.newHRegion(HRegion.java:5893)
        at org.apache.hadoop.hbase.regionserver.HRegion.openHRegion(HRegion.java:6203)
        at org.apache.hadoop.hbase.regionserver.HRegion.openHRegion(HRegion.java:6175)
        at org.apache.hadoop.hbase.regionserver.HRegion.openHRegion(HRegion.java:6131)
        at org.apache.hadoop.hbase.regionserver.HRegion.openHRegion(HRegion.java:6082)
[...]

PR #458, opened 2016-03-30
Set correct staging directory for HBase secure bulk load

amithkanand:
This PR sets correct directory location for `HBase` secure bulk load and fixes issue #457


ekund:
LGTM

PR #459, opened 2016-03-31
Pin ntp cookbook version to 1.10.1

http-418:
This PR pins the NTP cookbook to 1.10.1 in order to avoid a dependency
on 'windows >= 1.38.0'.   We have a pre-existing pin to an older
windows cookbook, which I have no desire to re-examine.


amithkanand:
LGTM.

PR #465, opened 2016-04-01
Setup hadoop yarn log directory with correct permissions

amithkanand:
This PR removes creation of `yarn log` directory from `NameNode` recipe that was creating and setting the permissions incorrectly causing `oozie` `mapreduce` action to fail on job's completion while moving it to history server. Since `yarn log` is related to `Yarn/Mapreduce`, it should be handled in `resource_manager` recipe only.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
+++ b/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
@@ -29,7 +29,7 @@ end
 end
 
 bash "create-hdfs-yarn-log" do
-  code "hdfs dfs -mkdir -p /var/log/hadoop-yarn && hdfs dfs -chmod 1777 /var/log/hadoop-yarn && hdfs dfs -chown yarn:mapred /var/log/hadoosp-yarn"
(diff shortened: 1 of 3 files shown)


cookbooks/bcpc-hadoop/recipes/resource_manager.rb
```diff
@@ -29,7 +29,7 @@
 end
 
 bash "create-hdfs-yarn-log" do
-  code "hdfs dfs -mkdir -p /var/log/hadoop-yarn && hdfs dfs -chmod 1777 /var/log/hadoop-yarn && hdfs dfs -chown yarn:mapred /var/log/hadoosp-yarn"
+  code "hdfs dfs -mkdir -p /var/log/hadoop-yarn && hdfs dfs -chmod 0777 /var/log/hadoop-yarn && hdfs dfs -chown yarn:mapred /var/log/hadoosp-yarn"
```
pu239ppy: Perhaps the problem is with hadoosp as well as its permissions

PR #474, opened 2016-04-07
Adding yarn.scheduler.maximum-allocation-mb to yarn-site.xml

hkothari:


pu239ppy:
Looks good, but lets use max host memory for this same as <name>yarn.nodemanager.resource.memory-mb</name>
https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/templates/default/hdp_yarn-site.xml.erb#L231

bijugs:
@hkothari . is this change tested? If so could you please label it as tested.

hkothari:
I haven't tested it. Lev mentioned he was going to trigger a test for me
since I'm not familiar with the test infrastructure you guys have set up.
Em sex, 15 de abr de 2016 às 11:16, Biju Nair <notifications@github.com>
escreveu:

bijugs:
The logic used in the memory calculation matches the one used to set the value for ``yarn.nodemanager.resource.memory-mb`` and I am assuming the intention is to be able to build a cluster on which containers can take up the all the allocated memory in NMs. But there are other cases where we would want the container sizes to be restricted i.e. the value for  ``yarn.scheduler.maximum-allocation-mb`` should be configurable.

hkothari:
Yeah, I'm with you. @pu239ppy any thoughts? I think you preferred that we just reused the other setting.

pu239ppy:
We already ran against this issue once where maximal allocation was left to default and yarn jobs were not created.  @bijugs what would be the case where we would not allow a yarn job to take up as much as it can?  Also lets keep in mind that we have other (scheduler) controls that are designed to protect against abuses.

bijugs:
In general users who wrap ``chef-bach`` should have the same flexibility what underlying components provide. By including logic this, we are taking away the flexibility and making assumptions for the users. If an user want to set the per container max value to the max available memory on the ``yarn`` node manager or something similar in other component like ``hdfs``, ``hbase`` etc, that can be done in the wrapper cookbook.

pu239ppy:
@bijugs, I agree with the separation of declarations, however currently the case is not that @hkothari aliased the meaning of one variable to use in two different places.  Same formula is being used twice, when we get around to fixing our templates and how values are passed into them, then we can ask for separate variables to be created as of now we do not have that.

PR #479, opened 2016-04-08
sample oozie test job

ekund:
Added a tools directory where we can use small smoke-test tools.

Added a sample oozie test case that tests an oozie shell action and pushes the resulting job id to hdfs for verification.

I envision more simple tools being added as time goes on.

### DIFF
--- a/tools/oozie/oozie_simple_test/README.md
+++ b/tools/oozie/oozie_simple_test/README.md
@@ -1,0 +1,27 @@
+# Oozie Simple Tester
+
+Use these files for a simple oozie test case that runs a simple oozie shell action.
+
+In order to run the test, simple update coordinator.properties with:
+ - The yarn task tracker url
+ - The HDFS name node url
+ - An HDFS directory to store the files
+ - A start and end time set to today
+
+Then, create your test HDFS directory, and create 2 subdirectories underneath - lib and output.
+
+Finally, put the following files into that HDFS directory.
+ - coordinator.xml  
+ - logoozieid.sh  
+ - test.sh  
+ - workflow.xml
+
+To run your test, run:
+` oozie job -oozie <oozie url> -config coordinator.properties  -run`
+
+This will output your oozie coordinator job id.
+
+To check on your job status:
+`oozie job -oozie <oozie url>  -info <jobid>`
+
+To check the job output from HDFS, retrieve the oozieId.log file from the HDFS output directory you created above.
--- a/tools/oozie/oozie_simple_test/setup.sh
+++ b/tools/oozie/oozie_simple_test/setup.sh
@@ -1,0 +1,3 @@
+sudo hdfs dfs -mkdir -p /tmp/oozie/shell-hadoopsh/
+sudo hdfs dfs -copyFromLocal job.properties /tmp/oozie/shell-hadoopsh/
+sudo hdfs dfs -copyFromLocal workflow.xml /tmp/oozie/shell-hadoopsh/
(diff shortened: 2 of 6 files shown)


tools/oozie/oozie_simple_test/workflow.xml
```diff
@@ -0,0 +1,51 @@
...
+      <shell xmlns="uri:oozie:shell-action:0.1">
+         <job-tracker>${jobTracker}</job-tracker>
+         <name-node>${nameNode}</name-node>
+         <configuration>
+            <property>
+               <name>mapred.job.queue.name</name>
+               <value>${queueName}</value>
+            </property>
+            <property>
+               <name>oozie.launcher.mapreduce.map.memory.mb</name>
```
pu239ppy: There is already a default in oozie-site.xml it mirros same settings as mapred-site.xml, but prepends oozie.launcher
https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/templates/default/ooz_oozie-site.xml.erb#L357

https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/templates/default/ooz_oozie-site.xml.erb#L367

http-418:
I really want to merge this, because it would be useful to me, but I don't understand whether Lev's comment should be addressed, first.

ekund:
@pu239ppy - the overrides provide a way to override default values, allowing to test if the defaults are set properly or not, as I've seen it cause issues.

saurabh313:
I think oozie coordinator jobs are more appropriate for recurrent actions. And since this just a test job, shouldn't we have oozie workflow  job ? In that case, we avoid the onus of updating start and end time each time we run the job.

ekund:
I have replaced the job with an even simpler shell action that is just a workflow and has no overrides.  I have also added a setup script to make running it easier.

amithkanand:

ekund:
I will, once we all agree that the code is good and ready to merge.

ekund:
Commits have been squashed.

tools/oozie/oozie_simple_test/setup.sh
```diff
@@ -0,0 +1,3 @@
+sudo hdfs dfs -mkdir -p /tmp/oozie/shell-hadoopsh/
+sudo hdfs dfs -copyFromLocal job.properties /tmp/oozie/shell-hadoopsh/
+sudo hdfs dfs -copyFromLocal workflow.xml /tmp/oozie/shell-hadoopsh/
```
cbaenziger: This `sudo` is incorrect; I originally read this as `sudo -u hdfs` which would work to upload data as the HDFS user on machines which are running HDFS daemons such as namenodes.

tools/oozie/oozie_simple_test/README.md
```diff
@@ -0,0 +1,27 @@
...
+ - A start and end time set to today
+
+Then, create your test HDFS directory, and create 2 subdirectories underneath - lib and output.
+
+Finally, put the following files into that HDFS directory.
+ - coordinator.xml  
+ - logoozieid.sh  
+ - test.sh  
+ - workflow.xml
+
```
cbaenziger: Mention `kinit`'ing as a user first and creating a kerberos identity for one (e.g. `ubuntu`?) There's a pointer to doing that in our Wiki under [`Add Kerberos principals by using kadmin.local on bootstrap node`](https://github.com/bloomberg/chef-bach/wiki/Creating-a-Kerberized-cluster).

cbaenziger:
Instead of using `/tmp` which will collide if multiple users try to run this, how about using `${wf:user()}` in the paths?

PR #482, opened 2016-04-08
Add support for Kafka 0.9.0.1 and remove 0.8.x

amithkanand:
This PR fixes the URL for downloading `Kafka` packages. Also, support for downloading `Kafka 0.9.0.1` is added. Removing `0.8.x` from downloads as we don't need it anymore.

### DIFF
--- a/build_bins.sh
+++ b/build_bins.sh
@@ -62,16 +62,16 @@ fi
 FILES="jmxtrans-20120525-210643-4e956b1144.zip $FILES"
 
 # Fetch Kafka Tar
-for version in 0.8.1 0.8.1.1; do
+for version in 0.9.0.1; do
   mkdir -p kafka/${version}/
-  if ! [[ -f kafka/${version}/kafka_2.9.2-${version}.tgz ]]; then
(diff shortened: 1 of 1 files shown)


cbaenziger:
Amit, I would prefer to remove 0.8.X versions.

amithkanand:
As discussed with @cbaenziger, I have removed `Kafka 0.8.x`.

build_bins.sh
```diff
@@ -62,16 +62,16 @@ fi
 FILES="jmxtrans-20120525-210643-4e956b1144.zip $FILES"
 
 # Fetch Kafka Tar
-for version in 0.8.1 0.8.1.1; do
+for version in 0.9.0.1; do
```
pu239ppy: Is there a reason why we need/needed to account for multiple versions?

pu239ppy:
I had some trouble when I copied your code into my build, since I was having Kafka download issues.  I will try to build

pu239ppy:
Built a cluster from scratch everything works now

Issue #484, opened 2016-04-08
Customizing chef-bach component configuration should be made easier


bijugs:
There will be a need to change configuration property values of individual components in ``chef-bach`` for e.g. to cater different workloads. This would mean

- Users should be able to override attribute values during cluster creation without changing the ``chef-bach`` code.
- Should be able to vary the attribute values based on the cluster workflow profile
- Should be able to dynamically generate config files like site.xml files so that additional properties can be added dynamically as done in [Kafka cookbook](https://github.com/mthssdrbrg/kafka-cookbook) so that clusters can have configuration parameters which are not available in ``chef-bach`` cookbook.

Following ``chef`` pattern, this can be done by the user using a ``wrapper`` cookbook. But as it is ``chef-bach`` doesn't expose all the configuration properties as attributes to be able to overwrite using a ``wrapper`` cookbook.

bijugs:
The following is the high level outline of the proposed changes to satisfy the first two requirements in the enhancement description and this will get us to a good starting point to be able to generate configuration files dynamically.

- New environment level attribute for cluster profile will be added. For e.g node["bcpc"]["hadoop"]["environment"]["profile"]
- All the parameters in site.xml, env.sh for which values can be customized for e.g. for cluster workload profile will be moved into attributes in chef-bach with default values. For e.g. a subset of HBase attributes on chef-bach

```
default["bcpc"]["hadoop"]["hbase_rs"]["xmn"]["size"] = 256
default["bcpc"]["hadoop"]["hbase_rs"]["xms"]["size"] = 1028
default["bcpc"]["hadoop"]["hbase_rs"]["xmx"]["size"] = 1028
```
- All the attributes which need to be changed for individual environment will be added to a ``wrapper`` cookbook attribute list. For e,g the same subset of attributes which reflects large memory sizes of cluster physical machines in  ``bach_wrapper`` cookbook

```
default["bcpc"]["hadoop"]["hbase_rs"]["xmn"]["size"] = 1024
default["bcpc"]["hadoop"]["hbase_rs"]["xms"]["size"] = 8182
default["bcpc"]["hadoop"]["hbase_rs"]["xmx"]["size"] = 8192
```
 - If there are attributes which need to have different values based on the workload profile will be set based on the cluster profile attribute value. For e.g. if some of the attributes need to be changed for read_heavy cluster the ``bach_wrapper`` cookbook attribute list can be updated to 

```
[...]

Issue #481, opened 2016-04-08
dfs.client.read.shortcircuit HBase property not set correctly


bijugs:
In the current code we have ``dfs.client.read.shortcircuit`` hbase property is [hardcoded](https://github.com/bloomberg/chef-bach/blob/06d0496b68ddebca673fbd23ea6e673c0c6a79a4/cookbooks/bcpc-hadoop/templates/default/hb_hbase-site.xml.erb#L110-L111) and at the same time the property is set based on [attribute value](https://github.com/bloomberg/chef-bach/blob/06d0496b68ddebca673fbd23ea6e673c0c6a79a4/cookbooks/bcpc-hadoop/templates/default/hb_hbase-site.xml.erb#L187-L192). We should use the latter.

bijugs:
Fixed on PR #504.

PR #485, opened 2016-04-08
Changes to customize chef-bach component configuration easier

bijugs:
This PR includes
- New attribute at the ``Chef`` environment level which can be used to identify the profile of the cluster being built i.e. read-heay or write-heavy etc
- New ``hbase`` related attributes and parametrization of ``hbase-env.sh`` and ``hbase-site.xml`` so that these values can be changed easily for e.g. say using a wrapper cookbook
- Removed duplicates properties in ``hbase-site.xml`` which was reported in issue #481.

This PR is the first of a series of changes to address enhancement request #484


amithkanand:

environments/Test-Laptop.json
```diff
@@ -41,6 +41,9 @@
         "hdp_utils": "http://public-repo-1.hortonworks.com/HDP-UTILS-1.1.0.20/repos/ubuntu12"
       },
       "hadoop" : {
+        "environment" : {
```
pu239ppy: We already specify the environment name by the **name** key

environments/Test-Laptop.json
```diff
@@ -41,6 +41,9 @@
...
+          "profile" : ""
```
pu239ppy: Either here or probably better on the source issue #484, please expand on how profile will be used.  I am starting to think, that perhaps more then a single profile may be a better approach such that we can have 
`profiles = [ "HBase_over_transatlantic_links",  "vm_install", "etc" ]`

environments/Test-Laptop.json
bijugs: @pu239ppy it is a known that there is a environment name in ``chef`` environment file. This attribute is for users to identify the workload profile which can be different from the environment name and if users wish they can set this attribute to the same value as the ``chef`` environment name.

environments/Test-Laptop.json
bijugs: @pu239ppy since this new attribute is for users to include logic in a ``chef`` wrapper cookbook, users can choose any data structure they would need to build the logic in the wrapper cookbook.

bijugs:
@amithkanand , since it is taking long time for reviews and merge of changes, it would be good to have small PRs so that it is easier for the reviewer and get the merges going. There is no other reason for having multiple PRs. 

Regarding squashing of the commits, is your review complete? are there any other comments? The idea of having separate commits for the reviewers to give some idea about the changes which are made for various reasons. Also since the commits in this case are for different reasons personally prefer to keep it separate but if the standard being followed is one commit per PR will squash them if your review is complete.

bijugs:
Closing to merge this change to {{friendly}} branch of {{chef-bach}}.

PR #486, opened 2016-04-09
fixing check to add proxy to bootstrap

### LINKED ISSUE #439: Fix logic to set /etc/apt/apt.conf on the bootstrap node in chef-bach/vbox_create.sh
line https://github.com/bloomberg/chef-bach/blob/master/vbox_create.sh#L247
never produces empty string anymore since there was a change (per Clay) to redirected vagrant output from stderr to stdout.
Please fix this logic so apt-get can work on the boostrap node

saurabh313:
This fixes #439 by making sure the grep check for proxy in bootstrap::/etc/apt/apt.conf is implemented correctly.


pu239ppy:
LGTM

PR #399, opened 2016-02-11
enable jmx for yarn

pu239ppy:
**BREAKING**
- For the graphite display to be useful, latest version of jmxtrans cookbook
- Role **roles/BCPC-Hadoop-Head-ResourceManager.json** needs to be assigned to both resourcemanagers

**Improvements**

- Enable cluster stat collection for YARN resourcemanager, unfortunately for this to work properly **bcpc-hadoop[recipe::resourcemanager]** had to be assigned a separate role **BCPC-Hadoop-Head-ResourceManager**.  A lot of it has to do with how jmxtrans cook book discovers the data it needs to poll.
- ResourceManager QueMetrics in graphite will display per queue statistics.  This will work properly now that https://github.com/jmxtrans/jmxtrans-cookbook/pull/15 is merged

**Testing**
**Nodemanager**
nodemanager_jmx

PR #489, opened 2016-04-09
bcpc-bach::resource_manager recipe has its own role BCPC-Hadoop-Head-ResourceManager

pu239ppy:
In #399 JMX was enabled for yarn, however as a result we needed a new roles created in order to provide hints to jmxtrans.  This is a follow up PR to remove bcpc-hadoop::resourcemanager from BCPC-Hadoop-Head-MapReduce role since it is redundant

# Changes
- removed bcpc-bach::resource_manager from BCPC-Hadoop-Head-MapReduce.json

# Breaking
Every resource manager node (previously node with BCPC-Hadoop-Head-MapReduce or bcpc-bach::resource_manager assigned) will need this roles assigned in cluster.txt.  Following that run cluster-assign-role.sh

### DIFF
--- a/roles/BCPC-Hadoop-Head-MapReduce.json
+++ b/roles/BCPC-Hadoop-Head-MapReduce.json
@@ -6,7 +6,6 @@
   "run_list": [
     "role[Basic]",
     "role[BCPC-Hadoop-Head]",
-    "recipe[bcpc-hadoop::resource_manager]",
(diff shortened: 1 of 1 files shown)


bijugs:
LGTM. As a next step we should move ``oozie`` into a new role out of the ``BCPC-Hadoop-Head-MapReduce`` role.

PR #491, opened 2016-04-12
Add -XX:MaxPermSize=512m to GC_OPTS for Hive

amithkanand:
This PR adds `-XX:MaxPermSize=512m` to `GC_OPTS` for Hive.  This will prevent:

```
java.lang.OutOfMemoryError: PermGen space
        at java.lang.ClassLoader.defineClass1(Native Method)
```


amithkanand:
Tested successfully it on my VM cluster.

bijugs:
The change LGTM. Is this the recommended value? Is there a reason why ``hive`` in ``hdp2.3.2`` requires a large permanent gen space given that the earlier version was running with the default which I think is ``64 mb``.

PR #492, opened 2016-04-12
Remove BCPC specific files

amithkanand:
This  PR removes `bcpc` specific files that are not needed in deployment of `chef-bach` clusters.


amithkanand:
This is no longer needed the code was merged to friendly.

Issue #493, opened 2016-04-12
Phoenix tracing aborts chef-run


pu239ppy:
chef-client exists with a long stack trace that boils down to the fact that SYSTEM.TRACING_STATS already exists
````
^[[1;31mAborting command set because "force" is false and command failed: "CREATE TABLE SYSTEM.TRACING_STATS (trace_id BIGINT NOT NULL, parent_id BIGINT NOT NULL, span_id BIGINT NOT NULL, description VARCHAR, start_time BIGINT, end_time BIGINT, hostname VARCHAR, tags.count SMALLINT, annotations.count SMALLINT, CONSTRAINT pk PRIMARY KEY (trace_id, parent_id, span_id));"^[[m
Closing: org.apache.phoenix.jdbc.PhoenixConnection
---- End output of "bash"  "/tmp/chef-script20160412-15193-1srstf8" ----
Ran "bash"  "/tmp/chef-script20160412-15193-1srstf8" returned 2
[2016-04-12T06:15:59-04:00] ERROR: Chef::Exceptions::ChildConvergeError: Chef run process exited unsuccessfully (exit code 1)
````

bijugs:
The changes to enable Phoenix tracing was idempotent on HDP 2.2 where the original changes were made. Looks like the behavior of ``sqlline.py`` from Phoenix must have changed in the later version included in HDP 2.3.x. Will look into this.

bijugs:
Looking ``sqlline.py`` code in HDP repo [2.2](https://github.com/hortonworks/phoenix/blob/HDP-2.2.0.0-tag/bin/sqlline.py) and [2.3.4](https://github.com/hortonworks/phoenix/blob/HDP-2.3.4.1/bin/sqlline.py), there has been a change in the way return code is [propagated up to the shell](https://github.com/hortonworks/phoenix/blob/HDP-2.3.4.1/bin/sqlline.py#L69-L71). This change is also made available on Apache Phoenix 4.6 as part of [PHOENIX-2323](https://issues.apache.org/jira/browse/PHOENIX-2323). This change lets us to conditionally run "CREATE TABLE ..."

bijugs:
Resolved in PR #496 .

PR #494, opened 2016-04-13
chef-bach-493 conditionally create Phoenix SYSTEM.TRACING_STATS table

bijugs:
This PR is to resolve issue #493 . This change will conditionally create the Phoenix tracing table which was not the case earlier. This is required due to the change in the behavior of ``sqlline.py`` Phoenix program in HDP 2.3.x compared to earlier HDP version.


pu239ppy:
Were you able to test this out and make sure that it no longer aborts the chef-client run and in general works as expected?

bijugs:
@pu239ppy , yes tested on non-kerberized and kerberized cluster.

PR #497, opened 2016-04-14
Fix for pig in HDP2.3.4

ekund:
Fix for pig in HDP2.3.4 to set number of attempted connections to yan timeline service to 0.

To test:
run pig and see if the grunt shell comes up without 30 attempts to hit the timeline service, which is not installed by default on the cluster.


bijugs:
LGTM

amithkanand:
LGTM

PR #499, opened 2016-04-15
Do not install openjdk

amithkanand:
This PR removes chef resource that installs `openjdk`. This code was removed as part of PR #422 but for some reason during merge of `HDP 2.3.4` it came back.


bijugs:
LGTM

PR #508, opened 2016-04-21
Update README.md

asears:
Fix cobbler link


cbaenziger:
Thanks @asears! I see we have the same outdated link at on https://github.com/bloomberg/cobbler-cookbook as well; if you would be interested in pushing there too?

PR #509, opened 2016-04-21
update hive config to prevent memory leaks

ekund:
This is a workaround to prevent memory leaks in hiveserver.


amithkanand:
Can we move these new parameters to attributes?

cbaenziger:
@amithkanand I think this is no longer needed?

vt0r:

PR #513, opened 2016-04-25
Added triggers for failures to retreive or parse data from graphite

dbahir:
3 triggers created
Error returned from graphite query
Error parsing graphite result
And empty result from graphite


pu239ppy:
LGTM

PR #520, opened 2016-04-28
Updating hbase bulkload staging dir to /user/hbase

saurabh313:
Updating hbase bulkload staging directory to '/user/hbase' so it can also be shared for both hbase bulkloading and mapred jobs run by hbase user.


vt0r:

PR #522, opened 2016-04-29
This will make hbase run mapreduce jobs

saurabh313:
Hbase user currently can't run mapreduce jobs since its id < min.user.id incontainer-executor.config. This change lowers the user id constraint and restarts yarn/mapred services to take changes into effect.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
+++ b/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
@@ -85,6 +85,7 @@ service "hadoop-yarn-resourcemanager" do
   subscribes :restart, "template[/etc/hadoop/conf/yarn-env.sh]", :delayed
   subscribes :restart, "template[/etc/hadoop/conf/yarn-site.xml]", :delayed
   subscribes :restart, "template[/etc/hadoop/conf/mapred-site.xml]", :delayed
+  subscribes :restart, "template[/etc/hadoop/conf/container-executor.cfg]", :delayed
   subscribes :restart, "bash[hdp-select hadoop-yarn-resourcemanager]", :delayed
 end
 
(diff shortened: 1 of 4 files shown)


cookbooks/bcpc-hadoop/recipes/resource_manager.rb
```diff
@@ -85,6 +85,7 @@
   subscribes :restart, "template[/etc/hadoop/conf/yarn-env.sh]", :delayed
   subscribes :restart, "template[/etc/hadoop/conf/yarn-site.xml]", :delayed
   subscribes :restart, "template[/etc/hadoop/conf/mapred-site.xml]", :delayed
+  subscribes :restart, "template[/etc/hadoop/conf/container-executor.cfg]", :delayed
```
pu239ppy: Do all three services need to be restarted or just job history server and resource manager.

cookbooks/bcpc-hadoop/templates/default/hdp_container-executor.cfg.erb
```diff
@@ -2,4 +2,4 @@ yarn.nodemanager.local-dirs=<%=@mounts.map{ |d| "file:///disk/#{d}/yarn/local" }
 yarn.nodemanager.linux-container-executor.group=yarn
 yarn.nodemanager.log-dirs=<%=@mounts.map{ |d| "file:///disk/#{d}/yarn/logs" }.join(",")%>
 banned.users=hdfs,yarn,mapred,bin      
-min.user.id=1000
+min.user.id=100
```
pu239ppy: Rather then doing this, can we find where hbase is being created and make a deterministic change

amithkanand: I don't believe we want to change this to 100 from 1000. Changing this to 100 will open doors for all other users and we don't want to do that.

saurabh313: I dont see hbase user being created in chef recipes. Can some one point where that happens ?

amithkanand: User `HBase` is created when `hbase` package is installed.

cbaenziger: We can pre-create the HBase user to give it a deterministic UID. I'm not particularly concerned about someone running with a "system" UID. Amit, are there particular users you would like to prohibit from running jobs which we could blacklist?

saurabh313:
I'm not able to reproduce this fix, may be I saw it wrong earlier. I played around more with the min.user.id properties and it seems that yarn is just not respecting it.

vt0r:

PR #523, opened 2016-04-30
YARN_HOME is not defined and this will error on resource manager startup

pu239ppy:
Need to use YARN_CONF_DIR
````shell
'"${YARN_CONF_DIR:-$YARN_HOME/conf}"'
````


pu239ppy:
Already tested on vm bootstrap, where I saw this fail first

Issue #531, opened 2016-05-05
Bad binding precedence in datanode recipe for HDP2.3 upgrade


cbaenziger:
In `datanode.rb` for upgrading the following packages, there is a [binding precedence failure](https://github.com/bloomberg/chef-bach/blob/ca85220c23a8fbe5f27c2b0439bd4626188c2860/cookbooks/bcpc-hadoop/recipes/datanode.rb#L16) which causes one to install `hadoop-hdfs-datanode` instead of `hadoop-2-3-4-0-3841-hdfs-datanode`.

PR #532, opened 2016-05-05
Fix binidng precedence to install proper datanode package versions

cbaenziger:
This pull request fixes the issue identified in Issue #531 with bad binding precedence for the `datanode.rb` recipe.


pu239ppy:
LGTM

PR #537, opened 2016-05-06
updated repxe script

ekund:
Clearer rebooting instructions,
Takes care of a graceful shutdown of the host before beginning,
Corrected the order of operations.


cbaenziger:
I like the graceful shutdown work! Do you have an idea on how we can still use this for a new machine (e.g. a whole machine swap or just to automate the process of PXE booting a brand new machine)?

PR #549, opened 2016-05-25
Upgrade issues

amithkanand:
This PR fixes issues that I found during upgrade of LAB cluster from 2.2 to 2.3.4.

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/mapred.rb
+++ b/cookbooks/bcpc-hadoop/attributes/mapred.rb
@@ -15,6 +15,10 @@ default[:bcpc][:hadoop][:mapreduce][:site_xml].tap do |site_xml|
   hdp_path =
     File.join('/usr/hdp',
               node[:bcpc][:hadoop][:distribution][:active_release])
+
+  hdp_apps_path =
+    File.join('/hdp/apps',
+              node[:bcpc][:hadoop][:distribution][:active_release])
   
   site_xml['mapreduce.admin.user.env'] =
     'LD_LIBRARY_PATH=' +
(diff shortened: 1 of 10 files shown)


cookbooks/bcpc-hadoop/attributes/mapred.rb
```diff
@@ -15,6 +15,10 @@
   hdp_path =
     File.join('/usr/hdp',
               node[:bcpc][:hadoop][:distribution][:active_release])
+
+  hdp_apps_path =
+    File.join('/hdp/apps',
```
cbaenziger: Should this have the HDFS path prepended?

cookbooks/bcpc-hadoop/attributes/yarn.rb
```diff
@@ -90,21 +90,29 @@
 
   site_xml['yarn.nodemanager.remote-app-log-dir'] = '/var/log/hadoop-yarn/apps'
   
-  avail_memory = node['bcpc']['hadoop']['yarn']['nodemanager']['avail_memory']
-  site_xml['yarn.nodemanager.resource.memory-mb'] =
-    avail_memory['size'] or
-    [1024, (node['memory']['total'].to_i * avail_memory['ratio']/1024).floor].max
+  yarn_max_memory = lambda do
```
cbaenziger: Huh? Why a function here. What are you avoiding referencing before it is set?

cookbooks/bcpc-hadoop/attributes/mapred.rb
amithkanand: We never had HDFS parth prepended in the past and I kept it the same way.

cookbooks/bcpc-hadoop/attributes/yarn.rb
http-418: I don't know what the underlying issue is.  This was a "try it and see."  I haven't had time to revisit the problem.

Issue #441, opened 2016-03-21
Zabbix Leader Election Idempotency


cbaenziger:
If one is cleaning up Zabbix, one could do a `DROP DATABASE zabbix` to try and reconfigure from scratch but at this time, if the `template[/usr/local/share/zabbix/leader_election.sql]` resource does not run, the stored procedure is not recreated. As such, one needs to manually remove this file.

Ideally, there would be verification of the stored procedure and not a check on this file only.

If one sees an error for `* bash[elect_leader] action run` this is likely the failure.

PR #552, opened 2016-05-27
created a script to uninstall zabbix on head nodes

ekund:
This script will:
1. stop zabbix agent and servers
2. drop the zabbix table  from mysql
3. remove zabbix api ruby gems
4. remove all software and logs from the head nodes

### DIFF
--- a/uninstall_zabbix.rb
+++ b/uninstall_zabbix.rb
@@ -1,0 +1,278 @@
+#!/usr/bin/env ruby
+#
+# uninstall_zabbix.rb
+#
+# This script coordinates the uninstall of zabbix from the head nodes.
+#
+# Provisos:
+#
+# To use from the command line:
+#
+#  1. If necessary, configure your local rubygems mirror.
+#     Replace 'http://mirror.example.com' with your actual mirror.
+#     ```
+#     bundle config mirror.https://rubygems.org http://mirror.example.com
+#     ```
+#
+#  2. Run 'bundle install --deployment' on a bootstrap node with
+#     access to a rubygems mirror.
+#
+#  3. If not already using the target bootstrap, sync the updated
+#     repository, including 'vendor' directory, to the target
+#     bootstrap.
+#
+#  4. Run 'bundle exec ./uninstall_zabbix.rb -p <mysql root password>
+#      ' to begin the process.
+#
+# It is also possible to use methods from this script at a ruby REPL
+# instead of running the script from a UNIX shell.  To load methods
+# into `irb`:
+#
+#  1. Change to the repo directory.
+#
+#  2. Verify that dependencies are installed:
+#     `bundle list`
+#
+#  3. Run irb inside the repo directory.
+#     `bundle exec irb`
+#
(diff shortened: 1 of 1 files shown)


kiiranh:
@ekund You'll have to clear Zabbix [DB customization](https://github.com/bloomberg/chef-bach/blob/4768996a896beab1b93acc04e0f1c23980120988/cookbooks/bcpc/recipes/zabbix-head.rb#L121) and [Leader election](https://github.com/bloomberg/chef-bach/blob/4768996a896beab1b93acc04e0f1c23980120988/cookbooks/bcpc/recipes/zabbix-head.rb#L142) related files too.

cbaenziger:
Please follow `rubocop` recommendations

cbaenziger:
@kiiranh #441 covers the leader election issue but could you update that issue for tuning idempotency too; I suspect that has been missed for folks on upgrade?

@ekund, perhaps tackling #441 independently would be better than doing that here?

ekund:
I have updated and cleaned up the script so that it passes rubocop, removed unneeded functions, and it can now work on both hardware and vm clusters.

uninstall_zabbix.rb
```diff
@@ -0,0 +1,274 @@
...
+  # This is really gross because Ruby 1.9 lacks Array#to_h.
+  File.readlines(File.join(repo_dir, 'cluster.txt'))
+    .map { |line| Hash[*fields.zip(line.split(' ')).flatten(1)] }
+end
+
+def repo_dir
+  File.dirname(__FILE__)
+end
+
+def check_status(c, on_fail_msg, on_success_msg, dofail = false)
```
cbaenziger: Hi Esther, could you please provide a function comment here? (What is `c`?)

E.g.
````
##################
# check_status -- check status via parsing a c object
# expects: c - a c object
#              on_fail_msg - string
#              on_success_msg - string
#              dofail - to fail for fun and profit without checking something
# side affect: prints a message directly to the user via puts()

File: uninstall_zabbix.rb [hunk 2]
```diff
@@ -0,0 +1,274 @@
...
+  end
+  mysqlcmd = 'mysql -uroot -p' + password + ' -e \'drop database zabbix;\''
+  c = Mixlib::ShellOut.new('./nodessh.sh', chef_env, vm_entry['ip_address'],
+                           mysqlcmd, 'sudo')
+  c.run_command
+  check_status(c, 'Could not drop zabbix database ', 'Dropped zabbix database',
+               true)
+end
+
+def uninstall_zabbix(chef_env)
```
cbaenziger: Maybe for sanity's sake make a mention that this is to workaround Chef-BACH #441

uninstall_zabbix.rb
cbaenziger: It looks like c is a `Mixlib::ShellOut` object?

cbaenziger:
Please squash commits and otherwise, things look good. The comment for `check_status` would be nice but is mainly a nit.

PR #554, opened 2016-05-31
Remove cgroups

ekund:
Remove attributes for cgroup creation.  This attribute will cause an issue as we upgrade our machines with the code from friendly.


ekund:
Made a mistake on the PR.  Will try this again.

PR #555, opened 2016-05-31
Remove physical CPU usage limit for Yarn containers

ekund:
Remove cgroups support so it does not interfere with the new code pulled in from the friendly branch.


amithkanand:
LGTM.

PR #556, opened 2016-06-01
Changes to use the latest version of JMXTrans

bijugs:
This PR is to pull in newer version of JMXTrans to use with chef-bach.


leochen4891:
Using the new_jmx branch, I have built a new vm cluster with Java 1.7
Have verified that jmxtrans works by checking the metrics in graphite.
Next step is to upgrade Java to 1.8

leochen4891:
Verified the java version is "1.8.0_74" and javac version is 1.8.0_74 on the vm1 node.
Verified jmxtrans works by checking the metrics in graphite.

leochen4891:
Has manually tested the jmxtrans version 254 on one worker node in a physical cluster.

PR #558, opened 2016-06-06
HBase 1.x related changes to configure bucket cache

bijugs:
The changes in this PR is to handle the difference in how bucketcache is configured in HBase 1.x.


amithkanand:
LGTM

PR #563, opened 2016-06-13
fixes #561

### LINKED ISSUE #561: #556 breaks jmxtrans install
#556 stopped using 
default['jmxtrans']['sw']="jmxtrans-20120525-210643-4e956b1144.zip"
and replaced it with
default['jmxtrans']['version'] = "254"

the default recipe relies on ['jmxtrans']['sw'] and it being a full filename to compose a download url

See https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc_jmxtrans/recipes/default.rb#L42-L43

pu239ppy:
jmxtrans bcpc recipe is unable to download the package as we cannot form a url
fixes #561


amithkanand:
LGTM

PR #566, opened 2016-06-14
use spark 2.0 preview

nimbusgo:
Change the download url for spark to use the latest 2.0 build.

I'm not able to test this currently, but wanted to create this pr to reference.


pu239ppy:
#738 will addd spark 2.x support, it turned out to be a bit more involved since the directory structure changed

PR #574, opened 2016-06-17
corrected service_cmd string in BCPC-Hadoop-Head-ResourceManager role

### LINKED ISSUE #573: ResourceManager jmx is not being collected
https://github.com/bloomberg/chef-bach/blob/master/stub-environment/roles/BCPC-Hadoop-Head-ResourceManager.json#L16 should be 
`"service_cmd": "org.apache.hadoop.yarn.server.resourcemanager.ResourceManager"
#575
#574 
filed for fix

pu239ppy:
This is intended to fix #573


ekund:
LGTM

PR #576, opened 2016-06-17
Add configs for HBASE-12954 JIRA

amithkanand:
This PR adds properties that are required for [HBASE-12954](https://issues.apache.org/jira/browse/HBASE-12954). After this one can specify a particular hostname that will be published in `Zookeeper` by `HBase` for master and regionserver.


pu239ppy:
Looks good

PR #582, opened 2016-06-23
Backup JN data to support re-pxe of JN machine

amithkanand:
This PR adds logic to backup JN data to support re-pxe of JN machine.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/journalnode.rb
+++ b/cookbooks/bcpc-hadoop/recipes/journalnode.rb
@@ -3,20 +3,17 @@ include_recipe 'bcpc-hadoop::hadoop_config'
 ::Chef::Recipe.send(:include, Bcpc_Hadoop::Helper)
 ::Chef::Resource::Bash.send(:include, Bcpc_Hadoop::Helper)
 
+hdprel=node[:bcpc][:hadoop][:distribution][:active_release]
+hdppath="/usr/hdp/#{hdprel}"
+
 %w{hadoop-hdfs-namenode hadoop-hdfs-journalnode}.each do |pkg|
-  package hwx_pkg_str(pkg, node[:bcpc][:hadoop][:distribution][:release]) do
+  package hwx_pkg_str(pkg, hdprel) do
     action :install
   end
-  hdp_select(pkg, node[:bcpc][:hadoop][:distribution][:active_release])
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc-hadoop/recipes/journalnode.rb
```diff
@@ -32,6 +32,24 @@
     action :create
     recursive true
   end
+
+ if get_config("jn_txn_fmt") then
```
cbaenziger: This block should be done outside the for loop (though due to the `not_if` guard it will be done only once regardless of the duplicate attempts to create file resources).

cookbooks/bcpc-hadoop/recipes/journalnode.rb
```diff
@@ -32,6 +32,24 @@
...
+     user 0644
+     content Base64.decode64(get_config("jn_txn_fmt"))
+     not_if { node[:bcpc][:hadoop][:mounts].all? { |d| File.exists?("/disk/#{d}/dfs/jn/#{node.chef_environment}/current/VERSION") } }
+   end
+ end
+
+ bash "unpack-jn-fmt-image-to-disk-#{d}" do
+   user "root"
+   cwd "/disk/#{d}/dfs/"
+   code "tar xpzvf #{Chef::Config[:file_cache_path]}/jn_fmt.tgz"
```
cbaenziger: Instead of preserving ownership and such, should we not use the `-p` option to tar and do a `chown -r hdfs <path>` after the tar?

cookbooks/bcpc-hadoop/recipes/journalnode.rb
```diff
@@ -32,6 +32,24 @@
...
+   notifies :restart, "service[hadoop-hdfs-journalnode]"
+   only_if { not get_config("jn_txn_fmt").nil? and not File.exists?("/disk/#{d}/dfs/jn/#{node.chef_environment}/current/VERSION") }
```
cbaenziger: I don't see the harm in running this for an old jn_txn_fmt? I think `not get_config("jn_txn_fmt").nil? and` is likely unnecessary; if `#{Chef::Config[:file_cache_path]}/jn_fmt.tgz` exists, I say we go with it. (Throw an existence check in for the tarball?)

ekund:
Looks good.

PR #585, opened 2016-06-24
Enable oozie HA

amithkanand:
This PR enables high availability for `Oozie`. The following are reqired to enable OOZIE HA:
-  At least two instances of Oozie server
- DNS record for load balancer host name
- Kerberos keytab for load balancer host name
- Re-generate all existing keytabs and include load balancer `HTTP` principal in each keytab


http-418:
This PR needs to be updated to use the `database` providers and oozie mysql databags.

http-418:
HDP docs say that Oozie HA can be enabled in any version > 2.2, so I'm going to go ahead and add this to the 2.1 milestone.  Surely we can do it.

amithkanand:
Closing this pull request. Will create a new one soon.

PR #591, opened 2016-06-29
Add HDP select for hadoop-client on NN nodes

amithkanand:
This PR adds `hdp_select` for `hadoop-client` to point to correct release.


pu239ppy:
Looks good

Issue #590, opened 2016-06-28
Hadoop/Java services doesn't restart when Java version is changed


bijugs:
In the current set-up if user changes the Java version or the flavor none of the Java based services gets restarted i.e. the new Java will not be in use by the services until it is restarted explicitly. This is not desirable.

bijugs:
Refer issue [#356](https://github.com/agileorbit-cookbooks/java/issues/356) against Java cookbook.

bijugs:
Fixed in PR #628, #620.

PR #592, opened 2016-06-30
Changes to automatically restart HBase services when Java is changed

bijugs:
This change requires ``Java`` cookbook version ``1.40.0`` where [this](https://github.com/agileorbit-cookbooks/java/pull/358) change in committed. For existing clusters (created before this change) what it means is that the ``Java`` cookbook need to be updated manually using ``knife cookbook site download java --config .chef/knife.rb``. 

Refer ``chef-bach`` issue #590 for additional details.

**To be double sure for someone who may not noticed the labels, this is a breaking change**.


ekund:
How is this change to be rolled out, as it is breaking?

bijugs:
it will break only existing clusters and it can be mitigated by staging the latest ``java`` cookbook as mentioned in the PR comments.

ekund:
Looks good.

PR #593, opened 2016-07-01
Additional hdfs config properties to allow customization through wrapper cookbooks

bijugs:
The ``hdfs`` properties included in this PR helps with reducing data loss and improve performance. Also these can be overridden through wrapper cookbooks depending on cluster needs.


cbaenziger:
Instead of adding more one off variables, how about simply extending [`node[:bcpc][:hadoop][:hdfs][:site_xml]`](https://github.com/bloomberg/chef-bach/blob/6ddf96c660903e50a408dcfd470cc571fed6a5fd/cookbooks/bcpc-hadoop/recipes/hdfs_site.rb#L3)?

bijugs:
@cbaenziger,  By adding these variables to [default[:bcpc][:hadoop][:hdfs][:site_xml].tap](https://github.com/bijugs/chef-bach/blob/15d20421b3d8e9f8b3648c3f2a4802bc279fa5cd/cookbooks/bcpc-hadoop/attributes/hdfs.rb#L28)... we are extending  ``node[:bcpc][:hadoop][:hdfs][:site_xml]``. no?

bijugs:
Initial PR was following the current style of code. Updated to not use the one off variables as suggested by @cbaenziger .

PR #601, opened 2016-07-11
fixes for repxe script

ekund:
- fix to get graceful shutdown to work
- updated Gemfile that has correct dependencies


amithkanand:
LGTM

PR #603, opened 2016-07-11
use fs statistical aggregation in hive

pu239ppy:
This will prevent the following error from happening
````
[Error 30017]: Skipping stats aggregation by error 
org.apache.hadoop.hive.ql.metadata.HiveException: [Error 30001]: StatsPublisher 
cannot be initialized. There was a error in the initialization of 
StatsPublisher, and retrying might help. If you dont want the query to fail 
because accurate statistics could not be collected, set 
hive.stats.reliable=false
````


ekund:
LGTM

PR #605, opened 2016-07-12
Generate yarn exclude file using template

amithkanand:
This PR generates `yarn.exclude` file using correct template.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/yarn_config.rb
+++ b/cookbooks/bcpc-hadoop/recipes/yarn_config.rb
@@ -3,6 +3,9 @@ include_recipe 'bcpc-hadoop::yarn_schedulers'
 include_recipe 'bcpc-hadoop::yarn_site'
 
 file "/etc/hadoop/conf/yarn.exclude" do
+  content node["bcpc"]["hadoop"]["decommission"]["hosts"].join("\n")
   mode 0644
-  content ''
+  owner 'yarn'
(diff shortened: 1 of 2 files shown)


cbaenziger:
I would like to see this stay a `file` resource since the format is very simple. Further, we should not be editing templates for cluster operations. Templates are code and with the merge of the friendly branch we have moved away from editing code for operational changes.

amithkanand:
Changed it back to a `file` resource and added a new attribute for decommissioning hosts.

cookbooks/bcpc-hadoop/recipes/yarn_config.rb
```diff
@@ -3,6 +3,9 @@
 include_recipe 'bcpc-hadoop::yarn_site'
 
 file "/etc/hadoop/conf/yarn.exclude" do
+  content node["bcpc"]["hadoop"]["decommission"]["hosts"].each {|hst| puts hst}
```
kiiranh: Nit: For readability, `|host| puts host` is much better than `|hst| puts hst`.

amithkanand:
This is tested now and is working.

bijugs:
@amithkanand , this change tested in a new VM cluster? It fails since user ``yarn`` is not on the nodes when this resource executes. User ``yarn`` I believe gets created when ``yarn`` software gets installed. Refer issue #615 for details.

PR #607, opened 2016-07-15
Change the permissions on the history server directories

bijugs:
The current permissions doesn't allow ``historyserver`` process to delete elements created in the history directory while ``map-reduce`` jobs are run. The changes in the PR is to set the correct permissions so that ``historyserver`` which is running as ``mapred`` user that belongs to ``mapred`` group can rename/delete elements under  ``/user/history/done_intermediate`` and ``/user/history/done`` directories.


bijugs:
Testing Done:
- Built a new cluster with the changes
- Ran a map reduce job as user ``ubuntu``
- Verified that the job entry and the logs shows up on history server UI

PR #608, opened 2016-07-15
Download homebrew cookbook

bijugs:
Current version of ``java`` cookbook depends on ``homebrew`` and this PR is to download the same. Without this change building of VM cluster will fail.


pu239ppy:
Looks good

Issue #609, opened 2016-07-15
VM cluster build fails due to failure in starting hbase


bijugs:
Start of ``hbase`` master fails which prevents a new VM cluster to be built. The following is the error.

```
Mixlib::ShellOut::ShellCommandFailed
------------------------------------
Expected process to exit with [0], but received '1'
---- Begin output of /etc/init.d/hbase-master restart ----
STDOUT:
STDERR: /etc/hbase/conf/hbase-env.sh: line 10: export: `-Djava.security.auth.login.config=/etc/hbase/conf/regionserver.jaas -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authe                                             nticate=false -Dcom.sun.management.jmxremote.port=10102': not a valid identifier
---- End output of /etc/init.d/hbase-master restart ----
Ran /etc/init.d/hbase-master restart returned 1
```

bijugs:
Current code is adding the following in ``hbase-env.sh`` which is causing issues when trying to start ``hbase`` services.

```
 +export HBASE_REGIONSERVER_OPTS=""$HBASE_REGIONSERVER_OPTS -Djava.security.auth.login.config=/etc/hbase/conf/regionserver.jaas" $HBASE_JMX_BASE -Dcom.sun.management.jmxremote.port=10102"
```

bijugs:
Also I don't see an ``jaas`` file entry in ``HBASE_MASTER_OPTS``. In ``hbase-env.sh`` the following is entry for ``HBASE_MASTER_OPTS``

```
export HBASE_MASTER_OPTS="$HBASE_MASTER_OPTS $HBASE_JMX_BASE -Dcom.sun.management.jmxremote.port=10101"
```

bijugs:
Looks like PR #600 introduced these bugs.

PR #610, opened 2016-07-15
Fix incorrect entry written to hbase-env.sh: issue #609

bijugs:
This is to fix issue #609 due to which ``hbase`` services were not starting. After these changes made the following are the entries in ``hbase-env.sh`` for the variables ``HBASE_MASTER_OPTS`` and ``HBASE_REGIONSERVER_OPTS``

```

export HBASE_REGIONSERVER_OPTS="$HBASE_REGIONSERVER_OPTS -Djava.security.auth.login.config=/etc/hbase/conf/regionserver.jaas $HBASE_JMX_BASE -Dcom.sun.management.jmxremote.port=10102"
export HBASE_MASTER_OPTS="$HBASE_MASTER_OPTS -Djava.security.auth.login.config=/etc/hbase/conf/hbase-server.jaas $HBASE_JMX_BASE -Dcom.sun.management.jmxremote.port=10101"
```

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/hbase_config.rb
+++ b/cookbooks/bcpc-hadoop/recipes/hbase_config.rb
@@ -65,6 +65,9 @@ end
 
 subnet = node["bcpc"]["management"]["subnet"]
 
+#
+# Add common hbase-site.xml properties
+#
 generated_values = {
   'hbase.zookeeper.quorum' => 
     node[:bcpc][:hadoop][:zookeeper][:servers].map{ |s| float_host(s[:hostname])}.join(","),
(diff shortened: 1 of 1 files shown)


amithkanand:
LGTM

cbaenziger:
Before merging another patch onto what looks like quite the quilt of a file, let's try to find a way to make this more tractable?

It looks like order in the hash is preserved as per [a change in Ruby 1.9](https://www.igvita.com/2009/02/04/ruby-19-internals-ordered-hash/) and that happens to work for the [generic-env.sh template](https://github.com/bloomberg/chef-bach/blob/29646b3fb92e91848ce6c6b9c510c87081f70534/cookbooks/bcpc-hadoop/templates/default/generic_env.sh.erb) but there is still a lot of non-obvious stuff going on here which would be nice to make explicit:
* Why do we reference variables in other variables instead of making clean, one-line assertions about how HBase will run? (E.g. `HBASE_JMX_BASE`?)
* Here or in the code can you articulate the order these variables are written in and why?
* In [lines 191 and 192](https://github.com/bijugs/chef-bach/blob/39c44f8b8c2e79af6e0012e39fba3be14ca21baa/cookbooks/bcpc-hadoop/recipes/hbase_config.rb#L191-L192) can we do something cleaner for what looks like sealing these variables in quotes?

bijugs:
@cbaenziger , probably @pu239ppy maybe able to address some of the questions you have since they seem to correspond to [this commit](https://github.com/bloomberg/chef-bach/commit/0ebd9900a35037c1cce76035dbdca8e90467fe7f#diff-c64ad000fe7f11db160ac26de5e88c96). The changes in this PR followed the current standards but from your comments looks like we haven't settled on the standards yet.

PR #611, opened 2016-07-19
Add paxos directory to support NN recovery

amithkanand:
This PR adds logic to create `paxos` directory under `JN` current directory. This is needed by `NN` to perform recovery when a `JN` node is re-installed. The `paxos` directory is only created during `initializedEdits`.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/journalnode.rb
+++ b/cookbooks/bcpc-hadoop/recipes/journalnode.rb
@@ -34,6 +34,7 @@ end
 
 jndisk="/disk/#{node[:bcpc][:hadoop][:mounts][0]}"
 jnfile="/dfs/jn/#{node.chef_environment}/current/VERSION"
+jncurrent="/dfs/jn/#{node.chef_environment}/current"
 jnfile2chk=jndisk + jnfile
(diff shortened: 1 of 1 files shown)


cookbooks/bcpc-hadoop/recipes/journalnode.rb
```diff
@@ -63,6 +64,15 @@
   action :nothing
 end
 
+directory "#{jndisk}#{jncurrent}/paxos" do
+  owner "hdfs"
+  group "hdfs"
+  mode 0755
+  action :create
+  recursive true
+  only_if { !File.directory?("#{jndisk}#{jncurrent}/paxos}") }
```
http-418: This only_if guard is unnecessary

cookbooks/bcpc-hadoop/recipes/journalnode.rb
```diff
@@ -63,6 +64,15 @@
   action :nothing
 end
 
+directory "#{jndisk}#{jncurrent}/paxos" do
```
http-418: it's best to prefer File.join() to string interpolation

PR #612, opened 2016-07-20
Pin logrotate cookbook dependency to 1.9.x, for Chef 11 compat

http-418:
This commit includes the most recent set of metadata to retain Chef 11.x support.

Most importantly, it adds 'logrotate' to the list of pinned cookbooks.  (A new version broke master this morning.)


pu239ppy:
Looks good

PR #613, opened 2016-07-20
Create HBase GC log directory and fix HBASE_REGIONSERVER_OPTS

bijugs:
This PR has two fixes

- Create ``/var/log/hbase/gc`` directory. Even though ``hbase`` processes run with ``-Xloggc`` option ``gc`` logs are not written due to the absence of this directory.
- Fix the bug in the [current code](https://github.com/ekund/chef-bach/blob/4c0c1d8bfd2857f7f8d7401061ec8a70a210ce13/cookbooks/bcpc-hadoop/recipes/hbase_config.rb#L173) so that  ``HBASE_REGIONSERVER_OPTS`` is set with all the JVM related parameters so that they take in effect when ``hbase`` region server processes start.


amithkanand:
LGTM

PR #616, opened 2016-07-20
Use Berkshelf for VM builds

http-418:
This PR uses a Berkshelf provided by ChefDK to resolve dependency for test builds in VM clusters.

This is a breaking change only because it updates build_bins.sh with the expectation of a ChefDK 0.9.0 package.

### DIFF
--- a/Berksfile
+++ b/Berksfile
@@ -32,7 +32,6 @@ cookbook 'kafka-bcpc', path: './cookbooks/kafka-bcpc'
 #
 cookbook 'apt'
 cookbook 'chef-client'
-cookbook 'chef-ingredient'
 cookbook 'java'
 cookbook 'maven'
 cookbook 'ntp'
@@ -40,27 +39,25 @@ cookbook 'pam'
 cookbook 'ubuntu'
 
 #
-# Cookbooks absent from the supermarket
+# Transitive dependencies that have been forked for one reason or another.
 #
 
 # bfd is not on the supermarket
 cookbook 'bfd',
   git: 'https://github.com/bloomberg/openbfdd-cookbook'
 
+
 # cobblerd forked, pending destruction of earth by moon.
-cookbook 'cobblerd',
-  git: 'https://github.com/cbaenziger/cobbler-cookbook',
-  branch: 'cobbler_profile'
+cookbook 'cobblerd', 
+  git: 'https://github.com/bloomberg/cobbler-cookbook'
 
 # jmxtrans 1.0+ isn't on the supermarket.
-cookbook 'jmxtrans',
-  git: 'https://github.com/jmxtrans/jmxtrans-cookbook',
-  revision: 'd8ee2396eca91ef8e0e558490bde075869b787de'
+cookbook 'jmxtrans', 
+  git: 'https://github.com/bijugs/jmxtrans-cookbook',
+  branch: 'ver-2.0'
 
 # 'kafka' has an entry on the supermarket, but it's the wrong cookbook.
 cookbook 'kafka',
   git: 'https://github.com/mthssdrbrg/kafka-cookbook.git',
-  revision: '0e05b5bf562c39b7c1a2e059412be91b6402703f'
-
-cookbook 'graphite_handler',
-  git: 'https://github.com/mkoni/chef-graphite_handler.git'
+  tag: 'v2.0.2'
+ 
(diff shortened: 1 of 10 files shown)


setup_chef_cookbooks.sh
```diff
@@ -53,25 +53,5 @@ no_proxy no_proxy_string
...
-    rm ${cookbook% *}*.tar.gz
-  fi
-done
-if [[ ! -d kafka ]]; then
-  git clone https://github.com/mthssdrbrg/kafka-cookbook.git kafka
-fi
-[[ -d jmxtrans ]] || git clone https://github.com/bijugs/jmxtrans-cookbook.git -b ver-2.0 jmxtrans
-[[ -d cobblerd ]] || git clone https://github.com/bloomberg/cobbler-cookbook.git cobblerd
-[[ -d bfd ]] || git clone https://github.com/bloomberg/openbfdd-cookbook.git bfd
+/opt/chefdk/bin/berks vendor ./cookbooks
```
cbaenziger: Is this idempotent and does this work in disconnected clusters (as `build_bins.sh` is run with no Internet access).

cbaenziger: Yea!

cbaenziger:
Awesome to see this! Sad it's been nearly 1 1/2 years since Chef-BACH #104 ...

setup_chef_server.sh
```diff
@@ -54,6 +63,7 @@ else
   # we can take about 45 minutes to Chef the first machine when running on VMs
   # so follow tuning from CHEF-4253
   printf "erchef['s3_url_ttl'] = 3600\n" >> /etc/chef-server/chef-server.rb
+  export NO_PROXY=${NO_PROXY-127.0.0.1}
```
cbaenziger: Why set this here and not in [proxy_setup.sh](https://github.com/bloomberg/chef-bach/blob/88b52d57668492b5fb8ab376d1ea3408cf88a86b/proxy_setup.sh#L18)?

build_bins.sh
```diff
@@ -45,14 +45,15 @@ apt-get -y update
...
-
+# Install json gem first to avoid a too-new version being pulled in by other gems.
 if [[ -z `gem list --local json | grep json | cut -f1 -d" "` ]]; then
   gem install json --no-ri --no-rdoc -v 1.8.3
 fi
 
+if [[ -z `gem list --local cabin | grep cabin | cut -f1 -d" "` ]]; then
+  gem install cabin --no-ri --no-rdoc -v 0.7.2
+fi
+
```
cbaenziger: Any desire to use a `GEMFILE` and `bundler` in the future for such insanity?

cbaenziger:
Also, if we're pulling in ChefDK we should likely make sure that knife from ChefDK doesn't explode.I wrote a commit a while ago "[All for Chef 12 knife use ](https://github.com/cbaenziger/chef-bach/commit/cc438a70a4645544389e7f97c0b4d61dcc2dae1b)" for this.

http-418:
Clay is correct, I'm missing a necessary cherry-pick.   At some point I discovered installation order is important for chef vs chefdk, because they fail to use the `alternatives` system for symlinks.

This should not be merged until I locate the third commit.

http-418:
For the record, I originally wanted to get Berkshelf configured just as a gem, and forgo the chefdk dependency entirely.  This proved too much of a challenge on ruby 1.9.3, and upstream is not very interested in maintaining the Gem.

setup_chef_server.sh
http-418: proxy_setup.sh will only conditionally set NO_PROXY

build_bins.sh
http-418: Yeah, that would be better, but I don't see it happening as part of the shell script builds.

http-418:
OK, now the missing commit has been found and commits have been squashed.  This is what was missing:

```
# It's important to install the chefdk before chef, so that
# /usr/bin/knife and /usr/bin/chef-client are symlinks into /opt/chef
# instead of /opt/chefdk.
apt-get -y install chefdk
apt-get -y install chef=11.12.8-2
```

Berksfile
```diff
@@ -32,35 +32,32 @@ cookbook 'kafka-bcpc', path: './cookbooks/kafka-bcpc'
...
 cookbook 'bfd',
   git: 'https://github.com/bloomberg/openbfdd-cookbook'
 
+
 # cobblerd forked, pending destruction of earth by moon.
-cookbook 'cobblerd',
-  git: 'https://github.com/cbaenziger/cobbler-cookbook',
-  branch: 'cobbler_profile'
+cookbook 'cobblerd', 
+  git: 'https://github.com/bloomberg/cobbler-cookbook'
```
cbaenziger: Gah! Thanks for pointing this to the Bloomberg version; I'm happy I haven't accidentally broken us this long.

chefignore
```diff
@@ -2,7 +2,11 @@
 .kitchen
 .vagrant
 chef
+cookbooks
```
cbaenziger: How does this not ignore changes in `cookbooks/bcpc`, etc.? We used to have to [explicitly not ignore](https://github.com/bloomberg/chef-bach/blob/18b218eb3a5abf15d07089f860c6540cd375e4d0/.gitignore#L15-L22) them?

amithkanand:
LGTM.

PR #617, opened 2016-07-21
Do not create paxos directory during namenode format

amithkanand:
This PR makes sure that the `paxos` directory is only created when `JN` is being restored. Creating `paxos` directory during initial format of `NameNode` causes  failure to format the `hdfs` file system.


pu239ppy:
Looks good

Issue #615, opened 2016-07-20
VM cluster build fails in ``yarn_config`` recipe


bijugs:
The reason for failure is user ``yarn`` not created yet when the recipe executes.

```
Recipe: bcpc-hadoop::yarn_config
  * file[/etc/hadoop/conf/yarn.exclude] action create
    * cannot determine user id for 'yarn', does the user exist on this system?
================================================================================
Error executing action `create` on resource 'file[/etc/hadoop/conf/yarn.exclude]'
================================================================================


Chef::Exceptions::UserIDNotFound
--------------------------------
cannot determine user id for 'yarn', does the user exist on this system?


Resource Declaration:
---------------------
# In /var/chef/cache/cookbooks/bcpc-hadoop/recipes/yarn_config.rb

  5: file "/etc/hadoop/conf/yarn.exclude" do
  6:   content node["bcpc"]["hadoop"]["decommission"]["hosts"].join("\n")
  7:   mode 0644
  8:   owner 'yarn'
  9:   group 'hdfs'
 10:   only_if { !node["bcpc"]["hadoop"]["decommission"]["hosts"].nil? }
 11: end
```

bijugs:
Fixed in PR #618 .

PR #618, opened 2016-07-21
Create yarn.exclude after yarn user exists

amithkanand:
This PR makes sure that `yarn.exclude` file is only created after `yarn` user is created on the machine.


bijugs:
Assuming this is for #615 . Did we test building a fresh VM cluster?

amithkanand:
Yes. This was tested on a fresh cluster build.

PR #619, opened 2016-07-21
Upgade to Java version 8u101

amithkanand:
This PR changes default `java` version for `chef-bach` to Java 8 update 101.

### DIFF
--- a/build_bins.sh
+++ b/build_bins.sh
@@ -79,27 +79,12 @@ for version in 0.9.0.1; do
 done
 
 # Fetch Java Tar
-if ! [[ -f jdk-7u51-linux-x64.tar.gz ]]; then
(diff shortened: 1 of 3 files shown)


bijugs:
- Do we need to remove the attributes to be able to support Java 7 since [Apache Hadoop](http://wiki.apache.org/hadoop/HadoopJavaVersions) and related projects like [HBase](https://hbase.apache.org/book.html#java) hasn't yet announced the full support of Java 1.8 for the version we are running?

 - Assuming we did tests to make sure all the components work fine with Java 1,8?

amithkanand:
I don't believe we support different java versions for different components.

amithkanand:
I tested this on my VM cluster and ran few sanity tests. All looks good to me.

PR #620, opened 2016-07-22
Enable hbase service restart when Java version or flavor is changed

bijugs:
``java`` community cookbook now provides a [standard resource](https://github.com/agileorbit-cookbooks/java/blob/7fdd8507991d35065c833fff3554f4642b072366/recipes/notify.rb#L29-L33) to subscribe to get notified when Java version or flavor got changed. This PR is to use this resource and restart ``HBase`` services when there is a change.

**Testing**
- Verified that the ``hbase`` services restart when Java version is changed from 7 to 8.

**Note**
- This PR requires ``java`` cookbook version >= 1.41.0 to be available in chef-server i.e. this can break chef-client runs in an existing cluster.


amithkanand:
Is this only required for `HBase`? What about `NameNode`, 'DataNode`, `NodeManager` and other `Hadoop` related services?

bijugs:
@amithkanand I thought we discussed about this offline y'day. Refer PR #628 raised y'day.

PR #623, opened 2016-07-25
deployment of patched hadoop-common.jar for nested ldap groups support

ekund:
This will support rolling out a jar file that has a temporary patch for nested ldap groups support.  

**This must be rolled out with a corresponding change to the Namenode and Standby Name Node roles that add the new recipe to the runlist, as well as access to the specified jar.**

In order to deploy:
1. add jar to bins directory
2. deploy_bins.sh
3. upload cookbooks
4. CAR on the name node machines.
To test:
make sure the following command would work when the nested structure looks like this:
group_parent
   |
group_child
  |
user1

hdfs groups <user1> | <grep group_parent>

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/core_site.rb
+++ b/cookbooks/bcpc-hadoop/recipes/core_site.rb
@@ -56,6 +56,8 @@ if node[:bcpc][:hadoop][:hdfs][:ldap][:integration]
 
      'hadoop.security.group.mapping.ldap.search.attr.group.name' =>
        'cn',
+     'hadoop.security.group.mapping.ldap.search.group.hierarchy.levels' =>
+       '1',
     }
   core_site_generated_values.merge!(ldap_properties)
 end
--- a/cookbooks/bcpc-hadoop/recipes/hadoop_patch.rb
+++ b/cookbooks/bcpc-hadoop/recipes/hadoop_patch.rb
@@ -1,0 +1,55 @@
+directory '/usr/local/hadoop' do
+  owner 'root'
+  group 'root'
+  mode '0755'
+  action :create
+end
+remote_file '/usr/local/hadoop/hadoop-common-2.7.1.jar' do 
+  #source 'http://10.0.100.3/hadoop-common-2.7.1.jar'
+  source "#{get_binary_server_url}/hadoop-common-2.7.1.jar"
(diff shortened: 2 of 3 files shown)


cookbooks/bcpc-hadoop/recipes/core_site.rb
```diff
@@ -56,6 +56,8 @@
 
      'hadoop.security.group.mapping.ldap.search.attr.group.name' =>
        'cn',
+     'hadoop.security.group.mapping.ldap.search.group.hierarchy.levels' =>
+       '1',
```
bijugs: Do we want this value to be configurable through an attribute? or `1` is sufficient assuming that that is the max number of hierarchy levels in a typical installation?

cookbooks/bcpc-hadoop/recipes/hadoop_patch.rb
```diff
@@ -0,0 +1,55 @@
...
+remote_file '/usr/local/hadoop/hadoop-common-2.7.1.jar' do 
+  #source 'http://10.0.100.3/hadoop-common-2.7.1.jar'
+  source "#{get_binary_server_url}/hadoop-common-2.7.1.jar"
+  checksum = 'sha256:9ed127f5a9a21c2f3efced7fd73e36d782b73b8eb9820c7ebef5169a94ded8f0'
+  owner 'root'
+  group 'root'
+  mode '0644'
+  action :create
+end
+# backup the previous file
```
bijugs: Do we want to execute this on the condition that the attribute [default["bcpc"]["hadoop"]["distribution"]["release"]](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/attributes/default.rb#L10) being set to ``2.3.4.0`` so that the code will be fine when new release with the ``hdfs`` patch comes in?

ekund: This is a temporary fix until the new release comes out.  If we get a new release, then this patch needs to be removed.  We could put in a guard if you think that is necessary.

cookbooks/bcpc-hadoop/recipes/core_site.rb
ekund: As each additional hierarchy is quite expensive, and our ldap only has 1 extra level, I don't foresee us ever raising it higher, and if we do need to, we'd need to think about that as it will slow down every single hdfs request.  So I think it is safe to leave it here.

amithkanand:
This is how we roll out patches. This PR is for future reference.

PR #626, opened 2016-07-27
Enable short-circuit reads by default

cbaenziger:
This enables short-circuit reads by default. We have had physical clusters happily running this on HDP 2.3.4 for some time.


pu239ppy:
Looks good

PR #628, opened 2016-07-28
Add restart logic to hdfs and yarn services for Java version/flavor changes

bijugs:
The changes are to restart ``hdfs``, ``yarn`` and ``zookeeper`` services when ``java`` version or flavor is changed in a cluster. The change is same as the one in PR #620 which was for ``hbase`` services.

**Note**
As mentioned in PR #620, this change requires ``java`` cookbook version >= 1.41.0 i.e this is a *breaking* if the ``java`` cookbook used doesn't match this requirement.


http-418:
This is no longer a breaking change, as Biju's update to metadata.rb prevents upload in the event of a stale cookbook.

PR #629, opened 2016-07-29
prevent precise-server-cloudimg-amd64-vagrant-disk1.box from being uploaded

pu239ppy:
We don't want this file to vendor as it is large and will cause "Entity too large" from nginx when cookbooks are being uploaded

### DIFF
--- a/chefignore
+++ b/chefignore
@@ -12,3 +12,4 @@ provisioning
 roles
 vendor
 vbox
+precise-server-cloudimg-amd64-vagrant-disk1.box


http-418:
Lev, how did you end up with a .box file in the top level? On my builds that file is downloaded to the vbox subdirectory, which is already in `chefignore`

pu239ppy:
It is possible that at some point 
https://github.com/bloomberg/chef-bach/blob/master/vbox_create.sh#L40
Did not work properly, I will try to do another build and see

cbaenziger:
Lev would `*.box` be a better pattern anyhow?

PR #633, opened 2016-08-03
Should be able to modify the retention period of aggregated YARN logs

bijugs:
Depending on the environment the number of days aggregated ``yarn`` application logs should be retained for different period. This change is to enable that.


cookbooks/bcpc-hadoop/recipes/yarn_site.rb
```diff
@@ -1,3 +1,6 @@
+node.default['bcpc']['hadoop']['yarn']['site_xml']['yarn.log-aggregation.retain-seconds'] =
```
pu239ppy: The default for this is -1 for never expire, while I agree that we want to clean out these files once in a while, can we also have a setting that if default['bcpc']['hadoop']['yarn']['yarn.log-aggregation.retain-days'] == -1 just bypass all the math and set to yarn.log-aggregation.retain-seconds -1?

bijugs:
Thanks @pu239ppy for the review comments. Updated the PR with changes to allow disabling deletion of aggregated ``yarn`` log.

pu239ppy:
Looks good

PR #637, opened 2016-08-05
Changes to be able to modify HBase master JVM heap size

bijugs:
Currently HBase master JVM runs with the default heap size of Java. This PR is to enable users to be able to modify the heap size of the master JVM process.


ekund:
LGTM

PR #641, opened 2016-08-11
Add a plugin to set the VBox Machine Folder before 'vagrant up'

http-418:
This commit adds an inline plugin to automatically set the VBox
default machine folder to a local directory before 'up' when using
Vagrantfile.baremetal.


cbaenziger:
Would you think it prudent to remove the other place this was [attempted](https://github.com/bloomberg/chef-bach/blob/e9f40c07be07afabb2884a0078d7d3c1b31f3fbf/Vagrantfile.baremetal#L117)?

vt0r:

Issue #644, opened 2016-08-15
YARN unusable after PR#635?


cbaenziger:
I believe this results in the following in `yarn-site.xml`:
```
export YARN_OPTS="-Dhadoop.log.dir=#{env_sh[:YARN_LOG_DIR]} -Dyarn.log.dir=#{env_sh[:YARN_LOG_DIR]} " -Dhadoop.log.file=#{env_sh[:YARN_LOGFILE]} -Dyarn.log.dir=#{env_sh[:YARN_LOGFILE]} -Dyarn.id.str=#{env_sh[:YARN_IDENT_STRING]} -Dhadoop.root.logger=INFO,CONSOLE -Dyarn.root.logger=INFO,CONSOLE -Dyarn.policy.file=#{env_sh[:YARN_POLICYFILE]} -Djute.maxbuffer=#{node[:bcpc][:hadoop][:jute][:maxbuffer]}"
export YARN_NODEMANAGER_OPTS="-Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=3131"
export YARN_RESOURCEMANAGER_OPTS="-Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.port=3131"
```

This results in the following error:
````
$ yarn logs -applicationId application_1470689085082_14432
/usr/hdp/2.3.6.2-2/hadoop/conf/yarn-env.sh: line 20: unexpected EOF while looking for matching `"'
/usr/hdp/2.3.6.2-2/hadoop/conf/yarn-env.sh: line 21: syntax error: unexpected end of file
/var/log/hadoop-yarn/apps/yarn/logs/application_1470689085082_14432 does not exist.
Log aggregation has not completed or is not enabled.
````

bijugs:
Looks like this may be a pre PR#635 issue. Verified by reverting the change in #635 and C-A-R ed Yarn RM node.

jmh045000:
It doesn't make it unusable, but I think this may be the reason there's difference in DNJ2-GEN and DOB2-GEN yarn log aggregation.

cbaenziger:
Thanks Biju

PR #645, opened 2016-08-16
Fixes unresolved values in YARN_OPTS environment variable

bijugs:
This was reported in issue #644 .

**Testing**
- [x] - verify that the contends of ``yarn-env.sh`` are correct on ``yarn`` ``resourcemanager`` and ``nodemanager`` nodes.


pu239ppy:
Looks good

PR #648, opened 2016-08-24
Generate correct memory configuration for Yarn/MapReduce

amithkanand:
This PR makes sure that the configuration for `MapReduce` memory parameter is calculated correctly based on value defined in the wrapper cookbook.


ekund:
What is the reason to move the memory config out of the attributes into the recipe file?

ekund:
Putting it in recipe so overrides are easier, I see.

Issue #652, opened 2016-08-29
Incorrect port set for Yarn app attempt log


amithkanand:
After  `BACH-341` the functionality for viewing `Yarn` app attempts log is broken due to incorrect port set for [`yarn.log.server.url`](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/yarn_site.rb#L225). The correct port is `19888`.

cbaenziger:
Looks like this can be closed...

PR #653, opened 2016-08-29
Set correct port for yarn.log.server.url

amithkanand:
This PR fixes the port number for `yarn.log.server.url` reported via issue #652


ekund:
LGTM

Issue #656, opened 2016-08-29
UseConcMarkSweep for HBase and java.net.preferIPv4Stack=true for all daemons


cbaenziger:
The merge of friendly introduced a number of regressions with regards to how we initialize Java for HBase and YARN:

For HBase we started using the serial GC:
In our recipe [hbase_config:167](https://github.com/bloomberg/chef-bach/blob/edc6e54727658fc15d8e4225f32234c71712458f/cookbooks/bcpc-hadoop/recipes/hbase_config.rb#L167) we try to set use CMS and set Java to prefer IPv4. Unfortunately we override this `node` attribute with:
[hbase_config:210](https://github.com/bloomberg/chef-bach/blob/edc6e54727658fc15d8e4225f32234c71712458f/cookbooks/bcpc-hadoop/recipes/hbase_config.rb#L210) which does not pull in the previously configured options.

This means region servers use serial (pause the world) GC!

Further, in [`yarn-env.sh`](https://github.com/bloomberg/chef-bach/blob/736fc13c094e0b19c95357a6c10d28919dc6f5c6/cookbooks/bcpc-hadoop/attributes/yarn.rb) no where do we set `-Djava.net.preferIPv4Stack=true`.

amithkanand:
This is taken care by PR #655

PR #655, opened 2016-08-29
Generate Heap Dump on OutOfMemory Error

amithkanand:
This  PR adds `GC` options to generate `Heap` dump when a `Java` process goes `OOM`.

Processes now seem to have correct options:
* HBase Master
```
[...]

PR #657, opened 2016-08-30
fix hbase params for issue #656

ekund:
Include  -XX:+UseConcMarkSweepGC as well as  -Djava.net.preferIPv4Stack=true for hbase master and RS.


ekund:
This is being taken care of in [https://github.com/bloomberg/chef-bach/pull/655|#655]

Issue #654, opened 2016-08-29
Provide a way to overwrite GC options through hbase-env.sh


amithkanand:
The current code base creates a [local variable](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/hbase_config.rb#L164) to generate configurations for `hbase-env`. This prevents overwriting or adding any additional values through environment or wrapper cookbook.  We need a chef attribute instead of local variable to be able to provide custom values through wrapper/environment.

http-418:
Whoops, this issue was fixed long ago in PR #659 and we kept adding it to milestones.

PR #659, opened 2016-09-07
Environment file changes

amithkanand:
This PR provides a node attribute for `hbase-env.sh` file. The issue was reported via issue #654. Also, it standardized other `*-env.sh` files.


cbaenziger:
Lines 164, 167, 170, 179, 184 in  cookbooks/bcpc-hadoop/recipes/hbase_config.rb  all need to be `node.default`...

PR #660, opened 2016-09-07
change the filename of nodemanager log

pu239ppy:
1. a default to include the hostname
2. add -resourcemanager or -nodemanager


pu239ppy:
Looks like this may already have been done elsewhere

Issue #663, opened 2016-09-09
Master build is failing without http_proxy on Mac


amithkanand:
I am getting following error while building chef-bach master branch on a Mac and Ubuntu Desktop

```
+ USER=vagrant
+ ENVIRONMENT=Test-Laptop
+ load_binary_server_info Test-Laptop
+ environment=Test-Laptop
+ bootstrap_server_key='["override_attributes"]["bcpc"]["bootstrap"]["server"]'
+ binary_server_key='["override_attributes"]["bcpc"]["binary_server_url"]'
+ load_json_frag='import json; print json.load(file('\''environments/Test-Laptop.json'\''))'
++ python -c 'import json; print json.load(file('\''environments/Test-Laptop.json'\''))["override_attributes"]["bcpc"]["binary_server_url"]'
++ echo -n http://
++ python -c 'import json; print json.load(file('\''environments/Test-Laptop.json'\''))["override_attributes"]["bcpc"]["bootstrap"]["server"]+'\'':80'\'''
Traceback (most recent call last):
  File "<string>", line 1, in <module>
IOError: [Errno 2] No such file or directory: 'environments/Test-Laptop.json'
+ export binary_server_url=http://
+ binary_server_url=http://
++ ruby -e 'require '\''uri'\''; print URI('\''http://'\'').host'
/usr/lib/ruby/1.8/uri/common.rb:406:in `split': bad URI(absolute but no path): http:// (URI::InvalidURIError)
        from /usr/lib/ruby/1.8/uri/common.rb:485:in `parse'
        from /usr/lib/ruby/1.8/uri/common.rb:608:in `URI'
        from -e:1
+ export binary_server_host=
+ binary_server_host=
+ [[ -f .chef/knife.rb ]]
+ mkdir .chef
+ cat
++ pwd
++ pwd
++ pwd
+ mkdir -p ./vendor
+ /opt/chefdk/bin/berks vendor ./vendor/cookbooks
[2016-09-09T13:22:33+00:00] WARN: Ohai::Config[:log_level] is set. Ohai::Config[:log_level] is deprecated and will be removed in future releases of ohai. Use ohai.log_level in your configuration file to configure :log_level for ohai.
[2016-09-09T13:22:33+00:00] WARN: Ohai::Config[:disabled_plugins] is set. Ohai::Config[:disabled_plugins] is deprecated and will be removed in future releases of ohai. Use ohai.disabled_plugins in your configuration file to configure :disabled_plugins for ohai.
[...]

PR #674, opened 2016-09-21
Correct http_proxy issue and failure to find environment/Test-Laptop.json

cbaenziger:
This should fix Chef-BACH #663. It works by moving the expansion of `cluster.tgz` to before anything needing `../cluster` is run. Particularly the following is fixed:
````
++ python -c 'import json; print json.load(file('\''environments/Test-Laptop.json'\''))["override_attributes"]["bcpc"]["bootstrap"]["server"]+'\'':80'\'''
Traceback (most recent call last):
  File "<string>", line 1, in <module>
IOError: [Errno 2] No such file or directory: 'environments/Test-Laptop.json'
````
As it was caused by the following execution chain:
* [`bootstrap_chef.sh`:77](https://github.com/bloomberg/chef-bach/blob/9600c76c1288501e341cd6e82ee4b884d5fbf0c0/bootstrap_chef.sh#L77) calls `setup_chef_cookbooks.sh`
* [`setup_chef_cookbooks.sh`:16](https://github.com/bloomberg/chef-bach/blob/9600c76c1288501e341cd6e82ee4b884d5fbf0c0/setup_chef_cookbooks.sh#L16) calls `load_binary_server_info`
* [`load_binary_server_info`](https://github.com/bloomberg/chef-bach/blob/9600c76c1288501e341cd6e82ee4b884d5fbf0c0/proxy_setup.sh#L36) is defined in `proxy_setup.sh` and references `environments/${environment}.json`
* [`environments`](https://github.com/bloomberg/chef-bach/blob/6f9131102002525778c3cfa4309949d03003c1c9/environments) is a symlink to `../cluster/environments` -- which does not exist at this point

It is unsatisfying that this happened to work for folks in the past but I believe that to be due to poor state clean-up of the `../cluster` directory in the past. Any new clone in a new directory location would hit this issue.

### DIFF
--- a/bootstrap_chef.sh
+++ b/bootstrap_chef.sh
@@ -69,6 +69,9 @@ else
 fi
 
 
+# cluster external environment will go here
+#$SSH_CMD "mkdir -p cluster"
+$SSH_CMD "tar -xzf /chef-bcpc-host/cluster.tgz -C /home/vagrant" || (echo '### Vagrant ssh failed to deploy cluster environment returned $? ###'; exit 1)
 echo "Building bins"
 $SSH_CMD "cd $BCPC_DIR && sudo ./build_bins.sh"
 echo "Setting up chef server"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: bootstrap_chef.sh [hunk 1]
```diff
@@ -69,6 +69,9 @@ else
 fi
 
 
+# cluster external environment will go here
+#$SSH_CMD "mkdir -p cluster"
```
http-418: What does this mean?

PR #668, opened 2016-09-15
all hbase-env generation moved to recipe

pu239ppy:
Addresses #667 
This is done to make sure we can modify defaults with wrapper cookbooks  and those values are used.

PR #675, opened 2016-09-21
Re-order hbase env attribute loading to fix PR 668

http-418:
This PR should fix both the node.set issue, and the attribute precedence issues that pull #668 sought to remedy.

I moved most attributes back into attribute files, and deleted duplicate settings.  We were setting the hbase env settings for JMX and Kerberos twice, in two different recipes.  Lev's PR caught one of two, and the second ended up being a nil access without the first one happening.


http-418:
This is no longer needed, as we've located a corrected set of changes.

PR #683, opened 2016-10-08
Perform a BFD install in VM cluster builds

http-418:
I discovered last night that hardware cluster builds are currently broken out of the box, because the BFD package is absent.

This PR adds BFD to all builds, regardless of target.

Additionally, chef runs can fail on a bootstrap with only 1024 MB of RAM.  I've upgraded the bootstrap to 2048 MB of RAM.  (Forcing macbooks to swap is ostensibly better than having the build fail)


pu239ppy:
Looks good

PR #636, opened 2016-08-04
UEFI and Ubuntu 14.04 support for chef-bach

http-418:
This PR is a roll-up of a month's work to get chef-bach working on Ubuntu 14.04.   (Unfortunately most of the commit history is lost even if we wanted it.)

Current Testing Status
------------------
- [x] supports Ubuntu 14.04 "Trusty" clients
- [x] supports Ubuntu 12.04 "Precise" clients
- [x] VM clusters are expected to converge cleanly on the first try.   
- [x] UEFI hardware boot
- [x] Legacy BIOS hardware boot (expected to work, but untested.)
- [x] HDFS
- [x] Map/Reduce
- [x] Oozie
- [x] HBase
- [x] Hive
- [x] Spark
- [x] Amit's HBase performance tool test

Additional Test Cases
-------------------
- [x] upgrade a 12.04 bootstrap to 14.04 
- [x] Build a type 1 machine with 12.04 - boot
- [x] boot a type 1 machine with 14.04 - boot
- [ ] build a type 1 hardware 12.04 hadoop cluster (head and worker node)
- [ ] build a type 1 hardware clusrer with 14.04 (head and worker node) 
- [ ] upgrade a 12.04 cluster to 14.04 (bootstrap, head, and workers)
- [x] build a type 2 machine with 14.04 code - boot
- [x] add a type 2 14.04 machine to a type 1 14.04 cluster (worker and head)

To Do
-------
 - [x] Fix check for Oozie schema version before Oozie schema installer is executed
 - [x] Centralized internal apt mirrors

Major changes
-------------------
* As part of the upgrade process, the MySQL, Zabbix, Graphite, and Oozie recipes were rewritten to use the 'database' cookbook providers.
* Cobbler has been upgraded to 2.6, using upstream packages.
* MySQL has been updated from Percona XtraDB 5.5 to Percona XtraDB 5.6
* Bootstrap VMs now rely on Ubuntu 14.04 with the 'trusty' LTS kernel.
* Heads and workers use Ubuntu 14.04 with the 'xenial' HWE kernel.
* NIC bonding is performed by Chef after first boot
* Mirrors are no longer hosted on the bootstrap node.

Detailed history
-------------------
 - First attempt at updating shell scripts for Ubuntu 14.04
 - Make cobbler recipe pass rubocop
 - Remove hardcoded paths from gem installation recipes.
 - Update bcpc::ufw recipe for 14.04 and rubocop compliance
 - Update bcpc::apache-mirror for 14.04
[...]

PR #685, opened 2016-10-12
Fix regressions related to PR #636 and PR #683

http-418:
The commit history has it all:
- Add the bootstrap-hosted apt repository to the chef recipes
- Store the bootstrap's apt key in a data bag item
- Use node.run_state for memoization in bcpc utils.rb 
- Affirmatively set permissions on /var/run/mysql to avoid umask issues 
- Fix permissions-related regression in graphite-web.conf

### DIFF
--- a/cookbooks/bcpc/recipes/bootstrap.rb
+++ b/cookbooks/bcpc/recipes/bootstrap.rb
@@ -39,7 +39,6 @@ node.default[:bfd].tap do |bfd|
 end
 include_recipe 'build-essential'
 include_recipe 'bfd::install'
-include_recipe 'bfd::default'
 
 require 'pathname'
 require 'rubygems'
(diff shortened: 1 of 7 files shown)


cbaenziger:
This seems frivilous since we need the repo to install Chef but I do like having Chef ensure the repo is still there too...

cookbooks/bcpc/recipes/bootstrap.rb
```diff
@@ -109,7 +217,7 @@
   end
 
   execute 'killall bfdd-beacon' do
-    ignore_failure true
+    only_if 'pgrep bfdd'
```
http-418: `service bfdd-beacon stop` is buggy

cookbooks/bcpc/recipes/mysql.rb
```diff
@@ -100,6 +100,11 @@
           '-o Dpkg::Options::="--force-confold"'
 end
 
+directory '/var/run/mysql' do
```
http-418: The percona 5.6 init script doesn't expect Chef's funny umask

cookbooks/bcpc/recipes/cluster_local_repository.rb
```diff
@@ -0,0 +1,111 @@
...
+  components ['main']
+
+  #
+  # If we have a data bag item for the public key, use it.
+  # If we don't, trust the repo blindly.
+  #
+  if bcpc_public_key
+    key node.run_state['bootstrap_gpg_fingerprint']
+    trusted false
+  else
```
http-418: I replaced it with a get_config! because the key should never be missing from the repo -- the bootstrap should be chef'd with 2.0 before any nodes do.

cbaenziger:
Thanks for the changes; all looks happy now!

PR #687, opened 2016-10-14
Restart HBase Master and ResourceManager when ldap password file is created

amithkanand:
This PR restarts `HBase` master and `ResourceManager` when `LDAP` password file is created/modified.


pu239ppy:
Looks good

PR #690, opened 2016-10-19
Disable NIC renaming to make installs go more smoothly on hardware with PCI NICs

http-418:
Ubuntu bundles udev rules to store persistent names for NICs.  On our current hardware, this can result in install-time enumeration orders being persisted to run-time, which breaks our ability to hardcode the 'slave' NICs in a bonding configuration.

This PR includes a "late_command" for the preseed and run-time changes in recipes to prevent NIC renaming.

### DIFF
--- a/cookbooks/bcpc/recipes/networking.rb
+++ b/cookbooks/bcpc/recipes/networking.rb
@@ -55,6 +55,26 @@ package 'powernap' do
   action :remove
 end
 
+log 'udev persistent-net.rules file found! Your NICs may have been renamed.' do
+  level :warn
+  only_if { File.exists?('/etc/udev/rules.d/70-persistent-nic.rules') }
+end
+
+#
+# Disable NIC renaming on future boots by creating an empty
+# persistent-net-generator to override the one in /lib/udev
+#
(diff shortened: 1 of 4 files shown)


cbaenziger:
It looks like for a reference on the techniques applied here one can look at [this](https://lists.ubuntu.com/archives/ubuntu-devel/2015-May/038761.html) Ubuntu-Devel mailing list thread.

cookbooks/bcpc/recipes/networking.rb
```diff
@@ -55,6 +55,26 @@
...
+  only_if { File.exists?('/etc/udev/rules.d/70-persistent-nic.rules') }
+end
+
+#
+# Disable NIC renaming on future boots by creating an empty
+# persistent-net-generator to override the one in /lib/udev
+#
+file '/etc/udev/rules.d/75-persistent-net-generator.rules' do
+  content "# This file was created by Chef.\n" \
+          "# (It is intentionally empty.)\n"
```
cbaenziger: @http-418 You're usually really consistent on these things but any reason `"` is needed here instead of `'`?

PR #691, opened 2016-10-19
Suppress Standby Failover Warning Messages in DN logs during Webhdfs

saumilmayani:
The below Logs in DN only arise during Webhdfs calls to DN. WebHDFS call is a state less engine where it does not maintain / cache in NN information. Every time a request is made, Webhdfs Client on DN looks up NN information. These are benign warning messages. This Pull request is to change the log level to "ERROR".

`org.apache.hadoop.ipc.RemoteException(org.apache.hadoop.ipc.StandbyException): Operation category READ is not supported in state standby multiple times in datanode log`


cbaenziger:
Saumil, will other consumers of this log4j config potentially see negative impacts? (Are there any less than ERROR level `org.apache.hadoop.ipc.Client` messages of interest?)

saumilmayani:
There is another one which might be of interest but not sure, if we see this often.
```
LOG.warn("Address change detected. Old: " + server.toString() +
                                 " New: " + currentAddr.toString());
```

This means that something in your DNS infra changed .

[src code](https://github.com/apache/hadoop/blob/trunk/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/ipc/Client.java)

http-418:
The problem is that there are lots of WARN errors we may want to see in the ipc.Client log4j queue.  Even some StandbyExceptions are interesting.  Just not specifically the combination of a WARN level and a RemoteException(StandbyException).

There exists a log4j filter in the log4j-extras project that can handle that level of specific filtering:
http://blog.trifork.com/2011/08/23/filtering-specific-exceptions-when-using-log4j/

The log4j-extras jar is included in HDP, but only in the flume and hive libs.  Is there an easy way for us to use this filter more broadly?

vt0r:

PR #697, opened 2016-10-21
exposing dfs.datanode.balance.max.concurrent.moves so that it may be overridden and adjust balancer bandwidth

pu239ppy:
set dfs.datanode.balance.max.concurrent.moves to multiplier constant * number of disks mounted /disk
````
node.default["hadoop"]["hdfs"]["balancer"]["max_concurrent_moves"] =
  node[:bcpc][:hadoop][:mounts].length *
  node["hadoop"]["hdfs"]["balancer"]["max_concurrent_moves_multiplier"]
````

set dfs.datanode.balance.bandwidthPerSec to the bandwidth of the interface, failing that default that can be overridden **default["hadoop"]["hdfs"]["balancer"]["bandwidth"] = 1048576**

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/default.rb
+++ b/cookbooks/bcpc-hadoop/attributes/default.rb
@@ -83,11 +83,6 @@ default["bcpc"]["hadoop"]["topology"]["script"] = "topology"
 default["bcpc"]["hadoop"]["topology"]["cookbook"] = "bcpc-hadoop"
 default["bcpc"]["hadoop"]["yarn"]["scheduler"]["minimum-allocation-mb"] = 256
 
-# Setting balancer bandwidth to default value as per hdfs-default.xml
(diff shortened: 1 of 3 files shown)


cbaenziger:
I'd really like to do away with BACH specific vanity names (e.g. `node[:bcpc][:hadoop][:balancer][:max_concurrent_moves]`); let us use what Apache has defined `dfs.datanode.balance.max.concurrent.moves`. Can we use a multiple of the number of disks (`node[:bcpc][:hadoop][:mounts]`) we have by default, instead of 5 too?

Also, while here would you mind setting `dfs.datanode.balance.bandwidthPerSec` to the interface's link speed; e.g.
````
subnet = node[:bcpc][:float][:subnet]
interface = node[:bcpc][:networks][subnet][:float][:interface]
node[:bcpc][:hadoop][:hdfs][:site_xml][dfs.datanode.balance.bandwidthPerSec] = node[:network][:interfaces][interface][:link_speed] / 8
````

Issue #700, opened 2016-10-31
Chef-BACH 2.0 MySQL Recipe Issues


cbaenziger:
The recent rewrite of the MySQL code has caused the [service](https://github.com/bloomberg/chef-bach/blob/1e698d38c1c6d7ca4bcc0cfae29cb11b4ea39036/cookbooks/bcpc/recipes/mysql.rb#L108-L110) definition to no longer specify the [supports](https://github.com/bloomberg/chef-bach/blob/1828b275cb9d141b685ddb53c1e4cc9943c5b1c3/cookbooks/bcpc/recipes/mysql.rb#L141) clause needed.

This can often cause the following error when MySQL can not be started (but is not caught since `pgrep -f mysql` returns plenty of processes):
````
  * service[mysql] action enable (up to date)
  * service[mysql] action start (up to date)
  * mysql_database_user[sst] action create
================================================================================
Error executing action `create` on resource 'mysql_database_user[sst]'
================================================================================


Mysql2::Error
-------------
Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (111)
````

By [default](https://docs.chef.io/resource_service.html) services are configured with:
> Default value: `{ :restart => false, :reload => false, :status => false }` for all platforms 
>
> (except for the Red Hat platform family, which defaults to `{ :restart => false, :reload => false, :status => true }`.)

We need the behavior from Red Hat, so that chef checks the service status instead of running `pgrep -f`

cbaenziger:
Interestingly I've also seen this but I don't know how the system got in this state:
= Can Not Perform SST =
Currently we only install `percona-xtradb-cluster-56` but to perform SST we need `percona-xtrabackup`. The dependencies of `percona-xtradb-cluster-56` are:
* percona-xtradb-cluster-server-5.6 --> depends on `percona-xtrabackup (>= 2.2.5)`
* percona-xtradb-cluster-client-5.6
* percona-xtradb-cluster-galera-3

If one does not have xtrabackup on the donor node one will see the following in the `/var/log/mysql/error.log`:
````
WSREP_SST: [ERROR] FATAL: The innobackupex version is 1.5.1. Needs xtrabackup-2.3.5 or higher to perform SST (20161031 16:34:22.314)
````
The joiner node will not have anything particularly informative in its logs, however.

Unfortunately the system is no longer in this state to see if `apt-get install -f` would complain.

PR #701, opened 2016-11-02
MySQL fixes for github issue #700

http-418:
This is a set of fixes for the issues discovered in #700


cookbooks/bcpc/recipes/mysql.rb
```diff
@@ -100,13 +100,29 @@
...
+# use 'pgrep -f' to look for the running service.  This will always
+# succeed because java processes include MySQL in their classpaths.
+#
+# Setting it to true induces chef to check service status through the
+# 'service' command.
+#
 service 'mysql' do
   action [:enable, :start]
+  supports({restart: false,
+            reload: false,
```
cbaenziger: Up above in this recipe the service is notified with `:reload` often. Is this supported by the service or should those be changed to `:restart`?

cbaenziger: Oh and `:restart` is supported by this service too...

PR #702, opened 2016-11-02
MySQL fixes for github issue #700

http-418:
Fixes for github issue #700


cookbooks/bcpc/recipes/mysql.rb
```diff
@@ -56,7 +56,7 @@
   execute "percona-preseed-#{preseed_item}" do
     command 'echo "percona-xtradb-cluster-server-5.6 ' \
       "percona-xtradb-cluster-server/#{preseed_item} " \
-      "password #{root_password}\" | debconf-set-selections"
+      "password \"#{root_password}\" | debconf-set-selections"
```
http-418: Yep.  I think it may have been silently failing in the past, not sure how it ever worked with the missing quote mark.

Issue #89, opened 2015-02-17
Axioms of Chef-BACH Architecture


cbaenziger:
We should have an architecture plan for our sprawling codebase. This would help us understand where components should go and what future development should look like.
### Today

We have code in the following cookbooks today:
- [bcpc](https://github.com/bloomberg/chef-bach/tree/28e504e928a2cf3e964bd61237963540c23cfe73/cookbooks/bcpc) -  Cluster Core OS
- [bcpc-hadoop](https://github.com/bloomberg/chef-bach/tree/28e504e928a2cf3e964bd61237963540c23cfe73/cookbooks/bcpc-hadoop)  - Cluster Hadoop Centric Ecosystem Components
- [kafka-bcpc](https://github.com/bloomberg/chef-bach/tree/28e504e928a2cf3e964bd61237963540c23cfe73/cookbooks/kafka-bcpc) - Kafka Centric Wrappers
- [bcpc_jmxtrans](https://github.com/bloomberg/chef-bach/tree/28e504e928a2cf3e964bd61237963540c23cfe73/cookbooks/bcpc_jmxtrans) - JMXTrans Wrappers
- to be upstream Chef community cookbooks - [cobblerd](https://github.com/cbaenziger/cobbler-cookbook), [hannibal](https://github.com/bloomberg/chef-bach/tree/28e504e928a2cf3e964bd61237963540c23cfe73/cookbooks/hannibal), etc.
### Proposal for the future

One direction we can move to looks like:
- `bcpc` -> `bach_os`
  - Remove all OpenStack components
- `bcpc-hadoop` -> `hadoop`
  - Eventaully this would be wrapper cookbooks around a core of community Hadoop ecosystem component cookbooks
  - Move [`graphite_to_zabbix`](https://github.com/bloomberg/chef-bach/blob/28e504e928a2cf3e964bd61237963540c23cfe73/cookbooks/bcpc-hadoop/recipes/graphite_to_zabbix.rb) to `bach_os` as it is not Hadoop related
- `kafka-bcpc` -> `bach_kafka` (or `kafka_wrapper`?)
- `bcpc_jmxtrans` -> `bach_jmxtrans` (or `jmxtrans_wrapper`?)
#### Creation of `bach_bootstrap`

In the above envisioned architecture, the `bach_bootstrap` cookbook is envisioned. It would consist of the following code today:
- `bcpc::bootstrap*`
- `bcpc-hadoop::{java,maven}_config` wrappers
[...]

PR #706, opened 2016-11-07
Replace build_bins.sh with a cookbook

http-418:
- Replace build_bins.sh with an invocation of chef-client local mode
- Add the bach_repository cookbook to idempotently create the local repo
- Install only ChefDK using shell scripts
- Use S3 mirrors for Chef tools instead of packages.chef.io
- Move spark_package step to bach_repository
- First steps towards Chef 12.x client support


cookbooks/bach_repository/attributes/default.rb
```diff
@@ -0,0 +1,43 @@
...
+# "src" contains some builds
+#
+# The decision on whether to build in src_directory vs bins_directory
+# was made by the old build_bins.sh.  For now, we are just replicating
+# the behavior of the old script.
+#
+default[:bach][:repository][:bins_directory] = '/home/vagrant/chef-bcpc/bins'
+default[:bach][:repository][:src_directory] = '/home/vagrant/chef-bcpc/src'
+
+# What is this for?
```
vt0r: ¯\\\_(ツ)_/¯

setup_chef_server.sh
```diff
@@ -20,51 +20,30 @@ if [[ -z "$CURL" ]]; then
...
-touch /etc/chef/ohai/hints/openstack.json
+apt-get -y install chef=11.18.12-1
 
 if dpkg -s chef-server 2>/dev/null | grep -q ^Status.*installed && \
    dpkg -s chef 2>/dev/null | grep -q ^Version.*11; then
   echo chef-server is installed
 else
   apt-get -y install chef-server
-  mkdir /etc/chef-server
+  mkdir -p /etc/chef-server
```
vt0r: Just in case `/etc` is missing? :smiley:

http-418: Exactly the opposite.  If `/etc/chef-server` is already present, a `mkdir` without `-p` will return an error.

vt0r: Oh wow. Just learned something new from the man page:

> no error if existing

build_bins.sh
```diff
@@ -1,354 +1,90 @@
...
-fi
-FILES="rkerberos*.gem $FILES"
-
-# Get the Rubygem for webhdfs
-if ! [[ -f gems/webhdfs.gem ]]; then
-  gem fetch webhdfs -v 0.5.5
-  ln -s webhdfs-*.gem webhdfs.gem || true
-fi
-FILES="webhdfs*.gem $FILES"
+if [ $(dpkg-query -W -f='${Status}' chefdk 2>/dev/null | grep -c 'ok installed') -eq 0 ]; then
```
cbaenziger: Are you looking for a count (`-c`) somehwere I am missing or just if the string is present (`-q`)?

build_bins.sh
```diff
@@ -1,354 +1,90 @@
...
-if ! [[ -f diamond.deb ]]; then
-  git clone https://github.com/BrightcoveOS/Diamond.git
-  pushd Diamond
-  git checkout $VER_DIAMOND
-  make builddeb
-  VERSION=`cat version.txt`
-  popd
-  mv Diamond/build/diamond_${VERSION}_all.deb diamond.deb
-  rm -rf Diamond
+    if ! sha256sum chefdk_0.12.0-1_amd64.deb | grep 6fcb4529f99c212241c45a3e1d024cc1519f5b63e53fc1194b5276f1d8695aaa; then
```
cbaenziger: Here would a `grep -q` make sense?

build_bins.sh
```diff
@@ -1,354 +1,90 @@
...
-FILES="rkerberos*.gem $FILES"
-
-# Get the Rubygem for webhdfs
-if ! [[ -f gems/webhdfs.gem ]]; then
-  gem fetch webhdfs -v 0.5.5
-  ln -s webhdfs-*.gem webhdfs.gem || true
-fi
-FILES="webhdfs*.gem $FILES"
+if [ $(dpkg-query -W -f='${Status}' chefdk 2>/dev/null | grep -c 'ok installed') -eq 0 ]; then
+    wget -nc https://opscode-omnibus-packages.s3.amazonaws.com/ubuntu/14.04/x86_64/chefdk_0.12.0-1_amd64.deb
```
http-418: I don't know how to emulate `wget -nc` behavior with curl, and wget is pre-installed on both ubuntu and osx.

cookbooks/bach_repository/attributes/default.rb
cbaenziger: If this isn't in use for generating the local apt repo any more can we remove it?

build_bins.sh
```diff
@@ -1,354 +1,90 @@
...
-if ! [[ -f $PRECISE_IMAGE ]]; then
-    while ! $(file $PRECISE_IMAGE | grep -qE '(x86 boot sector)|(ISO 9660 CD-ROM)'); do
-	$CURL -o $PRECISE_IMAGE http://archive.ubuntu.com/ubuntu/dists/precise-updates/main/installer-amd64/current/images/trusty-netboot/mini.iso
-    done
-fi
-FILES="$PRECISE_IMAGE $FILES"
+pushd $DIR/bins/ > /dev/null
 
+if [ $(dpkg-query -W -f='${Status}' chefdk 2>/dev/null | grep -c 'ok installed') -eq 0 ]; then
+    wget -nc https://opscode-omnibus-packages.s3.amazonaws.com/ubuntu/14.04/x86_64/chefdk_0.12.0-1_amd64.deb
```
http-418: It frustrates me that $CURL is set up to disable SSL verification, but I suppose it doesn't matter, since we checksum the package with sha256sum.

build_bins.sh
```diff
@@ -24,23 +24,23 @@ rm -f /tmp/build_bins_chef_config.?????????.rb
...
     fi
 fi
 
-popd
+popd > /dev/null
 
-if pgrep 'chef-client'; then
-    echo 'Chef is already running, aborting build_bins.sh'
+if pgrep 'chef-client' > /dev/null; then
+    echo 'A chef-client run is already underway, aborting build_bins.sh' 1>&2
```
http-418: It exits successfully when chef is already running.

If a chef-client is already running, then we don't need build_bins.sh to perform any operations at all -- the existing chef-client will manage the repository.

cbaenziger:
Why move the Spark package code out of `bach_spark::spark_package` into an uber cookbook? (Doesn't this defeat the goals of Chef-BACH #89 trying to split into smaller cookbooks that are self-contained per Apache project?

build_bins.sh
```diff
@@ -24,23 +24,23 @@ rm -f /tmp/build_bins_chef_config.?????????.rb
 #
 DIR=`dirname $0`
 mkdir -p $DIR/bins
-pushd $DIR/bins/
+pushd $DIR/bins/ > /dev/null
 
 if [ $(dpkg-query -W -f='${Status}' chefdk 2>/dev/null | grep -c 'ok installed') -eq 0 ]; then
```
http-418: I am not worried about counting needlessly.   This is a reliable comparison that works correctly on all known dpkg-query output.

build_bins.sh
cbaenziger: Yes, the key part (as I'm currently arguing with) is when you get a bad download; the SSL disablement is just a legacy of broken proxies some have to deal with too.

http-418: someday we will get Vagrantfile.local.rb merged into chef-bach, and that problem will go away.

build_bins.sh
http-418: That was the first iteration of the code, yes.  I later stumbled across a bug on StackOverflow that made a simple grep fail a string comparison.  Counting turned out to be the easy solution.

cookbooks/bach_repository/attributes/default.rb
http-418: It's still used, but I don't understand what it's for.

cbaenziger: This is just a version identifier for the repo, it could (and perhaps should) be the BACH release number.

http-418:
The repository of weird local packages is maintained by `bach_repository`.    In my opinion, `bach_spark` shouldn't rely on having separate recipes executed on an unknown package host to support clients.   That makes the cookbook totally, completely non-reusable, because it injects every bcpc-ism into it.

cbaenziger:
But the only bcpc-ism is the `build_bin.sh` step right? If we can do away with that, it should stand on its own? I'm concerned that we already build the equivs package for spark to be spark-<version> but further, we will get into more Spark-specific wonkiness as we have to support both Spark 1.6.X and Spark 2.X for HDP 2.5.

http-418:
The sloppy spark packages built with FPM already account for side by side installation.   The bach_spark cookbook assumes your packages are named after the version, as is the norm on Debian.

The bcpc-isms injected by adding a package creation recipe are considerable.  The recipe is only run on the bootstrap, only early in the build process, and requires in-depth knowledge of the repository structure.  This is what the bach_repository cookbook exists to do.

cookbooks/bach_repository/recipes/graphite.rb
```diff
@@ -0,0 +1,94 @@
...
+ },
+].each do |package|
+
+  log "URL: #{package[:url]}"
+
+  ark "#{package[:name]}-#{package[:version]}" do
+    url package[:url]
+    path src_dir
+    checksum package[:checksum]
+    action :put
```
cbaenziger: Is this safe for disconnected use?

http-418: This was a serious bug.  It's broadly safe, but not in our context, because ark is using the chef cache as its target directory.

http-418: (It should be fixed now)

setup_chef_server.sh
```diff
@@ -20,51 +20,30 @@ if [[ -z "$CURL" ]]; then
...
+# build_bins.sh will run chef in local mode in order to create the
+# BCPC local repo.
+#
+./build_bins.sh
 
-if ! $GEM_PATH sources | grep $(pwd)/bins; then
-   $GEM_PATH sources --add file://$(pwd)/bins/
-fi
+# Make the bootstrap a client of its own BCPC local repo
+echo "deb [trusted=yes arch=amd64] file://$(pwd)/bins/ 0.5.0 main" > \
```
cbaenziger: What does `trusted=yes` do here?

vt0r: It avoids verifying gpg signatures on a repo.

cbaenziger: @http-418 Ah cool; I understand that.

cookbooks/bach_repository/recipes/python.rb
```diff
@@ -0,0 +1,90 @@
...
+  checksum 'd1f66b3848abc6fd1aeda3bb7461101f6a909c3b08efa3ecc1f561712269469c'
+  mode 0555
+end
+
+execute 'get-pip.py' do
+  command "#{get_pip_path} #{pip_proxy_option} #{pip_cert_option}"
+  not_if {
+    version_string =
+      if File.exists?('/usr/local/bin/pip')
+        `/usr/local/bin/pip show pip | grep ^Version | awk '{print $2}'`.chomp
```
cbaenziger: Can you use shutils mixlib here please for the sake of ChefSpec?

cookbooks/bach_repository/.kitchen.yml
```diff
@@ -0,0 +1,24 @@
...
+  name: chef_zero
+  http_proxy: <%= ENV['http_proxy'] %>
+  https_proxy: <%= ENV['https_proxy'] %>
+  client_rb:
+    no_proxy: localhost
+    http_proxy: <%= ENV['http_proxy'] %>
+    https_proxy: <%= ENV['https_proxy'] %>
+
+platforms:
+  - name: ubuntu-12.04
```
cbaenziger: 14.04 might be a more modern choice?

cookbooks/bach_repository/recipes/gems.rb
```diff
@@ -0,0 +1,104 @@
...
+  gem_binary = '/usr/bin/gem'
+
+  #
+  # We are using an execute resource because gem_package does not
+  # support --local.  We must use --local because without having
+  # 'builder' already installed, it is not possible to generate the
+  # gem index.
+  #
+  execute "gem-install-#{package_name}" do
+    command "#{gem_binary} install --local #{local_gem_path}"
```
cbaenziger: Perhaps add `--no-rdoc --no-ri` since we don't need the docs?

PR #717, opened 2016-12-01
Enable Yarn Aplication Timeline Server in chef-bach

pu239ppy:
Enable application timeline server primarily to make TEZ a first class citizen in the ecosystem
This pull request obsoletes #174 

Tested that applications complete with or without the timeline server running.

**BREAKING**: 
This change requires a cluster.txt change in order to assign BCPC-Hadoop-Head-YarnTimeLineServer role to a resource manager host, since YARN Timeline server is not yet HA, cluster operator should pick any ResourceManager

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
+++ b/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
@@ -157,6 +157,7 @@ service "hadoop-yarn-timelineserver" do
   subscribes :restart, "template[/etc/hadoop/conf/hdfs-site.xml]", :delayed
   subscribes :restart, "bash[hdp-select hadoop-yarn-resourcemanager]", :delayed
   subscribes :restart, "log[jdk-version-changed]", :delayed
+  only_if { node.roles.include?("BCPC-Hadoop-Head-YarnTimeLineServer") }
 end
 
 bash "reload mapreduce nodes" do
--- a/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
+++ b/cookbooks/bcpc-hadoop/recipes/resource_manager.rb
@@ -32,13 +32,33 @@ bash "create-hdfs-yarn-log" do
   not_if "hdfs dfs -test -d /var/log/hadoop-yarn", :user => "hdfs"
 end
 
+bash "create-hdfs-yarn-ats-log" do
+  code <<-EOH
+  hdfs dfs -mkdir -p /var/log/ats && hdfs dfs -chmod 1777 /var/log/ats && hdfs dfs -chown yarn:mapred /var/log/ats
+  hdfs dfs -mkdir -p /var/log/ats/active && hdfs dfs -chmod 1777 /var/log/ats/active && hdfs dfs -chown yarn:mapred /var/log/ats/active
+  hdfs dfs -mkdir -p /var/log/ats/done && hdfs dfs -chmod 0700 /var/log/ats/done && hdfs dfs -chown yarn:mapred /var/log/ats/done
+  EOH
+  user "hdfs"
+  only_if { node.roles.include?("BCPC-Hadoop-Head-YarnTimeLineServer") }
+  not_if "hdfs dfs -test -d /var/log/hadoop-yarn/ats/active && hdfs dfs -test -d /var/log/hadoop-yarn/ats/done", :user => "hdfs"
+end
+
 # list hdp packages to install
-%w{hadoop-yarn-resourcemanager hadoop-client hadoop-mapreduce-historyserver}.each do |pkg|
+yarn_packages = %w{hadoop-yarn-resourcemanager hadoop-client hadoop-mapreduce-historyserver}
+if node.roles.include?("BCPC-Hadoop-Head-YarnTimeLineServer")
+  yarn_packages.push("hadoop-yarn-timelineserver")
+end
+
+yarn_packages.each do |pkg|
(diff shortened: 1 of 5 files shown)


pu239ppy:
Rebased on current master will rebuild

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
```diff
@@ -32,22 +32,54 @@
   not_if "hdfs dfs -test -d /var/log/hadoop-yarn", :user => "hdfs"
 end
 
+bash "create-hdfs-yarn-ats-log" do
+  code <<-EOH
+  hdfs dfs -mkdir -p /var/log/ats && hdfs dfs -chmod 1777 /var/log/ats && hdfs dfs -chown yarn:mapred /var/log/ats
```
http-418: Why is this a bash resource instead of an hdfs_directory resource?

https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/libraries/hdfsdirectory.rb

cbaenziger: I think we need to add kerberos support to `hdfsdirectory.rb`? (I'd like us to use it some day but it's been neglected; Rui did some nice work updating it but it wasn't groked at the time and has gone unmerged...)

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
```diff
@@ -32,22 +32,54 @@
...
+  hdfs dfs -mkdir -p /var/log/ats/active && hdfs dfs -chmod 1777 /var/log/ats/active && hdfs dfs -chown yarn:mapred /var/log/ats/active
+  hdfs dfs -mkdir -p /var/log/ats/done && hdfs dfs -chmod 0700 /var/log/ats/done && hdfs dfs -chown yarn:mapred /var/log/ats/done
+  EOH
+  user "hdfs"
+  only_if { node.roles.include?("BCPC-Hadoop-Head-YarnTimeLineServer") }
```
cbaenziger: Just a question: Does searching for the role seem "better" than looking in the recipe collection? (A nice [discussion](https://gist.github.com/kwilczynski/c271f251ea7f48ba99b3) of the various API's and semantics.)

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
```diff
@@ -32,22 +32,54 @@
...
   package hwx_pkg_str(pkg, node[:bcpc][:hadoop][:distribution][:release]) do
     action :install
   end
-
   hdp_select(pkg, node[:bcpc][:hadoop][:distribution][:active_release])
 end
 
+package hwx_pkg_str("tez", node[:bcpc][:hadoop][:distribution][:release]) do
+  action :install
+  only_if { node.roles.include?("BCPC-Hadoop-Head-YarnTimeLineServer") }
```
cbaenziger: Instead of all the `only_if`'s since recipes must be known before we start converging you could wrap all this in a ruby if block too right?

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
```diff
@@ -107,6 +145,20 @@
   subscribes :restart, "log[jdk-version-changed]", :delayed
 end
 
+service "hadoop-yarn-timelineserver" do
+  action [:enable, :start]
```
cbaenziger: Don't we need an `only_if { node.roles.include?("BCPC-Hadoop-Head-YarnTimeLineServer") }` around the action to determine if we are going to touch this service or not?

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
pu239ppy: I don't need anything as fine grained as a complete run list, basing decisions on roles works just fine.

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
pu239ppy: Good point

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
pu239ppy: I am not sure what you ate proposing.  Currently this a is very simple and readable solution, why complicate it more?

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
pu239ppy: The current version of webhdfs does not support kerberos, neither does the resource in question, it would be relatively trivial to add all of that, but would like to leave that and the testing of thereof for later releases

cbaenziger: Captured in #783

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
cbaenziger: I'm stylistically opposed to the role's being used for action at a distance (e.g. instead of a feature flag) but this shouldn't cause an issue so just explaining my thought.

cookbooks/bcpc-hadoop/recipes/resource_manager.rb
cbaenziger: Ah just a thought to make it more simple (fewer lines) and clearly logically group the ATS bits.

PR #718, opened 2016-12-02
Hbase TimeLine Consistency off by default

pu239ppy:
Due to a critical bug [HBASE-17238](https://issues.apache.org/jira/browse/HBASE-17238) feature is to be off by default

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/hbase.rb
+++ b/cookbooks/bcpc-hadoop/attributes/hbase.rb
@@ -82,11 +82,12 @@ default[:bcpc][:hadoop][:hbase][:site_xml].tap do |site_xml|
   site_xml['hbase.replication'] = 'true'
   site_xml['hbase.coprocessor.abortonerror'] = node["bcpc"]["hadoop"]["hbase_rs"]["coprocessor"]["abortonerror"] 
   site_xml['hbase.regionserver.storefile.refresh.period'] = 30000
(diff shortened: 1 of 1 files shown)


bijugs:
Do we want to set [node["bcpc"]["hadoop"]["hbase"]["region"]["replication"]["enabled"]](https://github.com/bloomberg/chef-bach/blob/ec69f5b3540fc88e5e9095ef9244bdcf5796919e/cookbooks/bcpc-hadoop/recipes/hbase_config.rb#L139) to ``false`` so that ``site.xml`` will not have any properties related to region replication?

pu239ppy:
Will revert #713

cbaenziger:
@bijugs I would prefer to remove and avoid creating new [manual](https://github.com/bloomberg/chef-bach/blob/ec69f5b3540fc88e5e9095ef9244bdcf5796919e/cookbooks/bcpc-hadoop/recipes/hbase_config.rb#L139-L152) attributes. Further, this block is pretty painful since it stomps one's attempt to set attributes via the hbase-site.xml values (e.g. setting both `node[:bcpc][:hadoop][:hbase][:site_xml]['hbase.meta.replica.count']` and `node["bcpc"]["hadoop"]["hbase"]["meta"]["replica"]["count"]` will not be obvious as to which one will win; further many of these settings can be statically defined in the `attributes/hbase.rb` file and only used by HBase if region replication is enabled).

PR #722, opened 2016-12-05
Allow nokogiri installation on internet-disconnected systems

http-418:
By default, nokogiri will attempt to install libxml2, libxslt, and
zlib from the internet.

The "--use-system-libraries" switch is intended to force nokogiri
extconf.rb to compile against system libraries.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/default.rb
+++ b/cookbooks/bcpc-hadoop/recipes/default.rb
@@ -57,10 +57,26 @@ gem_package 'zabbixapi' do
   action :nothing
 end.run_action(:install)
 
-gem_package "nokogiri" do
-    gem_binary gem_path
-    version ">0.0"
-    action :nothing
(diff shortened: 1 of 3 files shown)


cookbooks/bcpc-hadoop/recipes/default.rb
```diff
@@ -57,10 +57,26 @@
...
+# The "--use-system-libraries" switch is intended to force nokogiri
+# extconf.rb to compile against system libraries.
+#
+gem_package 'nokogiri' do
+  options "--clear-sources -s #{get_binary_server_url} " \
+    '-- --use-system-libraries'
+  gem_binary gem_path
+  version '>=1.6.2'
+  action :nothing
+  timeout 1800
```
cbaenziger: Just a note this is due to how long compiling nokogiri's C extensions can take we need extra time here.

PR #723, opened 2016-12-06
Ensure the bootstrap is snapshotted and Chef environment passes through

cbaenziger:
This fixes using a non-standard environment name and ensures we sanpshot the bootstrap as the cluster is built, for later roll-back when developing.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/hive_config.rb
+++ b/cookbooks/bcpc-hadoop/recipes/hive_config.rb
@@ -129,6 +129,8 @@ elsif hive_site_vars[:hs2_auth] == 'LDAP'
     'hive.server2.authentication.ldap.Domain' =>
       hive_site_vars[:hs2_ldap_domain]
   }
+else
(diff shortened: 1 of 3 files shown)


cookbooks/bcpc-hadoop/recipes/hive_config.rb
```diff
@@ -129,6 +129,8 @@
     'hive.server2.authentication.ldap.Domain' =>
       hive_site_vars[:hs2_ldap_domain]
   }
+else
+  hs2_auth_values = {}
```
http-418: What is this?

pu239ppy: if no security is specified a blank hs2_auth_values is declared so that https://github.com/cbaenziger/chef-bach/blob/0b59a8ca80818836d450e3ff745e977127f419c9/cookbooks/bcpc-hadoop/recipes/hive_config.rb#L136 does not break

PR #735, opened 2016-12-20
pin homebrew version to 2.1.2

vijayk:


http-418:
Thanks, Vijay.  I also encountered this issue this morning.

PR #747, opened 2017-01-12
Set spark.dynamicAllocation.maxExecutors

jmh045000:
Set the max executors to 2.

### DIFF
--- a/cookbooks/bach_spark/attributes/default.rb
+++ b/cookbooks/bach_spark/attributes/default.rb
@@ -21,6 +21,7 @@ default.bach_spark.config.spark.eventLog.dir = "#{node['spark']['hdfs_url']}/spa
 default.bach_spark.config.spark.eventLog.enabled = true
 default.bach_spark.config.spark.logConf = true
 default.bach_spark.config.spark.dynamicAllocation.enabled = true
+default.bach_spark.config.spark.dynamicAllocation.maxExecutors = 2
 default.bach_spark.config.spark.shuffle.service.enabled = true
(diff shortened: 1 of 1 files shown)


vt0r:
@jmh045000 - Would you mind clarifying what issue you're experiencing and how setting a static max of 2 executors resolves it, please? Is the default `infinity` utilizing too many resources somewhere? If so, perhaps we would rather base the max number on something we can pull from Ohai instead?

jmh045000:
`infinity` leaves the problem of malformed spark tasks taking all the resources within a cluster or shared queue.  For instance, reading a large, fragmented hive table can spawn >40k tasks.  With `infinity` `maxExecutors` we can end up spawning several hundred executors almost instantaneously.

Note, this can still be overridden by providing `--conf spark.dynamicAllocation.maxExecutors=0` on the command line (or some appropriate alternative).

The other resolution, that I'm not sure about, is preventing a single queue from taking an entire cluster to start with (maybe YARN's scheduler can do this, I'm not sure).

pu239ppy:
@jmh045000 You can definitely do this with schedulers

cbaenziger:
@jmh045000 I'm happy to merge this giving us the overrideable attribute then we can set that in a wrapper cookbook for our various physical clusters if you want to make the change?

vt0r:
We have made several scheduler improvements since this issue was created, and we've also _educated_ tenants who previously had abusive jobs. Things have been much better since, so I think we can safely close this.

cbaenziger:
This is better than not accepting it; merging.

PR #763, opened 2017-02-03
Allow bcpc::networking to kill interfaces overlapping bond0

http-418:
On some of our older hosts, we used a different NIC to boot from than we use to configure our bonded interfaces.  For your convenience and safety, bcpc::networking would not allow you to deconfigure your existing DHCP-configured NIC until the bond was up.

On newer hosts, we boot directly from one NIC in a degraded bond.   On these hosts, we need to proceed even though it's dangerous.  This PR adds the necessary check.


vt0r:
The change to the block that tests `default_routes.length` looks great to me, but in my own testing, I needed to move the following block lower than `bcpc-deconfigure-dhcp-interfaces`:

```ruby
execute 'bcpc-interfaces-up' do
  command 'ifup -a'
end
```

I'm also wondering if a similar change should be applied to this block as was done with `default_routes.length`, since I encountered errors in my own clusters when it did the test:

```ruby
    unless active_route_interfaces.any?
      raise 'Cannot deconfigure DHCP interfaces, ' \
        'no active interface would be configured with a default route!'
    end
```

PR #769, opened 2017-02-09
Fixes for use with both http_proxy and no_proxy

http-418:
This PR makes some minor changes to smooth VM builds in a proxied environment.

Among other things, it makes it much easier to use a mixture of proxied and un-proxied mirrors.

### DIFF
--- a/bootstrap_chef.sh
+++ b/bootstrap_chef.sh
@@ -55,6 +55,17 @@ if [[ ! -f "environments/${CHEF_ENVIRONMENT}.json" ]]; then
     exit
 fi
 
+#
+# Clear stale cookbooks in case we are re-running the
+# tests/automated_install.sh script.
+#
+# This isn't entirely safe, but this script already assumes it can use
+# rsync to overwrite anything in the vendor directory. Realistically,
+# it should only be run on VM clusters.
+#
+$SSH_CMD "sudo chown -R vagrant $BCPC_DIR; \
+	  sudo rm -rf /var/chef/cache/cookbooks; \
+          sudo rm -rf $BCPC_DIR/vendor/cookbooks"
 
 if [[ -z $VAGRANT ]]; then
   if [[ ! -f $KEYFILE ]]; then
--- a/cookbooks/bcpc/recipes/cluster_local_repository.rb
+++ b/cookbooks/bcpc/recipes/cluster_local_repository.rb
@@ -51,26 +51,20 @@ if apt_uri.host
 end
 
 unless node[:fqdn] == get_bootstrap
-  ruby_block 'bcpc-apt-key-create' do
-    block do
-      require 'tempfile'
-      node.run_state['bcpc_apt_key_path'] =
-        Tempfile.new('bootstrap-gpg-key').path
-    end
-  end
+  bcpc_apt_key_path = File.join(Chef::Config[:file_cache_path],
+                                'bcpc-cluster-local-repo.gpg')
 
-  file node.run_state['bcpc_apt_key_path'] do
+  file bcpc_apt_key_path do
     mode 0444
     content Base64.decode64(get_config!('bootstrap-gpg-public_key_base64'))
   end
 
   ruby_block 'get-bootstrap-gpg-fingerprint' do
     block do
-
       require 'mixlib/shellout'
       cc = Mixlib::ShellOut.new('gpg',
                                 '--with-fingerprint',
-                                node.run_state['bcpc_apt_key_path'])
+                                bcpc_apt_key_path)
       cc.run_command
       cc.error!
       node.run_state['bootstrap_gpg_fingerprint'] =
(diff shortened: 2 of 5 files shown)


bootstrap_chef.sh
```diff
@@ -55,6 +55,17 @@ if [[ ! -f "environments/${CHEF_ENVIRONMENT}.json" ]]; then
...
 
+#
+# Clear stale cookbooks in case we are re-running the
+# tests/automated_install.sh script.
+#
+# This isn't entirely safe, but this script already assumes it can use
+# rsync to overwrite anything in the vendor directory. Realistically,
+# it should only be run on VM clusters.
+#
+$SSH_CMD "sudo chown -R vagrant $BCPC_DIR; \
```
cbaenziger: `bins` should be owned by root in places?

cookbooks/bcpc/recipes/cluster_local_repository.rb
```diff
@@ -51,18 +51,26 @@
 end
 
 unless node[:fqdn] == get_bootstrap
-  require 'tempfile'
-  bcpc_apt_key_path = Tempfile.new('bootstrap-gpg-key').path
+  ruby_block 'bcpc-apt-key-create' do
+    block do
+      require 'tempfile'
+      node.run_state['bcpc_apt_key_path'] =
+        Tempfile.new('bootstrap-gpg-key').path
```
cbaenziger: Can we not create lots of tempfiles and use something under the Chef cache dir if this is getting touched? Or do we expect to need multiple files?

proxy_setup.sh
```diff
@@ -11,9 +11,21 @@ if [ -n "$http_proxy" ]; then
...
+  if domainname | grep '(none)'; then
+    DOMAIN="$(domainname),"
+  else
+    DOMAIN=""
+  fi
+
   local_ips=$(ip addr list |grep 'inet '|sed -e 's/.* inet //' -e 's#/.*#,#')
-  export no_proxy="$(sed 's/ //g' <<< $local_ips)localhost,$(hostname),$(hostname -f),.$(domainname),10.0.100.,10.0.100.*"
-  export NO_PROXY="$(sed 's/ //g' <<< $local_ips)localhost,$(hostname),$(hostname -f),.$(domainname),10.0.100.,10.0.100.*"
+  export no_proxy="$(sed 's/ //g' <<< $local_ips)$(hostname),$(hostname -f),$DOMAIN10.0.100.,10.0.100.*,$no_proxy"
```
cbaenziger: No need for a comma between `$DOMAIN` and `10.0.100.`?

bootstrap_chef.sh
http-418: Where, and why?

proxy_setup.sh
http-418: No.

cookbooks/bcpc/recipes/cluster_local_repository.rb
http-418: I have a more permanent fix in another PR.  This change leaves the existing tempfile in place and just moves it to converge time.

proxy_setup.sh
http-418: Specifically, we need the ability to have no $DOMAIN whatsoever, so the comma has to be part of the variable for string interpolation.

bootstrap_chef.sh
cbaenziger: Because `build_bins` runs as root; perhaps the latest round of `build_bins` changes to Chef sets ownerships to `vagrant` though?

proxy_setup.sh
cbaenziger: Ah thanks; I completely missed line 21

cookbooks/bcpc/recipes/cluster_local_repository.rb
cbaenziger: Ah thanks for pointing out the `:delete` later [down](https://github.com/http-418/chef-bach/blob/513cb474436dd48bf804f3283bacffe33ed6bf7c/cookbooks/bcpc/recipes/cluster_local_repository.rb#L87-L90)

bootstrap_chef.sh
http-418: The `bach_repository` cookbook will change any permissions that need to be changed for the repo to function during the chef run.

This change makes shell scripts work, possibly before chef runs.

cookbooks/bcpc/recipes/cluster_local_repository.rb
cbaenziger: The `:delete` should have a delayed notification here to ensure it does run should there be an error between here and there...

http-418: I cherry-picked the more durable fix out of another branch, using a cache file instead of temp.

(Strictly speaking the "delete" should never have been necessary, because the file should have been cleaned up when the ruby process exited, but Chef somehow broke that feature of the library.)

PR #751, opened 2017-01-19
Enable OOZIE HA

amithkanand:
This PR enables high availability for `OOZIE`.

**BREAKING** 
On an existing cluster following needs to happen:

1. Generate an `Oozie` Keytab for the VIP host that also has `HTTP` principal for vip host
2. Take existing `Oozie` keytab from host running the service and merge it with new VIP host keytab
3. Roll out new merged keytab to the oozie host and restart oozie services.

PR #780, opened 2017-02-17
Use a default for viphost(name)

cbaenziger:
This addresses code review issues from PR #751 that went unaddressed (and corrects to spaces from tabs for the environment's `node['bcpc']['bootstrap']['additional_packages']` entry).


cookbooks/bcpc/attributes/default.rb
```diff
@@ -102,6 +102,8 @@
 default['bcpc']['bootstrap']['pxe_interface'] = 'eth1'
 default['bcpc']['bootstrap']['server'] = '10.0.100.3'
 default['bcpc']['bootstrap']['vip'] = node['bcpc']['bootstrap']['server']
+default['bcpc']['bootstrap']['viphost'] = "#{node.chef_environment}." \
+                                          "#{node['bcpc']['domain_name']}"
```
pu239ppy: lower case hostname

cbaenziger: Addressed; thanks

PR #750, opened 2017-01-19
Move Chef-BACH to Chef 12

cbaenziger:
This is some work I've been doing to move us to Chef 12; I got it to converge for a test cluster. But this still needs polish, I'll need to ensure our tests move with us and what the upgrade steps will look like for a bootstrap need to be documented (though I've moved a bootstrap and built fresh).

This depends on bloomberg/openbfdd-cookbook#2 and will depend on a pull request into bloomberg/cobbler-cookbook once I can understand why the CentOS recipe is totally horked...

This description will be re-whacked to make it more reasonable as the code matures...

PR #788, opened 2017-02-23
Multi cluster test support

cbaenziger:
This allows one to run more than a single cluster at a time, if one has a sufficiently large test machine.

Other refactoring required to make this all work:
* Clusters are now specified solely by `cluster.txt` (e.g. if you want a 6 node cluster, simply add the lines to `cluster.txt`); no more editing shell scripts and for loops
* The ideas from bloomberg/chef-bcpc#118 were implemented so we now use the same Cobbler scripts for VMs as for physical hardware; no more `enroll_cobbler.sh` only `cluster-enroll-cobbler.sh` remains.
* The `lib/cluster_data` library was refactored and grew a number of rspec tests. (To run rspec one must add Gems to the Gemfile -- is fixed in Chef-BACH #750) and was refactored into three modules:
   * `BACH::ClusterData` -- functions only needing the Git repo
   * `BACH::ClusterData::HypervisorNode` -- functions needing VirtualBox
   * `BACH::ClusterData::ChefNode` -- functions needing a Chef Client or Chef Server
* Now many more consumers use the `lib/cluster_data` code instead of reimplementing in Ruby or Shell:
  * `uninstall_zabbix.rb`
  * `vm_to_cluster.rb` -- implements `vm_to_cluster.sh` (used to generate a usable `cluster.txt` for VM clusters
  * `Vagrantfile`
* To help in catching EFI issues `wait_for_hosts.rb` now tells if a VM downloaded the Linux `initrd.gz` and `linux` images to boot the installer
* Windows and non-Vagrant code paths were removed to clean-up the shell scripts for future obliteration


cbaenziger:
Hrm, just caught in testing this is inserting the cluster name twice into hostnames

aespinosa:
Was looking at virtualbox intnet https://www.vagrantup.com/docs/virtualbox/networking.html#virtualbox-internal-network and https://www.virtualbox.org/manual/ch06.html#network_internal 

* It doesn't have a dhcpserver by default.  We can prune checks in shellscripts for that
* There's no vboxnet* in the host so each instance of a cluster in a single virtualbox session don' t clobber each other.   (In my test, I have vboxnet[0-6]) spun up in my host (one user per VM cluster).  But networking scripts break because some hare hardwired to use vboxnet[0-3]

http-418:
the 'intnet' stuff sounds good, Allan.  We already don't really support old versions of VirtualBox, so relying on newer features doesn't bother me at all.

either way, using vboxnet* interfaces or intnet, provisioning scripts have to be aware of the mappings between the chef-bach multi-home networks and the layer 2 networks provided by virtualbox.  Using intnet with the human-readable names presumably gives us a larger namespace than just 0 to 128, though!

cbaenziger:
I just rebased; testing a new build now. I included the `knife.rb` changes from #896 in here too.

cbaenziger:
This actually built me a cluster!

vm_to_cluster.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
aespinosa: I don't see this referenced anywhere.  Is this supposed to be replaced by vm_to_cluster.rb?

aespinosa:
ipxe.vdi can't get loaded from a second checkout of chef-bach because virtualbox will complain of named sdisk collisions.  This will cause the second VM cluster to not boot into iPXE

aespinosa:
In my second cluster i'm getting this:


```
Snapshot taken. UUID: c92816ed-d630-4e2a-900e-cfccfb9e4ed7
+ vboxmanage showvminfo allan-foo-bcpc-vm1
+ grep -q '^State:.*running'
+ VBoxManage startvm allan-foo-bcpc-vm1 --type headless
Waiting for VM "allan-foo-bcpc-vm1" to power on...
VBoxManage: error: Nonexistent host networking interface, name '' (VERR_INTERNAL_ERROR)
VBoxManage: error: Details: code NS_ERROR_FAILURE (0x80004005), component ConsoleWrap, interface IConsole
```

Seems that the shell scripts (probably vbox_create.sh) didn't pickup the new host adapters (vboxnet4, 5 and 6) in my second cluster.  IIRC this was working before.

cbaenziger:
On rebasing I just hit the noneexistent host networking error too now. Investigating...

cbaenziger:
This again builds me a cluster; I am building a single cluster at this time -- not trying multi-clusters yet.
* I changed `vbox_create.sh` to bail out if `$VBN#` can not be set @aespinosa; I hit that once on my single-cluster testing but can not seem to reproduce. Are you able to reproduce reliably? 
* I pushed a fix for the ZabbixAPI issue we have been seeing. I do not see an issue for that filed? @pu239ppy 
* As mentioned in the issue a fix for #981 is now included -- but I do not have an OS X box to test with so will ensure it does not break an Ubuntu hypervisor for now.

vm_to_cluster.sh
cbaenziger: I think this is a legacy script which got added back? Will drop again...

cbaenziger:
@aespinosa I have not `Dismiss review`'d your review but can you confirm if I addressed your concerns or if the VirtualBox networking issue is reproducable for you?

cbaenziger:
One thing which concerns me is I do tend to have gem compilation issues in my tests; they're different gems often and while I'm in `bach_repository` but thanks to idempotency I just kick `./tests/automated_install.sh` again and all runs through fine. Are other folks seeing this? (I suspect this to be independent of my code but wonder why it is happening.)

cbaenziger:
Ugh after rebase it seems master broke our rspec tests for cookbooks, so trying to fix a circular dependency between `bcpc` and `bach_repository` now. I would like to show that the rspec's still pass but this may be something to take out of scope of this commit.. Investigating but happy to be merged in the meantime.

aespinosa:
My first cluster builds fine.  Since this is being rescope more of fixing master instead of being multi-cluster, I think we can now pull these changes in.

PR #794, opened 2017-02-28
Add nscd for host lookup caching

http-418:


http-418:
NSCD for host lookups was requested internally. I see no reason not to add it to the default configuration.

cookbooks/bcpc/recipes/nscd.rb
```diff
@@ -0,0 +1,55 @@
...
+  nscd['hosts']['enable_cache'] = 'yes'
+
+  # Run the cache daemon as 'nscd'
+  nscd['server_user'] = 'nscd'
+end
+
+# Create an nscd user and home directory if none exist.
+group node['nscd']['server_user'] do
+  action :create
+  not_if "getent group #{node['nscd']['server_user']}"
```
vt0r: Quick question: Don't the `user` and `group` resources already perform the equivalent check?

http-418: There's no `:create_if_missing` on `user` and `group`.  If the entity already exists, Chef may attempt to modify it.   This is a fine behavior for passwd files, but may not work so well for directory users.

cookbooks/bcpc/recipes/nscd.rb
```diff
@@ -0,0 +1,55 @@
...
+end
+
+user node['nscd']['server_user'] do
```
cbaenziger: This should be a `system` user should not it?

http-418: The name of the attribute is `server_user`
https://github.com/chef-cookbooks/nscd/blob/v1.0.1/attributes/default.rb#L28

cbaenziger: I don't think that changes the behavior of the [`user` Chef resource](https://docs.chef.io/resource_user.html); we do this [elsewhere][(https://github.com/bloomberg/chef-bach/blob/a29c0a271069bb4179a1a20d58e96ec3f9cfcd99/cookbooks/bcpc-hadoop/recipes/datanode.rb#L87).

Issue #679, opened 2016-09-26
Enable email action in Oozie


bijugs:
In order for users to send email from Oozie jobs, Oozie site.xml need to be updated with ["smtp" server](https://oozie.apache.org/docs/3.3.0/DG_EmailActionExtension.html) and other information. This is an enhancement req to add these properties to Oozie so that email can be send out from Oozie jobs.

bijugs:
Fixed in PR #799.

PR #799, opened 2017-03-03
Changes to Oozie site xml to enable email action

bijugs:
This is for enhancement req in #679.

**Note:** Not tested by creating a new cluster.


cookbooks/bcpc-hadoop/templates/default/ooz_oozie-site.xml.erb
```diff
@@ -384,4 +384,21 @@
...
 
+  <% if node['bcpc']['hadoop']['oozie']['smtp_host'] != nil then %> 
+
+  <property>
+    <name>oozie.email.smtp.host</name>
+    <value><%= node['bcpc']['hadoop']['oozie']['smtp_host'] %></value>
+  </property>
+
+  <property>
+    <name>oozie.email.from.address</name>
```
cbaenziger: Is there anyway we can set this to something akin to `${wf:user()}`?

bijugs: AFAIK... currently the from e-mail id can be set to only a constant value.

bijugs:
anything else required here?

bijugs:
This will be a good feature to have for users. Anything else needs to be done here?

bijugs:
Thanks @vt0r .

PR #818, opened 2017-03-27
make sure hbase.master.wait.on.regionservers.mintostart only appears in HBase Master configuration

pu239ppy:
Ensure that region servers do not restart when hbase.master.wait.on.regionservers.mintostart  changes


cookbooks/bcpc-hadoop/recipes/hbase_config.rb
```diff
@@ -73,6 +71,12 @@
   'dfs.client.read.shortcircuit' => node["bcpc"]["hadoop"]["hbase"]["shortcircuit"]["read"].to_s
 }
 
+# this configuration parameter only belongs in master
```
cbaenziger: I think this is really valuable! However, going forward, can we follow the same break where we separate configs in data structures instead of ad hoc conditionals? We kind of do this for [HDFS](https://github.com/bloomberg/chef-bach/blob/69fd59788be976e7aedbe1b8d6b1e372ce2f833d/cookbooks/bcpc-hadoop/recipes/hdfs_site.rb#L58-L82) for example.

PR #821, opened 2017-04-03
Requested fix for ruby-based C-A-R + unrelated cosmetic work

http-418:
This PR contains two important fixes:
- Warn the user and bail if cluster.txt is missing critical roles on head nodes
- Add `liblzo2` dependency for Spark 2.x to be able to read gzip files

Additionally there are minor, unrelated fixes:
- Make the yarn `min.user.id` parameterized to make VM clusters more pleasant to use with service principals
- Make sure ignore_failure is disabled on CentOS and Ubuntu 16.04 in `bcpc::ssh`
- Add copyright/license header to `bach_spark::default`

### DIFF
--- a/cookbooks/bach_spark/recipes/default.rb
+++ b/cookbooks/bach_spark/recipes/default.rb
@@ -37,11 +37,16 @@ link '/usr/spark/current' do
   to spark_bin_dir
 end
 
+# Spark 2.x expects to dynamically link LZO.
+package 'liblzo2-2' do
+  action :upgrade
+end
+
 # install fortran libs needed by some jobs
 package 'libatlas3gf-base' do
-  action :install
+  action :upgrade
 end
 
 package 'libopenblas-base' do
-  action :install
+  action :upgrade
 end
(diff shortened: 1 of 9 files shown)


cookbooks/bach_spark/recipes/default.rb
```diff
@@ -1,3 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# This recipe installs spark to /usr/spark/ using a package
```
cbaenziger: Not just `/usr/spark` but to `node[:spark][:bin][:dir]` right?

cbaenziger:
Being this uses `all_hadoop_head_nodes` does this also protect Kafka clusters too?

cookbooks/bach_spark/recipes/default.rb
http-418: It's hardcoded in the package, so no.

http-418:
The role is called BCPC-Hadoop-Head on both Hadoop and Kafka clusters.

cookbooks/bach_spark/recipes/default.rb
cbaenziger: Ouch!

PR #822, opened 2017-04-04
Add a test for parse_cluster_txt

cbaenziger:
This adds a unit test for the `cluster_data` library. More to come but want to get a small PR outstanding...

### DIFF
--- a/cluster_assign_roles.rb
+++ b/cluster_assign_roles.rb
@@ -36,12 +36,12 @@ class ClusterAssignRoles
   #
   def all_hadoop_head_nodes
     # All head nodes should have this specific role.
-    confirmed_head_nodes = parse_cluster_txt.select do |nn|
+    confirmed_head_nodes = parse_cluster_txt(cluster_txt).select do |nn|
       nn[:runlist].include?('role[BCPC-Hadoop-Head]')
     end
 
     # Any nodes that have something matching "Head"
-    possible_head_nodes = parse_cluster_txt.select do |nn|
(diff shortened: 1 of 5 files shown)


http-418:
Why does this change C-A-R?

cbaenziger:
Because of CAR's use of [cluster_data](https://github.com/cbaenziger/chef-bach/blob/0c4c92e53351f29d7c4ac0c3e66b9bf4d28deb49/cluster_assign_roles.rb#L27)

cbaenziger:
Fun rebaising this just pull all of my changes out without a conflict somehow. Finding out what's up...

cbaenziger:
Okay now this is happy

http-418:
I still don't understand why C-A-R needed to change.  What is the point of having a method called "parse_cluster_txt" take an argument that is always the same?  Shouldn't that be a default value?

cbaenziger:
@http-418 How can you test this function without being able to handle an input? Also for the sake of being able to handle a cluster.txt for a Hadoop or Kafka cluster, this aides in being able to select a desired cluster type by walking the various cluster.txt's we have in the stub-environments directory.

cbaenziger:
As to why `cluster_assign_roles.rb` changes, can you explain what the challenge is? The function's prototype changed and C-A-R is a consumer which was updated, there is no other change to C-A-R?

http-418:
OK.  Add a default argument -- the signature can be unchanged as far as C-A-R is concerned, and your new test has an easy way to inject alternate values.

aespinosa:
already included in #788

PR #823, opened 2017-04-05
Use an execute resource to install chef-rewind on bootstrap

http-418:
This fixes an issue in internet-disconnected environments that is not encountered in VM clusters.


http-418:
Since this PR came from my 'master' branch, it now also includes a fix to `bcpc::networking` for hosts with a single, non-bonded NIC.  I don't think we had ever used this configuration before.  

Since VM clusters are multi-homed, they didn't hit the bug, despite having no bonded NICs.

PR #828, opened 2017-04-18
Tune HBase to balance regions for a table

amithkanand:
This PR tunes `HBase Stochastic Load Balancer` to balance regions for a table and data locality for regions.


cookbooks/bcpc-hadoop/attributes/hbase.rb
```diff
@@ -86,7 +86,8 @@
   site_xml['hbase.ipc.server.callqueue.read.ratio'] = node["bcpc"]["hadoop"]["hbase"]["ipc"]["server_call_queue_read_ratio"]
   site_xml['hbase.ipc.server.callqueue.scan.ratio'] = node["bcpc"]["hadoop"]["hbase"]["ipc"]["server_call_queue_scan_ratio"]
   site_xml['hbase.replication'] = 'true'
-  site_xml['hbase.master.loadbalance.bytable'] = true
```
vt0r: Is `false` the default value, or should we explicitly set this to `false`?

PR #833, opened 2017-04-22
Remove Windows, Non-Vagrant and Old Shell Script Code

cbaenziger:
This PR proposes:
* Removing the long not maintained Windows
* Removing the non-Vagrant install code path (long not maintained)
* Removes a number of legacy BCPC files which no-longer have use:
  * `vbox_save`
  * `vbox_restore`
  * `ssh-config`
* Tries to update `vbox_update.sh`
* This also unifies the code locality for where we create `cluster.tgz` and do the `rsync` to the Bootstrap host. (This allows #788 to update `cluster.txt` based on VirtualBox)


cbaenziger:
This code was built and tested up to PXE'ing VMs.

PR #834, opened 2017-04-22
Start checking if machines booted

cbaenziger:
This PR provides insight into if a VM has booted the installer to let one quickly know if their VuirtualBox install is failing to PXE boot their VMs.

One will see a new output in the `wait-for-hosts` run e.g:
```
Booted hosts from apache log 2/3
  bcpc-vm1: stage-2
  bcpc-vm3: stage-1
```
_here one could potentially suspect that vm2 isn't properly booting and the stage is simply which file the machine last grabbed per the logs (in order of the boot process)_

### DIFF
--- a/wait_for_hosts.rb
+++ b/wait_for_hosts.rb
@@ -36,28 +36,64 @@ def parsed_apache_log
 
   match_data.map do |md|
     {
-     hostname: md[1],
-     identd: md[2],
-     user: md[3],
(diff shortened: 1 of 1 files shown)


http-418:
PXE support is being removed from chef-bach, so this PR is no longer useful.  (wait-for-hosts.rb needs to be deleted)

PR #836, opened 2017-04-27
Kerberos Enhancements

amithkanand:
Adding this PR to do few things:

1. Restructure`Kerberos` data structure and make it readable and easy to manage
2. Match `Kerberos` data structure with wrapper cookbook to avoid any mismatch between wrapper and upstream. Flume is one example that was missing from wrapper.
3. Add correct `Group` and grant permissions while creating `Keytabs`


cbaenziger:
This unfortunately breaks testing for us. Having different principals for each service means we have verification we have not typoed something in our configs (e.g. using JN creds to start a DN, etc.). Collapsing to only two principals for all of HDFS and YARN means we no longer have our VM test clusters verifying we have not typoed anything.

aespinosa:
this also broke rolling PXE upgrades in a VM cluster since the old namenode and journalnode still refers to the old principals to authenticate .

PR #837, opened 2017-04-28
Better kerberos deployment

amithkanand:
Creating this PR to:

- Remove merging of `SPNEGO` principal into service keytab
- Create `SPENGO` keytab on each host (so far we were relying on merged keytab for `spnego` principal)
- Configure `Hadoop` services to read `SPNEGO` keytab instead of merged keytab
- Fix group memebership for all `Hadoop` service accounts

Breaking:

- All service accounts must belong to `Hadoop` group
- Service account from KDC are being used on all the clusters


aespinosa:
This got clobbered in #750

PR #838, opened 2017-05-02
Fixes for recent merge of Chef 12.x

http-418:
- Unified gem handling

Chef 12.x includes updated gem_package and chef_gem classes that expect to operate based on querying rubygems, whether or not `options` is specified.   This can cause failures in an airgapped environment (depending on `Chef::Config`) or version mis-matches even with an internet connection (available gems on bootstrap don't match rubygems.org metadata)

This commit adds a single Chef definition to handle all of the workarounds and peculiarities of installing gems from a bootstrap server.

On Chef 13.x we can remove this definition and use the native resource again, but for now, this at least sends all gem installations through a common code-path.

- Fixed cookbook versions

During development, Clay used version 0.2.0 for all his cookbooks, presumably to upload to an existing cluster.  Internally, we use 0.1.0 to represent an un-versioned cookbook.  This commit sets the versions back to 0.1.0.

- Reload xinetd instead of restarting.

The init script for xinetd restart is unreliable on Ubuntu 14.04, and refreshed configurations do not always get loaded.  This commit uses the `:reload` action instead.

- Fix warning messages during cookbook upload.

We were missing maintainer_email fields on some cookbooks, and knife 12.x complains.  This adds a maintainer_email to every cookbook.


cbaenziger:
The change to 0.2.0 is intentional as our dependent Bloomberg cookbooks were also stepped to 0.2.0.

http-418:
The versions of cookbooks outside chef-bach are irrelevant, because they are managed by Berkshelf as git revisions during the release process.

Why were versions inside chef-bach changed?

cookbooks/bcpc/recipes/graphite_handler.rb
```diff
@@ -5,20 +5,23 @@
 # This recipe should be the first recipe in the run list for all the
 # nodes to collect chef-client stats and store into Graphite
 #
+simple_graphite_version = '>=2.1'
```
cbaenziger: Why pin a version here and not in the Gemfile?

http-418: The version was already pinned there.  I just extracted it to a variable because I used it twice.

Many versions are pinned in the Chef code, and that was not the problem I am trying to fix.

cbaenziger: Ah I missed it was pinned in the recipe before; thanks

cbaenziger:
The intention is for VM clusters we would have a handle to write an upgrade procedure but that said I meant to do that before the PR was submitted so I suppose it can go away. We will need to think on the upgrade process and I suspect a gist or something a bit more heavy-duty will be in order. This is a rather large core BCPC change so going from 0.1.0 to 0.2.0 seemed appropriate as well. I am not wedded to it though.

http-418:
Our release's major version will be incremented to reflect the major change.

0.1.0 is explicitly not a version.  It should never change because it is a magic value indicating _no version_.  

Chef does not allow us to indicate that something is a snapshot.  A three digit dotted version is mandatory.  0.1.0 is the best we could do to indicate "This is not a versioned release"

cbaenziger:
Thanks for the fix ups Andrew!

PR #840, opened 2017-05-02
Remove duplicate property from hbase_config

amithkanand:


cookbooks/bcpc-hadoop/recipes/hbase_config.rb
```diff
@@ -60,7 +60,6 @@
   'hbase.zookeeper.quorum' =>
     node[:bcpc][:hadoop][:zookeeper][:servers].map{ |s| float_host(s[:hostname])}.join(","),
   'hbase.master.dns.nameserver' => dns_server,
```
cbaenziger: Thanks to HBASE-12954, we should no longer need this at all

PR #849, opened 2017-05-12
Deterministic heap for namenode

pu239ppy:
The current calculation was setup in such a way that NameNode heap would either be 16GB or some fraction of system memory, whichever is the minimum.  This was designed so that we are able to support real systems and along side this VM cluster.  This is not deterministic so lets just set to Xmx to 1GB and NewSize to 128m which is what is recommended for 1,000,000 files per http://docs.hortonworks.com/HDPDocuments/HDP2/HDP-2.3.6/bk_installing_manually_book/content/ref-80953924-1cbf-4655-9953-1e744290a6c3.1.html, and is probably too much for a toy VM cluster anyway.

BONUS: All the confusing logic is removed (at least for the NameNode)
BREAKING: This will restart a NameNode since we are making configuration chnages

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/default.rb
+++ b/cookbooks/bcpc-hadoop/attributes/default.rb
@@ -72,7 +72,8 @@ default["bcpc"]["hadoop"]["namenode"]["gc_opts"] =
 
 default["bcpc"]["hadoop"]["mapreduce"]["framework"]["name"] = "yarn"
 default["bcpc"]["hadoop"]["namenode"]["handler"]["count"] = 100
-default["bcpc"]["hadoop"]["namenode"]["xmx"]["max_size"] = 16384
+default["bcpc"]["hadoop"]["namenode"]["xmx"]["max_size"] = 1024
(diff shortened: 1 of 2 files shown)


cbaenziger:
I think this was all a misguided `.min`, instead of a `.max`. There's further the same issue with the Datanode above these lines which also ignores the override and uses `.min` instead of `.max`. Really these should use the dynamic tunings unless an override is provided was the original thought.

pu239ppy:
I have not done any research for the datanode so I was not prepared to make any changes, especially since it  currently works.  For the NameNodes, the goal was to have a basic value and one that we can override, being able to support 1,000,000 files on a VM node seemed reasonable.

pu239ppy:
I added a dynamic calculation option so instead of `min/max` games one just needs to either set a value `default["bcpc"]["hadoop"]["namenode"]["xmx"]["max_size"]` and `default["bcpc"]["hadoop"]["namenode"]["xmn"]["max_size"]` or set them to nil for dynamic calculation to happen.

PR #856, opened 2017-05-22
fix yarn log filename, correct a duplicate property

pu239ppy:
Filename should be determined only once
yarn.log.dir is duplicated, should probably be yarn.log.file

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/yarn.rb
+++ b/cookbooks/bcpc-hadoop/attributes/yarn.rb
@@ -18,8 +18,16 @@ yarn_log_dir = '/var/log/hadoop-yarn'
 yarn_pid_dir = '/var/run/hadoop-yarn'
 yarn_user = 'yarn'
 yarn_conf_dir = '/etc/hadoop/conf'
-yarn_logfile = 'yarn-$(hostname).log'
 yarn_policyfile = 'hadoop-policy.xml'
+yarn_logfile = 
+  if node.run_list.expand(node.chef_environment)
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc-hadoop/attributes/yarn.rb
```diff
@@ -18,8 +18,16 @@
 yarn_pid_dir = '/var/run/hadoop-yarn'
 yarn_user = 'yarn'
 yarn_conf_dir = '/etc/hadoop/conf'
-yarn_logfile = 'yarn-$(hostname).log'
 yarn_policyfile = 'hadoop-policy.xml'
+yarn_logfile = 
+  if node.run_list.expand(node.chef_environment)
+         .recipes.include?('bcpc-hadoop::datanode')
+      'yarn-yarn-nodemanager-$(hostname).log'
```
amithkanand: I am not sure if we need this. We don't specify log file names for any other service.

cbaenziger:
We should be using these file names to upload with Flume; e.g. in [`datanode.rb`](https://github.com/bloomberg/chef-bach/blob/6ea6751e17f959b1f605e61b523787f27b5c477b/cookbooks/bcpc-hadoop/recipes/datanode.rb#L9) we don't yet upload the NM log but should.

pu239ppy:
It seems as we really do not need it, should we just have attributes we can use to drive flume with and leave yarn configuration alone?

cbaenziger:
@pu239ppy But wouldn't we want to make sure the two use the same values?

PR #861, opened 2017-05-25
Fixes to make master work in a disconnected environment

http-418:
Fixes to allow master to work without internet access:
- Force ruby-augeas onto the uniform bcpc_chef_gem definition
- Use the --deployment flag for bundler install when applicable
- Add `poise` to the recipe for install_stub in C-A-R

Miscellaneous:
- Update java 8 to b131, which is actually available on Oracle internet mirrors
- Add a chef-bach version string to the message of the day


pu239ppy:
Resubmitted in 875

PR #865, opened 2017-06-01
Move to using Java cookbook for source of Java

cbaenziger:
This uses the Java cookbook for our source URL and version for Java and corrects the current cookie setting failure. (It may be a bit overzealous in setting the cookies but works.)


vt0r:
cookiemonster

http-418:
This sets up an ugly loop -- the java attributes will be overridden in bcpc/bcpc-hadoop to point to the bootstrap host itself.   If bach_repository is run on an internet-connected bootstrap, but something is wrong with bins, the chef run will fail.

cbaenziger:
@http-418 Indeed, if a bootstrap has any `bcpc-hadoop`-ism in its run_list it would have a loop here; we could move the attribute into `recipe[bcpc_hadoop:configs]` to prevent that. However, I do not see bootstraps running code out of `bcpc-hadoop`?

PR #867, opened 2017-06-01
Revert "Move to using Java cookbook for source of Java"

http-418:
Reverts bloomberg/chef-bach#865


cbaenziger:
Any reason to revert? I wasn't able to build master without this.

PR #868, opened 2017-06-02
Automated install running screen again

cbaenziger:
This adds starting screen in `automated_install.sh` again. We launch to `wait-for-hosts.sh` but still allow one to easily catch the VM's booting without having to race the VM's. If one does not have `screen` installed it falls back to only `wait-for-hosts.sh`. I needed to get to the VM's quickly to see the step in the installer on which they failed -- screen launching at boot helped me a lot with this.

### DIFF
--- a/tests/automated_install.sh
+++ b/tests/automated_install.sh
@@ -88,6 +88,9 @@ for vm in ${VM_LIST[*]} bcpc-bootstrap; do
   vboxmanage showvminfo $vm | grep -q '^State:.*running' || VBoxManage startvm $vm --type headless
 done
 
+WAIT_FOR_HOST_CMD="vagrant ssh -c 'cd chef-bcpc; source proxy_setup.sh; "\
+"./wait-for-hosts.sh \"${VM_LIST[*]}\"'"
+KILL_SCREEN_CMD='sleep 5 && kill $(ps -o ppid=$$)'
 if hash screen; then
   if [ "$(uname)" == "Darwin" ]; then
     brew install coreutils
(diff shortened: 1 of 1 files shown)


cbaenziger:
Oof; forgot to add the `kill screen` back in to make sure it exits out happily once all VM's start...

tests/automated_install.sh
```diff
@@ -97,23 +100,28 @@ if hash screen; then
   fi
 
   # Create a new screenrc with our VM list in it
+  SCREENRC=../../cluster/screenrc
   cp ./screenrc ../../cluster/screenrc
```
vt0r: Probably want to use that shiny new variable right here :)

tests/automated_install.sh
```diff
@@ -97,23 +100,28 @@ if hash screen; then
   fi
 
   # Create a new screenrc with our VM list in it
+  SCREENRC=../../cluster/screenrc
```
vt0r: Can we use an absolute path here?

tests/automated_install.sh
```diff
@@ -97,23 +100,28 @@ if hash screen; then
...
+         "./tests/virtualbox_serial_console.sh $vm" >> $SCREENRC
     ((ii++))
   done
+  echo "screen -t \"Wait For Hosts\" 0 $WAIT_FOR_HOST_CMD; $KILL_SCREEN_CMD" >> $SCREENRC
   popd > /dev/null
-  echo "Enter this command to view VM serial consoles:"
-  echo "  screen -S 'BACH serial consoles' `readlink -f ../cluster/screenrc`"
-  echo
+
+  screen -S 'BACH VM Install Progress' -c `readlink -f ../cluster/screenrc`
```
vt0r: Also probably want to use the variable here, assuming an absolute path.

cbaenziger:
In #1242 serial support for VM clusters was slid in; closing this as needless now.

PR #869, opened 2017-06-02
Support getting DNS for 16.04 Network Manager managed hosts

cbaenziger:
On my 16.04 laptop, I have no `/etc/dhcp/dhclient*.lease` files. Further, out of the box, my machine refers to itself as `127.0.1.1` which fouls the VM's booting in a non-obvious "can not find mirror" issue. Here I add Network Manager support for DNS and check for a loopback network range for the DNS server.

### DIFF
--- a/tests/edit_environment.rb
+++ b/tests/edit_environment.rb
@@ -19,32 +19,64 @@
 
 require 'json'
 require 'uri'
+require 'open3'
 
 def get_name_server
   ns = File.readlines('/etc/resolv.conf').select do |ll|
     ll.include?('nameserver')
   end.first.chomp.gsub(/^\s*nameserver\s*/,'')
 
-  if ns == '127.0.0.1'
+  # see if we have a DHCP leases file to parse
+  if !ns || ns.start_with? '127.'
     # Find the newest DHCP client lease file
     latest_lease = Dir.glob('/var/lib/dhcp/dhclient*.lease*').sort do |a, b|
       File.mtime(a) <=> File.mtime(b)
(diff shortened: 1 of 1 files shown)


tests/edit_environment.rb
```diff
@@ -19,32 +19,64 @@
...
+require 'open3'
 
 def get_name_server
   ns = File.readlines('/etc/resolv.conf').select do |ll|
     ll.include?('nameserver')
   end.first.chomp.gsub(/^\s*nameserver\s*/,'')
 
-  if ns == '127.0.0.1'
+  # see if we have a DHCP leases file to parse
+  if !ns or ns.start_with? '127.'
```
vt0r: Should we be using `||` rather than `or` here to prevent possible precedence issues?

tests/edit_environment.rb
```diff
@@ -19,32 +19,64 @@
...
+      # If we found a DNS option, extract a name server.
+      ns = dns_option.chomp\
+        .gsub(/^\s*option domain-name-servers\s*([^,]*),.*/,'\1') if dns_option
     end
-  else
-    ns
   end
+  
+  # see if we are a network manager managed machine -- for Ubuntu >= 15
+  if !ns or ns.start_with? '127.'
```
vt0r: Also `or` --> `||` ?

cbaenziger:
Thanks! All three `!ns or ns.start_with? '127.'` changed to `!ns || ns.start_with? '127.'`, in case of future modification.

Issue #874, opened 2017-06-08
bundler --install fails with permission denied


pu239ppy:
while running cookbooks/bach_repository/recipes/gems.rb https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bach_repository/recipes/gems.rb#L36 will fail with permission denied.  This is due to /home/vagrant/.bundle being owned by root and not having sufficient privileges to modify the cache.  Temporary workaround is to run this prior

````
diff --git a/cookbooks/bach_repository/recipes/gems.rb b/cookbooks/bach_repository/recipes/gems.rb
index fbce0a5..af15b62 100644
--- a/cookbooks/bach_repository/recipes/gems.rb
+++ b/cookbooks/bach_repository/recipes/gems.rb
@@ -57,6 +57,14 @@ ruby_block 'determine-bundler-command' do
   end
 end

+execute 'permissions hack' do
+  cwd '/home/vagrant'
+  command 'chown -R vagrant:vagrant .bundle || /bin/true '
+end
+
+
+
+
 execute 'bundler install' do
   cwd node['bach']['repository']['repo_directory']
   command lazy { node.run_state[:bcpc_bootstrap_bundler_command] }````

**Comment by cbaenziger** [comment]
It looks like despite both bundler commands are being run as `vagrant` in [gem.rb](https://github.com/bloomberg/chef-bach/blob/c45e32bab089ced1dfb5778d2c36afe7914023f7/cookbooks/bach_repository/recipes/gems.rb#L36-L60) the first bundler command is using `sudo` internally. We are ending up with the following from `auditd` logs:
```
[...]
type=SYSCALL msg=audit(1496899595.169:390): arch=c000003e syscall=87 success=yes exit=0 a0=7f1c041c3f40 a1=7f1c041c6610 a2=0 a3=7f1c0ab6b130 items=2 ppid=26831 pid=27111 auid=900 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=14 comm="bundle" exe="/opt/chef/embedded/bin/ruby" key="foo"
[...]

Issue #873, opened 2017-06-08
chef-vault cookbook missing ChefVaultCookbook::Provider


pu239ppy:
I found that this error happens when trying to build master

````
================================================================================
  Recipe Compile Error in /var/chef/cache/cookbooks/bcpc/recipes/cobbler.rb
  ================================================================================

  NameError
  ---------
  uninitialized constant ChefVaultCookbook::Provider

  Cookbook Trace:
  ---------------
    /var/chef/cache/cookbooks/bcpc/recipes/cobbler.rb:58:in `block in from_file'
    /var/chef/cache/cookbooks/bcpc/recipes/cobbler.rb:53:in `from_file'

  Relevant File Content:
  ----------------------
  /var/chef/cache/cookbooks/bcpc/recipes/cobbler.rb:

   51:  root_password_salted = root_password.crypt('$6$' + rand(36**8).to_s(36))
   52:
   53:  chef_vault_secret 'cobbler' do
   54:    #
   55:    # For some reason, we are compelled to specify a provider.
   56:    # This will probably break if we ever move to chef-vault cookbook 2.x
   57:    #
   58>>   provider ChefVaultCookbook::Provider::ChefVaultSecret
   59:
   60:    data_bag 'os'
   61:    raw_data('web-password' => web_password,
   62:             'root-password' => root_password,
   63:             'root-password-salted' => root_password_salted)
   64:    admins node[:fqdn]
   65:    search '*:*'
   66:    action :create_if_missing
   67:  end
````

I fixed up *cookbooks/bcpc/metadata.rb* to require a specific version of chef-vault cookbook as opposed to v3.x that gets pulled in 

````
-depends 'chef-vault'
+depends 'chef-vault', '=1.3.3'
````

Testing fix now

**Comment by vt0r** [comment]
What happens if the `provider` parameter just gets removed? Assuming the cookbook got upgraded, we should be able to just remove that line from the resource, right?

**Comment by http-418** [comment]
I don't know why the resource doesn't work without specifying a provider.  You can try removing the `provider` line, or updating the provider to match the class name in a current version of the cookbook.

**Comment by http-418** [comment]
I also particularly enjoy line 56:
```
   56:    # This will probably break if we ever move to chef-vault cookbook 2.x
```

cbaenziger:
Cross posting from #876 :
[...]

Issue #848, opened 2017-05-10
Chef run failure in build_bins.sh while building VM cluster


vt0r:
When running `tests/automated_install.sh`, the `build_bins.sh` script is executed, and though no error output is displayed during the execution of `build_bins.sh`, a stacktrace is printed (below) to `/var/chef/cache/chef-stacktrace.out`, which causes the overall bootstrap setup procedure to fail later on when this file is detected. Error message will follow shortly when I can reproduce it again.

The error can be bypassed until this issue is resolved using this terrible hack:
```diff
--- a/bootstrap_chef.sh
+++ b/bootstrap_chef.sh
@@ -85,6 +85,8 @@ fi
 $SSH_CMD "tar -xzf /chef-bcpc-host/cluster.tgz -C /home/vagrant" || (echo '### Vagrant ssh failed to deploy cluster environment returned $? ###'; e
 echo "Building bins"
 $SSH_CMD "cd $BCPC_DIR && sudo ./build_bins.sh"
+echo "HACK - removing stacktrace file, since the build_bins run succeeded."
+$SSH_CMD "sudo rm -f /var/chef/cache/chef-stacktrace.out"
 echo "Setting up chef server"
 $SSH_CMD "cd $BCPC_DIR && sudo ./setup_chef_server.sh ${CHEF_ENVIRONMENT}"
 echo "Setting up chef cookbooks"
```

cbaenziger:
For reference, while I see the following when my `build_bins.sh` run kicks off:
```[2017-05-11T05:39:56+00:00] INFO: HTTP Request Returned 404 Not Found: Object not found: chefzero://localhost:8889/nodes/bcpc-bootstrap.bcpc.example.com
```

I can not replicate having a `/var/chef/cache/chef-stacktrace.out` by the type my PXE install kicks off. Are you sure your `build_bins.sh` Chef run is where you are getting this stack trace?

pu239ppy:
Ignoring this blows up during C-A-R
 * mysql_database[pdns] action create

````    ============================================================================
====
    Error executing action `create` on resource 'mysql_database[pdns]'
    ============================================================================
====
````

vt0r:
@cbaenziger - Yes. I wrapped the following line in debug output (`sudo cat /var/chef/cache/chef-stacktrace.out`), and it doesn't exist before the script runs, but the stacktrace is present afterwards (L87):

```bash
[...]

Issue #847, opened 2017-05-10
Possible race condition in Keepalived/HAProxy/OpenBFDd


vt0r:
Getting the below error when trying to run `tests/automated_install.sh` to do a new VM cluster build. Logging into the VM after this error is encountered, you can ping the VIP and connect to MySQL on it using the vaulted root password, hence the title.

```
Recipe: bcpc::powerdns
  * chef_vault_secret[mysql-pdns] action nothing (skipped due to action :nothing)
  * mysql_database[pdns] action create
    
    ================================================================================
    Error executing action `create` on resource 'mysql_database[pdns]'
    ================================================================================
    
    Mysql2::Error
    -------------
    Can't connect to MySQL server on '10.0.100.5' (113)
    
    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/database/libraries/provider_database_mysql.rb:105:in `new'
    /var/chef/cache/cookbooks/database/libraries/provider_database_mysql.rb:105:in `test_client'
    /var/chef/cache/cookbooks/database/libraries/provider_database_mysql.rb:37:in `block in <class:Mysql>'
    
    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/bcpc/recipes/powerdns.rb
    
     90:   mysql_database node['bcpc']['pdns_dbname'] do
     91:     connection lazy { mysql_connection_info.call }
     92:     notifies :run, 'execute[install-pdns-schema]', :immediately
     93:   end
     94: 
    
    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/bcpc/recipes/powerdns.rb:90:in `from_file'
    
    mysql_database("pdns") do
      provider Chef::Provider::Database::Mysql
      action :create
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      database_name "pdns"
      declared_type :mysql_database
      cookbook_name "bcpc"
      recipe_name "powerdns"
      connection #<Chef::DelayedEvaluator:0x00000002e64cc0@/var/chef/cache/cookbooks/bcpc/recipes/powerdns.rb:91>
    end
    
    Platform:
    ---------
    x86_64-linux
    
Recipe: bcpc::mysql
  * service[mysql] action reload
    - reload service service[mysql]
Recipe: bcpc::keepalived
  * service[keepalived] action restart
    - restart service service[keepalived]

Running handlers:
  - GraphiteReporting
Running handlers complete
[...]

PR #876, opened 2017-06-10
Added hacks in order to move master build forward

pu239ppy:
Merged hacks/workarounds for
- #873 
- #874 
- #848 
- #847

While we investigate correct/elegant solutions we still need a way to build master without having to apply patches every time


cbaenziger:
Since we use a modern version of Chef (>12.9) we should move to using Chef-Vault's custom provider likely (which removed the ancient provider we were using): [chef-cookbooks/chef-vault #58 change](https://github.com/chef-cookbooks/chef-vault/pull/58/files#diff-fb6cda3ef49a27a1c9f4ecdcf032c520L18)

vt0r:
Closing in favor of #882

PR #881, opened 2017-06-12
cluster_assign_roles.rb never runs install_bootstrap

### LINKED ISSUE #880: cluster-assign-roles Test-Laptop Hadoop fails on bcpc-vm1
Multiple failures in HDFS and HBASE related recipes 
````
* ruby_block[acquire_lock_to_restart_hbase_master] action createacquire_restart_lock : unable to 
get_restart_lock_holder : unable to connect to ZooKeeper quorum f-bcpc-vm1.bcpc.example.com:2181
  * bash[format-zk-hdfs-ha] action runznode_exists : Unable to connect to zookeeper quorum f-bcpc-vm1.bcpc.example.com:2181
````

pu239ppy:
This is intended to fix #880 however we may need extra logic and fixes before this can be merged

 - `target_x_nodes` vs `all_hadoop_x_nodes` and also does
 - we probbaly should not run `install_bootstrap` every time `install_hadoop` is invoked

### DIFF
--- a/cluster_assign_roles.rb
+++ b/cluster_assign_roles.rb
@@ -222,6 +222,9 @@ class ClusterAssignRoles
     target_head_nodes = target_nodes & all_hadoop_head_nodes
     target_worker_nodes = target_nodes - all_hadoop_head_nodes
 
+    # first prepare commonly used services
+    install_bootstrap(all_hadoop_head_nodes)
+
     #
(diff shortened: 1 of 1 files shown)


pu239ppy:
Not needed as tests/automated_install.sh already calls it separately
https://github.com/bloomberg/chef-bach/blob/master/tests/automated_install.sh#L134

PR #882, opened 2017-06-19
Master build progression

vt0r:
As mentioned in #876, we have several patches we are currently applying to make master builds complete. These patches are mostly hacks to work around things like race conditions and the like, however, we do require these patches to make a full build succeed on current master. This PR is a mixture of previously mentioned hacks, plus a few new ones to resolve issues in the master build and allow it to complete.

A short list of the issues resolved or worked around using this PR:
* #847
* #848
* #873
* #874

This PR may still be WIP, so I'll add the tag and leave it until I'm sure it's complete:
I encountered a missing role error regarding the YARN ATS role, however I'm running another test build to determine whether clearing and replacing `../cluster`'s contents will resolve it.

Supersedes #876


stub-environment/environments/Test-Laptop.json
```diff
@@ -71,7 +71,8 @@
       "cache_path": "/var/chef/cache",
       "config": {
          "verify_api_cert": false,
-         "ssl_verify_mode": "verify_none"
+         "ssl_verify_mode": "verify_none",
+         "rubygems_url": "http://10.0.100.3/"
```
http-418: This shouldn't be necessary if everything is using the bcpc_chef_gem definition

cookbooks/bach_repository/templates/default/chef-install.sh.erb
```diff
@@ -19,3 +19,13 @@ else
     echo 'Checksum did not match for <%= @download_url %>!'
     exit 1
 fi
+
+# Remove rubygems and install only our gemserver
+/opt/chef/embedded/bin/gem sources --add http://<%= node['bcpc']['bootstrap']['server'] %>/ || (echo "Failed to add local Gem repo")
+/opt/chef/embedded/bin/gem sources --remove http://rubygems.org/ || true
+/opt/chef/embedded/bin/gem sources --remove https://rubygems.org/ || true
+/opt/chef/embedded/bin/gem sources | grep -q "^http://<%= node['bcpc']['bootstrap']['server'] %>/$" || (echo "Failed to add bootstrap to gem sources")
```
http-418: This shouldn't be necessary if everything is using the bcpc_chef_gem definition

http-418:
Turns out the very unnecessary-looking global config changes are, in fact, necessary.  Chef 12.8 and above allows cookbooks to mandate `chef_gem` installations _before any cookbook code executes_.  As a result, it is no longer possible to correct any configuration after the chef run has started.

PR #884, opened 2017-06-27
Untangle bach_repository from bcpc

aespinosa:
Remove the circular dependency that the `solver :ruby` in Berkshelf was
borking about.


http-418:
This does not fix the entanglement problem, and additionally will not resolve our berkshelf issue.

On entanglement, we would need to split all bootstrap-related material from bcpc, so that bcpc could be depended upon by the bootstrap-related material.

On the berkshelf issue, we don't have a root cause yet.

PR #888, opened 2017-06-29
Yarn Enhancements

amithkanand:


aespinosa:
This is missing from 3.0 s.

3.0 cluster will have broken oozie hiveaction workflows because it connects to the timeline server  amd it returns the wrong principal (fromthe yarn.service.keytab instead of spnego.service.keytab)

PR #890, opened 2017-07-03
Allow disks to have arbitrary names in bcpc-hadoop::disks

http-418:
This PR follows on the install-time work in PR #861 and PR #875 to format disks correctly regardless of device enumeration order.

Instead of blindly formating sdb, sdc, sdd, sde in order, unformatted disks are detected by attempting to open() with O_EXCL.  Block devices that are already in use by the kernel cannot be opened with that flag, and fail.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/disks.rb
+++ b/cookbooks/bcpc-hadoop/recipes/disks.rb
@@ -47,9 +47,9 @@ ruby_block 'enumerate-disks' do
       # What disks will bcpc and bcpc-hadoop feel free to blank?
       # By default, all sd* devices (excluding sda) and all md* devices.
       #
-      # On our EFI-based VM builds, it's very important to the 32 MB
+      # On our EFI-based VM builds, it's very important to omit the 32 MB
       # image containing iPXE.  (It's relatively harmless to overwrite
-      # it, but it will cause graphite to fail when /disk/0 fills up.)
+      # the image, but it will cause graphite to fail when /disk/0 fills up.)
       #
       # We also reject any block device we are unable to open with O_EXCL,
       # because that means it is already in use by the kernel.
--- a/cookbooks/bcpc-hadoop/recipes/disks.rb
+++ b/cookbooks/bcpc-hadoop/recipes/disks.rb
@@ -47,17 +47,28 @@ ruby_block 'enumerate-disks' do
       # What disks will bcpc and bcpc-hadoop feel free to blank?
       # By default, all sd* devices (excluding sda) and all md* devices.
       #
-      # On our EFI-based VM builds, it's very important to omit sdb, as
-      # that is the 32 MB image containing iPXE.  (It's relatively
-      # harmless to overwrite it, but it will cause graphite to fail when
-      # /disk/0 fills up.)
+      # On our EFI-based VM builds, it's very important to the 32 MB
+      # image containing iPXE.  (It's relatively harmless to overwrite
+      # it, but it will cause graphite to fail when /disk/0 fills up.)
+      #
+      # We also reject any block device we are unable to open with O_EXCL,
+      # because that means it is already in use by the kernel.
       #
       all_drives = node[:block_device].keys.select do |dd|
-        dd =~ /sd[a-i]?[b-z]/ ||
-        dd =~ /md\d+/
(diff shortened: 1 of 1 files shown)


cookbooks/bcpc-hadoop/recipes/disks.rb
```diff
@@ -47,17 +47,28 @@
       # What disks will bcpc and bcpc-hadoop feel free to blank?
       # By default, all sd* devices (excluding sda) and all md* devices.
       #
-      # On our EFI-based VM builds, it's very important to omit sdb, as
-      # that is the 32 MB image containing iPXE.  (It's relatively
-      # harmless to overwrite it, but it will cause graphite to fail when
-      # /disk/0 fills up.)
+      # On our EFI-based VM builds, it's very important to the 32 MB
```
cbaenziger: I think you removed the subject of this sentence? (What is `it's` here?)

cookbooks/bcpc-hadoop/recipes/disks.rb
```diff
@@ -47,17 +47,28 @@
...
+      # image containing iPXE.  (It's relatively harmless to overwrite
+      # it, but it will cause graphite to fail when /disk/0 fills up.)
+      #
+      # We also reject any block device we are unable to open with O_EXCL,
+      # because that means it is already in use by the kernel.
       #
       all_drives = node[:block_device].keys.select do |dd|
-        dd =~ /sd[a-i]?[b-z]/ ||
-        dd =~ /md\d+/
+        dd =~ /sd[a-i]?[a-z]/ || dd =~ /md\d+/
```
cbaenziger: Why stop at `i` here?

http-418: The minor number in a device file is 8 bits, so the last possible device node, (135, 255), will be /dev/sdiv 

https://github.com/torvalds/linux/blob/v4.12/Documentation/admin-guide/devices.txt#L2067-L2072

cookbooks/bcpc-hadoop/recipes/disks.rb
http-418: That's a typo.  Will fix.

cookbooks/bcpc-hadoop/recipes/disks.rb
cbaenziger: Ah my mental regex parsing was faulty; indeed `[a-i]?` is zero or one `a-i`'s while the trailing `[a-z]` will cover `/dev/sd[a-z]` which I was failing to grok.

cookbooks/bcpc-hadoop/recipes/disks.rb
```diff
@@ -89,113 +100,127 @@
...
-              ee.run_action(:run)
-            end
-          end
-
-        end
+    unformatted_disks =
+      node.run_state[:bcpc_hadoop_disks][:unformatted_disks]
+
+    unformatted_disks.each_index do |i|
+      Chef::Resource::Directory.new("/disk/#{i}",
```
cbaenziger: If a new disk is added to a running system, then `unformatted_disks` will have one disk (e.g. `i` will be `0`) but if we had existing disks won't `/disk/0` already exist and this directory resource do nothing and the later `mount` action collide?

http-418: Yes.  This is not a complete scheme.  It solves one problem, but allows several others to fester.

Adding a new disk that enumerates in a new order will break the old code just as badly.

cbaenziger: But even a hot-swap disks will not get added back correctly, right?

http-418: I suppose not -- even if it is inserted at the same place in the order, it will not format correctly, because its number will be taken.

This is not a problem for Kafka hosts, running RAID-10.  I hadn't considered its implications on an un-RAIDed node.

http-418:
I'm going to revisit this next week.  The O_EXCL hack gets me closer to a grand, unified solution for creating software RAIDs and identifying unused disks in one sweep.

vt0r:
Based on the discussion, the fact that it looks fine to me (+ is tested) and the fact that even the mentioned cases missed by the logic don't put us in a worse position, I'll merge this as is.

PR #893, opened 2017-07-07
Add script to archive the bins/ directory

aespinosa:
Tarballs all the necessary files to be deployed on a Bootstrap node and
cluster that is due to be upgraded.


http-418:
it's not safe to leave the bins tarball inside /chef-bcpc-host, because its size will screw up future builds in the same hypervisor directory

the script should probably move the tarball to .. after it is created

aespinosa:
Good point.  Just added that line

PR #896, opened 2017-07-11
Fix additional Chef 12.x migration bugs, upgrade to Kafka 11

http-418:
This is the most recent set of changes I've stumbled across on my hardware cluster builds.

- Default to Ubuntu 14.04 + legacy BIOS to improve VM build experience
- Update to Kafka 0.11 from Apache.org
- Force restarts of the bootstrap xinetd on every chef run
- Remove make_databag.sh and friends
- Fix bach_repository::gems permissions
- Fix bad knife.rb using a Chef 12.x client


http-418:
This is blocked on fixing #890

cbaenziger:
Can pull request #890's code be pulled out of this to get it moving please?

cookbooks/bcpc/recipes/cobbler.rb
```diff
@@ -345,3 +350,13 @@
   action [:enable,:start]
   notifies :run, 'bash[cobbler-sync]', :delayed
 end
+
+#
+# After recurring problems with reloading the xinetd
+# configuration, we resort to a forced restart on every chef run.
```
cbaenziger: Do you have some concrete pointers which can be referenced so we know when this may be pulled out?

http-418:
This is partly tested.  It's running on my (hardware) lab cluster right now.  I would like to see a VM build complete, though.

cookbooks/bcpc/recipes/cobbler.rb
http-418: I still don't think this commit has fixed my problem.  xinetd fails to respond on the TFTP port when the VMs boot, and I don't know why.  Worse, it works again after another chef run.

ekund:
Setting oozie URL to the management vip doesn't work.  I believe you need to set it to the address of the node that actually runs oozie on it.  As it's not HA, it's just on one node.
Also, the url should be of the form: <url>:<port>/oozie so it won't work in any case.  Can you please fix?

PR #897, opened 2017-07-12
Upgrade to HDP 2.6.1.0-129

aespinosa:
Rehash of #858 for 2.6

Update: cluster builds

Smoke Tests:
- [x] hdfs
- [x] mapreduce
- [x] hbase
- [x] hive
- [x] oozie
- [x] spark

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/default.rb
+++ b/cookbooks/bcpc-hadoop/attributes/default.rb
@@ -6,7 +6,7 @@
 require 'pathname'
 
 default['bcpc']['hadoop'] = {}
-default['bcpc']['hadoop']['distribution']['release'] = '2.3.6.2-3'
+default['bcpc']['hadoop']['distribution']['release'] = '2.6.1.0-129'
 default['bcpc']['hadoop']['distribution']['active_release'] = node['bcpc']['hadoop']['distribution']['release']
 default['bcpc']['hadoop']['decommission']['hosts'] = []
 # disks to use for Hadoop activities (expected to be an environment or role set variable)
--- a/cookbooks/bcpc-hadoop/libraries/hive_metastore.rb
+++ b/cookbooks/bcpc-hadoop/libraries/hive_metastore.rb
@@ -15,8 +15,9 @@ module BcpcHadoop::HiveMetastore::Database
 
     attribute :root_password, kind_of: String
     attribute :hive_password, kind_of: String
+    attribute :schematool_path, kind_of: String
 
-    actions :create
+    actions :create, :init
   end
 
   class Provider < Chef::Provider
@@ -49,6 +50,24 @@ FLUSH PRIVILEGES;
           'WHERE SCHEMA_NAME = "metastore"\' | grep -q metastore'
       command.run_command
       command.exitstatus == 0
+    end
+
+    def action_init
+      return if schema_initialized?
+      converge_by 'initializing the schema to 1.2.0' do
+        execute 'initialize schema to 1.2.0' do
+          command "#{new_resource.schematool_path} -dbType mysql "\
+                  "-userName root -passWord #{new_resource.root_password} "\
+                  '-initSchemaTo 1.2.0'
+        end
+      end
+    end
+
+    def schema_initialized?
+      schema_info = Mixlib::ShellOut.new "#{new_resource.schematool_path} -dbType mysql -info"
+      schema_info.run_command
+      !(schema_info.stderr =~ /metastore.VERSION' doesn't exist/) &&
+        (schema_info.stdout =~ /Metastore schema version:\s+\d+\.\d+\.\d+/)
     end
   end
 end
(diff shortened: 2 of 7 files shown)


aespinosa:
Update: cluster builds

aespinosa:
Talked with @pu239ppy 

Hive schema needs to be updated from two state paths:

1. fresh cluster install
2. upgrade the schema from 0.14.0 to 1.2.1000 (?)

aespinosa:
Was able to fiddle with enough permissions in hdfs to create both internal and external hive tables.  Updated the wikipage [1] with it .

So our existing HDP 2.3.X is supposed to have hive 1.2.1. But for some reason metadatastore schema at 0.14.0 is still working.

[1] <https://github.com/bloomberg/chef-bach/wiki/Create-Hadoop-Cluster>

cookbooks/bcpc-hadoop/recipes/hive_hcatalog.rb
```diff
@@ -74,6 +74,9 @@
         FLUSH PRIVILEGES;
         USE metastore;
         SOURCE /usr/hdp/current/hive-metastore/scripts/metastore/upgrade/mysql/hive-schema-0.14.0.mysql.sql;
+        SOURCE /usr/hdp/current/hive-metastore/scripts/metastore/upgrade/mysql/upgrade-0.14.0-to-1.1.0.mysql.sql;
+        SOURCE /usr/hdp/current/hive-metastore/scripts/metastore/upgrade/mysql/upgrade-1.1.0-to-1.2.0.mysql.sql;
+        SOURCE /usr/hdp/current/hive-metastore/scripts/metastore/upgrade/mysql/upgrade-1.2.0-to-1.2.1000.mysql.sql;
```
aespinosa: Confirmed with schematool that these are indeed the sql scripts to be sourced:

```
schematool -upgradeSchema -dbType mysql -dryRun
2017-07-14 11:23:01,952 WARN  [main] conf.HiveConf: HiveConf of name hive.server2.logging.operation.verbose does not exist
Metastore connection URL:        jdbc:mysql:loadbalance://bcpc-vm1.bcpc.example.com:3306,bcpc-vm2.bcpc.example.com:3306/metastore?loadBalanceBlacklistTimeout=5000
Metastore Connection Driver :    com.mysql.jdbc.Driver
Metastore connection User:       hive
Starting upgrade metastore schema from version 0.14.0 to 1.2.1000
Upgrade script upgrade-0.14.0-to-1.1.0.mysql.sql
Upgrade script upgrade-1.1.0-to-1.2.0.mysql.sql
Upgrade script upgrade-1.2.0-to-1.2.1000.mysql.sql
schemaTool completed
```

http-418:
This looks OK as is, but there are a couple small things that could improve it.  

- We could get by with a lot less code by having the database and permissions handled by the normal mysql cookbook providers, instead of a bunch of mixlib::shellout to call to the mysql binary.  You could even call out to the mysql providers from inside your existing provider using poise inline resources.

- It would be nice to have the hive metastore credentials pre-created the same way as the other credentials, inside the mysql_data_bags recipe.

cookbooks/bcpc-hadoop/libraries/hive_metastore.rb
```diff
@@ -0,0 +1,96 @@
...
+      converge_by 'creating the metastore database' do
+        create_hive_database
+      end
+    end
+
+    def create_hive_database
+      privs = 'SELECT,INSERT,UPDATE,DELETE,LOCK TABLES,EXECUTE'
+      command = Mixlib::ShellOut.new "mysql -uroot -p#{new_resource.root_password}",
+        input: <<-eos
+CREATE DATABASE metastore;
```
vt0r: In addition, isn't there some MySQL provider already? I'll check the code and update this comment if I find one, since it would probably save you lots of work :)

cookbooks/bcpc-hadoop/recipes/hive_hcatalog.rb
```diff
@@ -62,34 +62,14 @@
   user 'hdfs'
 end
 
-ruby_block "hive-metastore-database-creation" do
```
vt0r: Thanks for moving this out to a library!

cookbooks/bcpc-hadoop/libraries/hive_metastore.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  class Provider < Chef::Provider
+    include Poise
+
+    provides :hive_metastore_database
+
+    def action_create
+      return if database_created?
+      converge_by 'creating the metastore database' do
```
vt0r: Thank you for supporting `why-run` by using `converge_by`!

cookbooks/bcpc-hadoop/attributes/default.rb
```diff
@@ -5,10 +5,10 @@
 #############################################
 require 'pathname'
 
-default['bcpc']['hadoop'] = {}
-default['bcpc']['hadoop']['distribution']['release'] = '2.3.6.2-3'
-default['bcpc']['hadoop']['distribution']['active_release'] = node['bcpc']['hadoop']['distribution']['release']
-default['bcpc']['hadoop']['decommission']['hosts'] = []
+default["bcpc"]["hadoop"] = {}
```
aespinosa: thanks for catching this.  i ran rubocop and foodcritic in my libraries but totatlly skipped the attributes i changed.

aespinosa:
Thanks @vt0r .  The style changes are now in place.

PR #903, opened 2017-07-19
Remove errant Gemfile.lock and update bach_repository Gemfile to match Chef-BACH versions

cbaenziger:
I have found on systems where the hypervisor ChefDK doesn't match the bootstrap ChefDK, one ends up with issues where bundler in `build_bins.sh` will throw an error about a Gem mismatch.


cbaenziger:
I've built a master cluster and run `kitchen verify` in `bach_repository`.

http-418:
In the long run it might be nice not to use the chef gems from chefdk, for this very reason.

PR #910, opened 2017-07-21
Correctly handle a fixnum FD from sysopen

http-418:
sysopen returns a fixnum, not an FD object, so we must create an FD object from the integer in order to close it successfully.

This has been breaking new VM builds


http-418:
I haven't tested this yet

PR #912, opened 2017-07-21
Additional disk fixes

http-418:
- Correctly derive UUID= line for mounts by UUID
- Wrap the fixnum FD with an IO object in order to close it

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/disks.rb
+++ b/cookbooks/bcpc-hadoop/recipes/disks.rb
@@ -155,7 +155,13 @@ ruby_block 'format-disks' do
         end
       end
 
-      mount_device = bcpc_uuid_for_device(dev_name) || dev_name
+      uuid_candidate = bcpc_uuid_for_device(dev_name)
+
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc-hadoop/recipes/disks.rb
```diff
@@ -155,7 +155,13 @@
         end
       end
 
-      mount_device = bcpc_uuid_for_device(dev_name) || dev_name
+      uuid_candidate = bcpc_uuid_for_device(dev_name)
+
+      mount_device = if uuid_candidate
+                       "UUID=#{uuid_candidate}"
```
aespinosa: `device_type` is probably the correct way to do it. https://docs.chef.io/resource_mount.html#properties

Just had this breaking in my build now.

```
     Chef::Exceptions::Mount
      -----------------------
      Device UUID=53cd9a71-5a7a-45e0-a9a5-cc19a696ee32 does not exist

      Cookbook Trace:
```

PR #917, opened 2017-07-26
additionally need to update kafka metadata

pu239ppy:


aespinosa:
the berksfile.lock probably needs an update as well

pu239ppy:
As far as bcpc_kafka entries are concerned they are already at 2.4.2

aespinosa:
confirmed it's there already https://github.com/pu239ppy/chef-bach/blob/e9ff5d61f470ecb719bd765d6e187efdb4df22ee/Berksfile.lock#L80

PR #920, opened 2017-07-28
Cleanup repxe

aespinosa:


repxe-host.sh
```diff
@@ -11,5 +11,5 @@ REPO_DIR="`dirname ${BASH_SOURCE[0]}`"
 cd $REPO_DIR
 bundle config --local PATH vendor/bundle
 bundle config --local DISABLE_SHARED_GEMS true
-bundle exec --keep-file-descriptors ./repxe_host.rb $*
+bundle exec --keep-file-descriptors $PWD/repxe_host.rb $*
```
http-418: The bundler configuration file contains relative paths.  What happens when you run bundle exec with an absolute path after configuring relative paths everywhere else?

aespinosa: The ruby bundled with the new chefdk has __FILE__ to absolute path and $PROGRAM_NAME to relative path (I may have swapped the two) in https://github.com/bloomberg/chef-bach/blob/master/repxe_host.rb#L449.


Bundler actually renders the GEM_PATH to absolute paths even though the configuration is relative:


```
# foo.rb
#!/usr/bin/env ruby
puts Gem.path
# fake-repxe-host.sh
bundle exec --keep-file-descriptors $PWD/foo.rb 
bundle exec --keep-file-descriptors ./foo.rb
```

http-418: Cool beans.

PR #921, opened 2017-07-28
Change to make HBase master heap memory configurable through attribute

bijugs:
Currently there is a bug which prevents HBase master JVM heap size takes the value set for region servers. This change will make the configuration of master heap non dependent on region server memory setting.

**Note:** This may require change to wrapper cookbooks if HBase memory values are overwritten through wrapper.


vt0r:
I don't think this will break anything, but applying the Breaking tag just in case someone has overridden the heap size in their env files and wants the same for the masters.

bijugs:
Thanks @vt0r .

PR #925, opened 2017-08-01
Allow jvmkill build recipe to work on internet-disconnected clusters

http-418:


leochen4891:
Nice catch. Thanks Andrew

PR #926, opened 2017-08-01
Fix PDNS dependency, add silversearcher

http-418:


http-418:
master is currently broken because pdns 4.0.0 changes the interface and dependencies _again_

PR #931, opened 2017-08-11
Adding VD Disk Support

NinjaLabib:
In cases such as OpenStack, it uses virtual disks which we don't currently account for


http-418:
The regex on this needs to be reversed, so that `[a-z]` is the optional stazla.  Currently it will not match `/dev/vdz`, for example.

Issue #729, opened 2016-12-16
Chef failing to create/populate `bootstrap-gpg-public_key_base64` on first run


vt0r:
The RuntimeError that occurs on at least one node during the run of `tests/automated_install.sh`:

```
10.0.100.13 [2016-12-15T17:29:02-05:00] WARN: Could not require chef-vault!
10.0.100.13   
10.0.100.13   ================================================================================
10.0.100.13   Recipe Compile Error in /var/chef/cache/cookbooks/bcpc/recipes/default.rb
10.0.100.13   ================================================================================
10.0.100.13   
10.0.100.13   RuntimeError
10.0.100.13   ------------
10.0.100.13   Failed to find value for bootstrap-gpg-public_key_base64!
10.0.100.13   
10.0.100.13   Cookbook Trace:
10.0.100.13   ---------------
10.0.100.13     /var/chef/cache/cookbooks/bcpc/libraries/utils.rb:151:in `get_config!'
10.0.100.13     /var/chef/cache/cookbooks/bcpc/recipes/cluster_local_repository.rb:54:in `from_file'
10.0.100.13     /var/chef/cache/cookbooks/bcpc/recipes/default.rb:38:in `from_file'
```

It gets populated on the second Chef run, so it's a convergence issue, but we should try to figure out how to make this convergence happen in a single Chef run of the bootstrap VM to avoid hacks like this monkey patch I used to sidestep the issue:

``` diff
--- a/tests/automated_install.sh
+++ b/tests/automated_install.sh
@@ -106,6 +106,9 @@ printf "Snapshotting post-Cobbler\n"
 printf "#### Chef all the nodes\n"
 vagrant ssh -c "sudo apt-get install -y sshpass"
 
+printf "#### Run Chef one time on the bootstrap node first\n"
+vagrant ssh -c "sudo chef-client --once"
+
 printf "#### Chef machine bcpc-vms\n"
 vagrant ssh -c "cd chef-bcpc; ./cluster-assign-roles.sh $ENVIRONMENT Basic"
 vagrant ssh -c "cd chef-bcpc; ./cluster-assign-roles.sh $ENVIRONMENT Bootstrap"
```

vt0r:
Affects current master and 2.1.1

http-418:
Fixed in #769

PR #933, opened 2017-08-14
bootstrap-gpg issue

NinjaLabib:
Repeating my comment: 
For some reason that I don't understand. My hypothesis is that  we need to do this before BCPC-Bootstrap so that we have os in the chef vault to allow bootstrap-gpg to be properly put inside the configs/Test-Laptop data bag as we aren't even installing the chef vault until after the first apt run.


vt0r:
Fixes #729 (again)

PR #934, opened 2017-08-14
Stop all clients before reindexing a chef server

http-418:
Allan discovered that `chef-server-ctl reindex` discards the index before rebuilding it.  If a client happens to chef during the re-indexing process, it can converge with a bad list of head nodes, which can break the cluster.

This change to C-A-R stops chef clients on all known nodes before attempting to re-index the chef server.

### DIFF
--- a/cluster_assign_roles.rb
+++ b/cluster_assign_roles.rb
@@ -69,7 +69,7 @@ class ClusterAssignRoles
   #   nothing
   # side-affect:
   #   updates Chef-server with runlists for nodes passed in
-  def assign_roles(nodes:, runlist: nil)
+def assign_roles(nodes:, runlist: nil)
     nodes.each do |node|
       chef_node_object = ridley.node.find(node[:fqdn])
 
@@ -113,6 +113,7 @@ class ClusterAssignRoles
       else
         if ii % 60 == 0
           if ii == 0
+            stop_all_clients
             reindex_chef_server
           end
           puts "Waiting for nodes to appear in search results (#{search})..."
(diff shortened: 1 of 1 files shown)


cluster_assign_roles.rb
```diff
@@ -420,6 +421,26 @@ def ssh(host:, username:, password:, command:, streaming: false)
...
+  def stop_all_clients
+    target_nodes = parse_cluster_txt.reject do |entry|
+      entry[:runlist].include?('SKIP')
+    end
+
+    target_nodes.each do |node|
+      result = ssh(host: node[:ip_address],
+                   username: 'ubuntu',
+                   password: cobbler_root_password,
+                   command: 'killall -9 chef-client',
```
vt0r: Probably needs to have `sudo` at the beginning. Might not hurt to also verify no `chef-client` processes are lingering after this, even though a `kill -9` should get the job done.

aespinosa:
Cherry-picking for 2.3.x

cluster_assign_roles.rb
```diff
@@ -420,6 +421,26 @@ def ssh(host:, username:, password:, command:, streaming: false)
...
+  def stop_all_clients
+    target_nodes = parse_cluster_txt.reject do |entry|
+      entry[:runlist].include?('SKIP')
+    end
+
+    target_nodes.each do |node|
+      result = ssh(host: node[:ip_address],
+                   username: 'ubuntu',
+                   password: cobbler_root_password,
+                   command: 'sudo killall -9 chef-client',
```
aespinosa: Discussed with @amithkanand that the chef-client will not shutdown all the time and in practice has to be done several times.

Code in https://github.com/bloomberg/chef-bach/blob/master/repxe_host.rb#L314-L357 probably needs to be copied over so that c-a-r will have the same consistency as the repxe scripts.

PR #936, opened 2017-08-14
Stop all clients before reindexing a chef server

aespinosa:
backport of pr #934


cluster_assign_roles.rb
```diff
@@ -418,6 +419,26 @@ def ssh(host:, username:, password:, command:, streaming: false)
...
+  def stop_all_clients
+    target_nodes = parse_cluster_txt.reject do |entry|
+      entry[:runlist].include?('SKIP')
+    end
+
+    target_nodes.each do |node|
+      result = ssh(host: node[:ip_address],
+                   username: 'ubuntu',
+                   password: cobbler_root_password,
+                   command: 'sudo killall -9 chef-client',
```
cbaenziger: I'd recommend not just kill -9'ing things; we hav ea lot of process which should be more gently nudged. Can we please do a reasonable `TERM`, `QUIT`, `KILL` escalation?

PR #944, opened 2017-08-16
Reindex the chef-server safely by stopping the chef-clients

aespinosa:
Addresses doing not simply doing a `killall -9 chef-client` as it is not reliable as discussed in #934 and #936 

Note: cherry-picked+squashed in 3.0.15


vt0r:
Might as well add another method to start them all back up with a flag (default: false I guess?).

aespinosa:
It is there https://github.com/bloomberg/chef-bach/blob/master/repxe_host.rb#L359-L373 already.  I can move it if it's needed now.

Just too lazy to refactor everything :)

vt0r:
Well at least it's easy to copy out to the library then. (Sorry - I couldn't resist).

Only reason I say we should have this is the repxe script will start one back up, but not all the others that are stopped by this new stop method during reindexing, meaning we'll have to manually start all chef-clients back up other than the C-A-R'd or repxed node.

http-418:
C-A-R starts chef clients when you run it on the cluster.  The user chooses the targets for C-A-R to touch.  If you specify no target, C-A-R touches the whole cluster.

When a re-index is necessary, we _must_ kill all the chef clients, to protect the cluster, regardless of whether the clients are on the target list or not.

Is it worse to:
1. Leave the clients dead on un-targeted nodes
or
2. Run chef-client on nodes that weren't specified as targets by the user

I favor option 1 -- leave stuff dead.   There is already a loud warning that C-A-R killed the service, and there is a simple remedy.  (Run C-A-R again with no target argument)

http-418:
Looking at the newest versions, this now relies on nodessh.sh, even though we already have ruby methods for this stuff.

I would much rather use the ruby methods than see a call out to a shell script.

aespinosa:
Use #1025 instead

PR #952, opened 2017-08-18
Move hive-scratch to user-directories

aespinosa:
So that hive sessions are under quota

Need to backport to master


cookbooks/bcpc-hadoop/recipes/hive_hcatalog.rb
```diff
@@ -53,15 +53,6 @@
...
 end
 
-bash 'create-hive-scratch' do
-  code <<-EOH
-  hdfs dfs -mkdir -p #{node['bcpc']['hadoop']['hive']['scratch']['dir']}
-  hdfs dfs -chmod 1777 #{node['bcpc']['hadoop']['hive']['scratch']['dir']}
-  hdfs dfs -chown hive:hdfs #{node['bcpc']['hadoop']['hive']['scratch']['dir']}
-  EOH
-  user 'hdfs'
-end
```
vt0r: Pretty sure this block is still required. We want to ensure the directory exists and is owned/permed properly. Obviously we'll have to replace `1777` with something more sane like `0700` and `hive:hdfs` with the proper user/group combo we use for user dirs.

aespinosa:
(Needs a cluster build)

cookbooks/bcpc-hadoop/recipes/hive_hcatalog.rb
cbaenziger: We shouldn't need hive to own something under the user's directory right @vt0r ?

vt0r: No, sir. We'd want to fill in the chown user dynamically to match the user it's being created for. Probably want `0700` perms as well

cbaenziger: Allan, can you check if Hive will create the directory if the user has perms and it doesn't exist? That way we can be lazy and blow this off (and nice not to run `O (users)` more HDFS actions).

aespinosa:
Rescope to 3.1

PR #932, opened 2017-08-13
HDFS DU

cbaenziger:
This is a working version of #86

PR #960, opened 2017-08-25
Maven enhancements

http-418:
I have been working on fixing Maven-related items in chef-bach in order to pursue my own items.  This PR is the result.

- Removes Kafka, MySQL JDBC, and JMXtrans from the bootstrap's repos.
- Created a proxy configuration for client nodes, so that VM clusters will work on a proxied network without mirroring more items to the bootstrap.
- Thanks to @cbaenziger , we have a final fix for the frustrating xinetd / tftp issue.


cbaenziger:
@ajones Did you verify the behavior of the Maven action when the Maven repository is down? My understanding is this will prevent our Chef from converging even if we have already installed this artifact?

cbaenziger:
Also, this seems to break those of us behind MITM proxies? I get:
```    [ERROR] Plugin org.apache.maven.plugins:maven-dependency-plugin:2.4 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-dependency-plugin:jar:2.4: Could not transfer artifact org.apache.maven.plugins:maven-dependency-plugin:pom:2.4 from/to central (https://repo.maven.apache.org/maven2): java.lang.RuntimeException: Unexpected error: java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty```

However, I can fix this for the bootstrap, since we have some SSL sharing code in [`Vagrantfile.local.rb`](https://github.com/bloomberg/chef-bach/blob/793e31b82b5269e753e15d29d2d50a889ee0ba78/Vagrantfile.local.rb) and in the #932 branch I [use](https://github.com/bloomberg/chef-bach/compare/master...cbaenziger:hdfsdu2#diff-8f8e11e8b7346a52105c3ff2157b6b50) that for Maven -- on the bootstrap. But how do you get your cluster VM's to talk https through a MITM, non-transparent proxy?

http-418:
I have only tested this with un-proxied maven repos.

It's definitely possible to configure maven for the proxies, but chef-bach isn't set up to do it, as that is not a configuration we have ever needed internally.

PR #961, opened 2017-08-25
Disable ForkJoin validation in Oozie

aespinosa:
Set as a workaround until the next HDP 2.6 maintenance release.

### DIFF
--- a/cookbooks/bcpc-hadoop/templates/default/ooz_oozie-site.xml.erb
+++ b/cookbooks/bcpc-hadoop/templates/default/ooz_oozie-site.xml.erb
@@ -11,6 +11,10 @@
         Refer to the oozie-default.xml file for the complete list of
         Oozie configuration properties and their default values.
     -->
+    <property>
+      <name>oozie.validate.ForkJoin</name>
+      <value>false</value>
+    </property>
 
     <property>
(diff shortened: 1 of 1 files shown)


leochen4891:
I would suggest to use an attribute

bijugs:
Better late than never. Can this be set at the workflow level based on the application need instead of disabling for the whole cluster?

bijugs:
Thanks @dbist . That is my suggestion i.e. not to set ``oozie.wf.validate.ForkJoin`` to ``false`` in the server property by to let users to set it in the workflow level.

dbist:
@bijugs the global property is `oozie.validate.ForkJoin=False`, per workflow property is `oozie.wf.validate.ForkJoin=false`, notice the subtle difference

bijugs:
:-)... setting "oozie.wf.validate.ForkJoin=false" at the server level will disable the validation for all workflows. Correct @dbist ? If so do we want to do that? Would it not be better to leave it to the users to determine whether they want to disable the validation on the workflow level.

dbist:
@bijugs `oozie.validate.ForkJoin=false` goes in oozie-site, `oozie.wf.validate.ForkJoin=false` goes in job.properties, its not meant to go server-side.

PR #969, opened 2017-09-08
Support rolling upgrade and determine installed HDP version

amithkanand:
This PR replaces use of hard coded `HDP` release number with a special variable `${hdp.version}` that is parsed by `Hadoop/Mapreduce` framework to determine installed `HDP` version on the node. This is required to support rolling upgrade of the cluster.

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/mapred.rb
+++ b/cookbooks/bcpc-hadoop/attributes/mapred.rb
@@ -10,15 +10,17 @@ default["bcpc"]["hadoop"]["yarn"]["app"]["mapreduce"]["am"]["staging-dir"] = "/u
 default[:bcpc][:hadoop][:mapreduce][:site_xml].tap do |site_xml|
   site_xml['mapreduce.admin.map.child.java.opts'] =
     '-server -Djava.net.preferIPv4Stack=true -Dhdp.version=' +
-    node[:bcpc][:hadoop][:distribution][:release].to_s
+    '${hdp.version}'
+
+  site_xml['mapreduce.admin.reduce.child.java.opts'] =
+    '-server -Djava.net.preferIPv4Stack=true -Dhdp.version=' +
+    '${hdp.version}'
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc-hadoop/attributes/mapred.rb
```diff
@@ -41,7 +43,7 @@
      '$PWD/mr-framework/hadoop/share/hadoop/hdfs/*',
      '$PWD/mr-framework/hadoop/share/hadoop/hdfs/lib/*',
      "#{hdp_path}/hadoop/lib/hadoop-lzo-0.6.0." +
-       "#{node[:bcpc][:hadoop][:distribution][:active_release]}.jar",
+       "${hdp.version}",
```
cbaenziger: Shouldn't this still need to end in `.jar`?

cookbooks/bcpc-hadoop/attributes/mapred.rb
```diff
@@ -10,15 +10,17 @@
 default[:bcpc][:hadoop][:mapreduce][:site_xml].tap do |site_xml|
   site_xml['mapreduce.admin.map.child.java.opts'] =
     '-server -Djava.net.preferIPv4Stack=true -Dhdp.version=' +
-    node[:bcpc][:hadoop][:distribution][:release].to_s
+    '${hdp.version}'
```
cbaenziger: Do we not need to pass `hdp.version` in on the command line still? I don't see `hdp.version` (or a fraction there of) in my ignorant use of `strings` in `/usr/hdp/current/hbase-regionserver/lib` on our machines.

aespinosa:
Thanks!

cherry-picked in v3.0.24 (with @cbaenziger 's corrections for the jar)

cbaenziger:
Thanks for the fix-up @aespinosa; @amithkanand any idea where the version string gets baked in?

PR #978, opened 2017-09-13
Correct Oozie's catalina path

aespinosa:
Makes oozie to boot back up to prevent the following error:

```
WARNING: Can't load server.xml from /usr/hdp/current/oozie-server/conf/server.xml
<Date> <Time> org.apache.catalina.startup.Catalina start
SEVERE: Cannot start server. Server instance is not configured.
```

UPDATE: To be backported to 3.0 (Backport done in 3.0.27)


cbaenziger:
@aespinosa It looks like there is a merge conflict here?

aespinosa:
Ah yes now that #982 is in.  This needs to be pulled together when #969 is resurrected.

PR #993, opened 2017-09-28
Use only chefdk on bootstrap

pu239ppy:
This resolves recent gem issues and version mismatches we've had during builds.  Also fixes long outstanding issues of gem and ruby runtime mismatches


pu239ppy:
Unfortunately I am still getting 
```
You cannot specify the same gem twice with different version requirements.
You specified: ridley (>= 0) and ridley (= 5.1.0)
```
Wll remove ridley from the root Gemfile

Issue #990, opened 2017-09-27
Add logic to disable HBase region replication


bijugs:
When HBase region replication is enabled it creates a peer  ``region_replica_replication``. Need to add logic to the code to disable this peer if region replication is disabled.

```
disable_peer 'region_replica_replication'

```

leochen4891:
Got this error

Got this error when building a VM cluster.

Recipe: bcpc-hadoop::hbase_repl
  * bash[stop_repl_remove_repl_peer] action run (skipped due to only_if)
  * bash[disable_hbase_region_repl_peer] action run
    [execute] /tmp/chef-script20171016-25575-1656kfr: line 1: hbase: command not found
    
    ================================================================================
    Error executing action `run` on resource 'bash[disable_hbase_region_repl_peer]'
    ================================================================================
    
    Mixlib::ShellOut::ShellCommandFailed
    ------------------------------------
    Expected process to exit with [0], but received '127'
    ---- Begin output of "bash"  "/tmp/chef-script20171016-25575-1656kfr" ----
    STDOUT: 
    STDERR: /tmp/chef-script20171016-25575-1656kfr: line 1: hbase: command not found
    ---- End output of "bash"  "/tmp/chef-script20171016-25575-1656kfr" ----
    Ran "bash"  "/tmp/chef-script20171016-25575-1656kfr" returned 127
    
    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/bcpc-hadoop/recipes/hbase_repl.rb
    
     35:   bash "disable_hbase_region_repl_peer" do
     36:     code <<-EOH
     37:       echo "disable_peer 'region_replica_replication'"|hbase shell
     38:       EOH
     39:     not_if "echo list_peers|hbase shell|grep region_replica_replication|grep ENABLED"
     40:     user 'hbase'
     41:   end
     42: end
    
    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/bcpc-hadoop/recipes/hbase_repl.rb:35:in `from_file'
    
    bash("disable_hbase_region_repl_peer") do
      action [:run]
      retries 0
      retry_delay 2
      default_guard_interpreter :default
      command "disable_hbase_region_repl_peer"
      backup 5
      returns 0
      user "hbase"
      code "      echo \"disable_peer 'region_replica_replication'\"|hbase shell\n"
      interpreter "bash"
      declared_type :bash
      cookbook_name "bcpc-hadoop"
[...]

PR #994, opened 2017-09-28
Change to disable region replication peer when region replication is disabled

bijugs:
This PR is for issue #990.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/hbase_config.rb
+++ b/cookbooks/bcpc-hadoop/recipes/hbase_config.rb
@@ -133,9 +133,9 @@ end
 #
 # if HBASE region replication is enabled the properties in this section will be included in hbase-site.xml
 #
-if node["bcpc"]["hadoop"]["hbase"]["region"]["replication"]["enabled"] == true then
+if node['bcpc']['hadoop']['hbase']['site_xml']['hbase.region.replica.replication.enabled'] then
   generated_values['hbase.regionserver.storefile.refresh.period'] = node["bcpc"]["hadoop"]["hbase_rs"]["storefile"]["refresh"]["period"]
-  generated_values['hbase.region.replica.replication.enabled'] = node["bcpc"]["hadoop"]["hbase"]["region"]["replication"]["enabled"]
+  generated_values['hbase.region.replica.replication.enabled'] = node['bcpc']['hadoop']['hbase']['site_xml']['hbase.region.replica.replication.enabled'].to_s
   generated_values['hbase.master.hfilecleaner.ttl'] = node["bcpc"]["hadoop"]["hbase_master"]["hfilecleaner"]["ttl"]
   generated_values['hbase.meta.replica.count'] = node["bcpc"]["hadoop"]["hbase"]["meta"]["replica"]["count"]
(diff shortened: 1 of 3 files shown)


cookbooks/bcpc-hadoop/recipes/hbase_config.rb
```diff
@@ -133,7 +133,7 @@
 #
 # if HBASE region replication is enabled the properties in this section will be included in hbase-site.xml
 #
-if node["bcpc"]["hadoop"]["hbase"]["region"]["replication"]["enabled"] == true then
+if node['bcpc']['hadoop']['hbase']['site_xml']['hbase.region.replica.replication.enabled'] == 'true' then
```
vt0r: Same question about the change from the bool (see above).

cookbooks/bcpc-hadoop/recipes/hbase_repl.rb
```diff
@@ -15,5 +16,27 @@
...
+
+#
+# To disable region replication peer
+#
+if (node['bcpc']['hadoop']['hbase']['site_xml']['hbase.region.replica.replication.enabled'] == 'false')
+  bash "disable_hbase_region_repl_peer" do
+    code <<-EOH
+      echo "disable_peer 'region_replica_replication'"|hbase shell
+      EOH
+    only_if "echo list_peers|hbase shell|grep region_replica_replication|grep ENABLED"
```
pu239ppy: Either way you invoke hbase shell to make a test, is there any harm to just blindly disabling without testing?

cookbooks/bcpc-hadoop/recipes/hbase_repl.rb
```diff
@@ -15,5 +16,27 @@
...
+    user 'hbase'
+  end
+else
+  bash "enable_hbase_region_repl_peer" do
+    code <<-EOH
+      echo "enable_peer 'region_replica_replication'"|hbase shell
+      EOH
+    only_if "echo list_peers|hbase shell|grep region_replica_replication|grep DISABLED"
```
pu239ppy: Either way you invoke hbase shell to make a test, is there any harm to just blindly disabling without testing?

cookbooks/bcpc-hadoop/recipes/hbase_repl.rb
bijugs: Hi @pu239ppy , with out the peer being present, ``hbase shell`` generate errors when ``disable/enable`` command is issued. The additional check is to make sure that the peer is defined and is in the correct state.

vt0r:
After talking with @bijugs about the move of this attribute to `site_xml`, I found out we're preferring any custom attributes we add that refer to K/V pairs in XML configs will be in one place - sort of like a custom configs namespace.

***Note**: the breaking tag was added because cluster owners who have enabled region replication explicitly using the old attribute will need to change their overrides as follows:*
```ruby
node['bcpc']['hadoop']['hbase']['region']['replication']['enabled'] = true
```
becomes:
```ruby
node['bcpc']['hadoop']['hbase']['site_xml']['hbase.region.replica.replication.enabled'] = true
```

Issue #997, opened 2017-09-29
princ/viphost@REALM principals and keytabs are not generated


pu239ppy:
````sudo kadmin.local -q 'ktadd -k /etc/security/keytabs/f-foo-bcpc-vm1.bcpc.example.com/hdfs.service.keytab -norandkey hdfs/f-foo-bcpc-vm1.bcpc.example.com@BCPC.EXAMPLE.COM hdfs/f-test-laptop.bcpc.example.com@BCPC.EXAMPLE.COM'

vagrant@foobcpc-bootstrap:~$ sudo klist -k /etc/security/keytabs/f-foo-bcpc-vm1.bcpc.example.com/hdfs.service.keytab
Keytab name: FILE:/etc/security/keytabs/f-foo-bcpc-vm1.bcpc.example.com/hdfs.service.keytab
KVNO Principal
---- --------------------------------------------------------------------------
   2 hdfs/f-foo-bcpc-vm1.bcpc.example.com@BCPC.EXAMPLE.COM
   2 hdfs/f-foo-bcpc-vm1.bcpc.example.com@BCPC.EXAMPLE.COM
   2 hdfs/f-foo-bcpc-vm1.bcpc.example.com@BCPC.EXAMPLE.COM
````

**Comment by pu239ppy** [comment]
Per kadmin.local
````
   ktadd
          ktadd [options] principal
          ktadd [options] -glob princ-exp

       Adds  a  principal, or all principals matching princ-exp, to a keytab file.  Each principal's keys are randomized in
       the process.  The rules for princ-exp are described in the list_principals command.

   list_principals
          list_principals [expression]

       Retrieves  all  or some principal names.  expression is a shell-style glob expression that can contain the wild-card
       characters ?, *, and [].  All principal names matching the expression are printed.  If no  expression  is  provided,
       all  principal names are printed.  If the expression does not contain an @ character, an @ character followed by the
       local realm is appended to the expression.
```

It does not appear that a simple list would work

pu239ppy:
perhaps this will fix it 
````ruby
    vip_principal = "#{service_principal}/#{viphost}@#{realm}"

    # Variable to hold all principals for a single keytab
    keytab_principals = [regular_principal, vip_principal]
   
    keytab_principals.each do |princ| 
      # Create the keytab file
      execute "creating-keytab-for-#{ke}" do
        command "kadmin.local -q 'xst -k #{keytab_path} " \
          "-norandkey #{princ}'"
        action :run
[...]

PR #998, opened 2017-09-29
vip host also needs principals (fix for #997)

pu239ppy:
FIX for #997 
as we no longer store vip hosts in cluster.txt we need to tell keytabs.rb about the vip host explicitly


amithkanand:
This change is required in `upload_keytabs.rb` as well.

pu239ppy:
Done

PR #1004, opened 2017-10-03
Add attribute to modify HDFS client socket timeout

bijugs:
Setting the DFS client socket timeout to 30 secs instead of the default 60 sec so that it can fail quickly in case a DN is down.


cbaenziger:
Biju is the default also 30 seconds? (If so, why's this change anything?)

bijugs:
My bad. The default is 60 secs not 30 secs.

cbaenziger:
Thanks @bijugs

Issue #1002, opened 2017-10-02
Failing to converge core-site


pu239ppy:
================================================================================
    Error executing action `run` on resource 'ruby_block[node.run_state[:core_site_generated_values]]'
    ================================================================================

    NoMethodError
    -------------
    undefined method `[]' for nil:NilClass

    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/bcpc-hadoop/recipes/core_site.rb:80:in `block (2 levels) in from_file'

https://github.com/bloomberg/chef-bach/blob/bee4c26390fda7d64366b6409f718691706d5500/cookbooks/bcpc-hadoop/recipes/core_site.rb#L78-L91

pu239ppy:
I am able to output run state from a ruby block that executes ````ruby puts node.run_state````

pu239ppy:
I think I found the vector :) https://github.com/bloomberg/chef-bach/commit/b9ad223ae7e01fde21b05b5889a6d8f5d014ec03#diff-9b58b641088e25d6d1218774d01fc2d9

PR #1007, opened 2017-10-04
Fix core site error seen

vineshcpaul:
Change the attribute reference from ":" to "'"    
    Error executing action `run` on resource 'ruby_block[node.run_state[:core_site_generated_values]]'

    NoMethodError
    -------------
    undefined method `[]' for nil:NilClass


pu239ppy:
This was already fixed in #1006 by reverting the strings back to symbols.  However, the symbil to string change was done due to a rubocop error.  Perhaps it may be worth it to revert #1006

pu239ppy:
Another fix for #1002

cbaenziger:
Thanks @vineshcpaul for confirming this built.

PR #1017, opened 2017-10-20
make smoke test user an admin of some hbase namespace

pu239ppy:
Currently for the hbase portion of the smoke test a list command is issued, for more complicated work in the future added a configurable namespace to be created


cookbooks/bcpc-hadoop/recipes/smoke_test_wrapper.rb
```diff
@@ -22,15 +22,25 @@
...
 
 include_recipe "bcpc-hadoop::smoke_test_user"
 
+# create hbase namespace for tester
+bash "HBASE namespace for smoke tests #{hbase_ns}" do
+  code <<-EOH
+  echo "create_namespace '#{hbase_ns}'" | hbase shell
+  EOH
+  user 'hbase'
+  not_if "hbase shell <<< 'list_namespace' | grep #{hbase_ns}", user: 'hbase'
```
vt0r: Style/Safety: Missing single quotes around the grep string.

PR #1018, opened 2017-10-20
Correct Chef-BACH #990

cbaenziger:
This addresses Chef-BACH #990's fix in #994 which then breaks a fresh VM cluster. This was tested on my HDFS-DU branch rebased on master last night.

### DIFF
--- a/stub-environment/roles/BCPC-Hadoop-Head-HBase.json
+++ b/stub-environment/roles/BCPC-Hadoop-Head-HBase.json
@@ -3,8 +3,8 @@
   "json_class": "Chef::Role",
   "run_list": [
     "role[Basic]",
-    "recipe[bcpc-hadoop::hbase_repl]",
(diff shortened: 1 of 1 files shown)


bijugs:
LGTM. BTW there was no changes made to role files in #994 and this bug seems to have originated long before that.

cbaenziger:
Thanks @bijugs; yes I think the issue is hbase_repl.rb [line 39](https://github.com/bloomberg/chef-bach/pull/994/files#diff-60f22c73c9cad53dd843044ffabacb42R39) will fail due to no `hbase` and being a `not_if`, then [line 37](https://github.com/bloomberg/chef-bach/pull/994/files#diff-60f22c73c9cad53dd843044ffabacb42R37) gets called and fails. Previous tests, if replication was disabled, were all `only_if` so would gracefully fail.

Issue #1019, opened 2017-10-24
unable to build cluster on ubuntu 14.04


saumilmayani:
```
cat /etc/os-release
NAME="Ubuntu"
VERSION="14.04.5 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 14.04.5 LTS"
VERSION_ID="14.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
```

Steps 
```
vboxmanage list dhcpservers
sudo vboxmanage list dhcpservers
sudo pkill -f VBox
git clone https://github.com/bloomberg/chef-bach.git
mv chef-bach/ chef-bcpc
cd chef-bcpc/
git branch

git remote add ipxe_delete https://github.com/cbaenziger/chef-bach.git
git remote -v
git pull ipxe_delete ipxe_delete

==========
remote: Counting objects: 43, done.
remote: Total 43 (delta 30), reused 30 (delta 30), pack-reused 13
Unpacking objects: 100% (43/43), done.
From https://github.com/cbaenziger/chef-bach
 * branch            ipxe_delete -> FETCH_HEAD
 * [new branch]      ipxe_delete -> ipxe_delete/ipxe_delete
Auto-merging vbox_create.sh
CONFLICT (content): Merge conflict in vbox_create.sh
Automatic merge failed; fix conflicts and then commit the result.
==========

vi vbox_create.sh

<<<<<<< HEAD
# populate the VM_LIST array from cluster.txt
# HACK: This is called prior to cluster.txt modification
code_to_produce_vm_list="
require './lib/cluster_data.rb';
include BACH::ClusterData;
cp=ENV.fetch('BACH_CLUSTER_PREFIX', '');
cp += '-' unless cp.empty?;
vms = parse_cluster_txt(File.readlines('cluster.txt'))
puts vms.map{|e| cp + e[:hostname]}.join(' ')
"
VM_LIST=( $(ruby -e "$code_to_produce_vm_list") )
export VM_LIST

=======
code_to_produce_vm_list="
require './lib/cluster_data.rb';
include BACH::ClusterData;
cp=ENV.fetch('CLUSTER_PREFIX', '');
cp += '-' unless cp.empty?;
vms = parse_cluster_txt(File.readlines(File.join(
[...]

PR #1020, opened 2017-10-24
Fixing ./tests/automated_install.sh: line 31: [: !=: unary operator expected

saumilmayani:
Fixing ./tests/automated_install.sh: line 31: [: !=: unary operator expected
Reported in #1019


cbaenziger:
@saumilmayani Thanks! For reference, I would recommend not having a space in `BACH_CLUSTER_PREFIX` -- lots of downstream things will explode later.

PR #1022, opened 2017-11-01
Clear /tmp every week

aespinosa:
Insert a cronjob recipe in base bcpc::default.

# Operation Notes

On first deploy, the deletion of `/tmp` might take a long time.  Chef failures need to be tracked on first run.

Or run it by hand at first to prevent chef failures from timeouts in the first place.

### DIFF
--- a/cookbooks/bcpc/attributes/cronjobs.rb
+++ b/cookbooks/bcpc/attributes/cronjobs.rb
@@ -1,0 +1,4 @@
+default['bcpc']['cronjobs'].tap do |cron|
+  cron['clear_tmp']['frequency'] = 604_800 # in seconds (at 1 week)
+  cron['clear_tmp']['atime_age'] = 7 # in days
+end
--- a/cookbooks/bcpc/recipes/cronjobs.rb
+++ b/cookbooks/bcpc/recipes/cronjobs.rb
@@ -1,0 +1,32 @@
+# Cookbook Name:: bcpc
+# Recipe:: cronjobs
+#
+# Copyright 2017, Bloomberg Finance L.P.
+#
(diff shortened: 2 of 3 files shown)


cbaenziger:
I'd much rather this run from Chef to keep on-top of this, keep this splayed and to record if we are having issues; otherwise we have a synchronize hit on the root-disks' performance across the cluster.

vt0r:
Also would be nice to use something like [tmpreaper](https://packages.ubuntu.com/trusty/admin/tmpreaper) for a bit more control, a la [this cookbook](https://github.com/customink-webops/tmpreaper), but maybe there's one that's been updated more recently. Also, it's not so awesome that `tmpreaper` is only in `universe`, but I guess it could be worse...

aespinosa:
@cbaenziger Update to use guard statement like in the pre-chef 13.x versions of the apt cookbook (See commit descrption in https://github.com/bloomberg/chef-bach/pull/1022/commits/2b27714626d22406c1f4c9b592b698a344dddd47 )

I'll need to look at how tmpreaper's packaged @vt0r .  Glossing over some docs, it installs entries in cron.daily.  Need to make sure there's no unwanted surprises before we pull it in.

http-418:
I feel pretty good about this PR as-is.  It's simple.

vt0r:
Found a pretty clean and simple provider that seems to address both my and Clay's concerns:

https://github.com/racker/cookbook-cleanup

Also sets up a nice resource we can use for any further FS cleanup tasks and should make splay easier. Could even be wrapped with a `locking_resource` if someone wanted to get super fancy (pls no).

aespinosa:
@cbaenziger done addressing all comments

cbaenziger:
Thanks Allan!

PR #1025, opened 2017-11-08
Remove chef-client process checks on repxe

aespinosa:
The operational assumption is that chef keeps running.  In a disabled chef-client, the cronjobs will simply fail at the beginning of its run


pu239ppy:
It may be preserving 
https://github.com/aespinosa/chef-bach/blob/0308bd2f71918c0d5db682ab999b2f7bc729a0eb/repxe_host.rb#L335
in case the client is running

aespinosa:
Retaining chef-client as a service for now

PR #1029, opened 2017-11-10
Specify alternative db for hive metastore

pu239ppy:
This pull requests grants us ability to specify alternative database parameters such as a different server and name for the metastore database


pu239ppy:
This is going to be much harder then initially thought as 'hive@' and 'metastore' are hardcoded all over the place

pu239ppy:
Squahed

PR #1031, opened 2017-11-24
ignore unenroll failure during repxe

pu239ppy:
during repxe if we do not indicate `-n`, repxe_host.rb will attempt to first unenroll the host,  it is possible however that the host may not already be enrolled.  As a result the script crashes and we have to manually enroll just to get it to work.  I propose we ignore this here, and let it fail at enroll phase if there is indeed an issue

### DIFF
--- a/repxe_host.rb
+++ b/repxe_host.rb
@@ -496,7 +496,11 @@ if __FILE__ == $PROGRAM_NAME
   unless options[:newmachine]
     delete_node_data(vm_entry)
     rotate_vault_keys
-    cobbler_unenroll(vm_entry)
+    begin
+      cobbler_unenroll(vm_entry)
+    rescue
+      puts "WARNING: Unenroll failed, will continue anyway" 
+    end
   end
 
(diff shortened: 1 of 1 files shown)


cbaenziger:
Any reason not to just see if the machine is enrolled woth `cobbler system list` before running unenroll - that way we know if Cobbler's horked or if the machine is simply not present?

repxe_host.rb
```diff
@@ -496,7 +496,11 @@ def graceful_shutdown(chef_env, vm_entry)
   unless options[:newmachine]
```
vt0r: We should probably also remove the `newmachine` option (`-n`) with this change.

pu239ppy: Makes sense at this point it is useless

pu239ppy:
@cbaenziger 1/2 of one .5 of the other I guess, it will fail either way.

vt0r:
I agree it saves a step to just attempt to remove without checking, and since ignoring it causes no discernible harm, I'm ok with merging this as is.

PR #1010, opened 2017-10-05
Bug fixes

cbaenziger:
This fixes:
* Halite `no_lazy_load` issues
* Removes stale Ubuntu 12.04 issues
* Allows YARN jobs to run with more containers
* Makes it clear that BIOS boot is not "legacy"
* Cleans-up old IPXE bits for multi-cluster attempts

PR #1033, opened 2017-11-30
Fix non-EFI build

http-418:
"Precise" doesn't work anymore, so we need to choose the trusty profile even for non-EFI hosts.


vt0r:
Already fixed in #1010, which should be ready to merge shortly. @cbaenziger is working on fixing the merge conflict as we speak.

PR #1034, opened 2017-12-04
Restart the chef-client when rotating logs

aespinosa:
The original call is a reload [1].  This makes the chef-client crash during logrotate because of  chef/chef#4578 . A restart is the prescribed workaround for this in Chef-client 12.19+.  The workaround is no longer needed on chef-client 13.2+.

[1] https://github.com/chef-cookbooks/chef-client/blob/master/attributes/default.rb#L180

### DIFF
--- a/cookbooks/bcpc/attributes/chef_client.rb
+++ b/cookbooks/bcpc/attributes/chef_client.rb
@@ -16,6 +16,8 @@ sudo_user = node['bcpc']['bootstrap']['admin']['user']
 #
 # XXX this keeps breaking Chef12 knife default['ohai']['disabled_plugins'] = [ 'passwd' ]
 
+# FIXME No longer needed in chef-client 13.2+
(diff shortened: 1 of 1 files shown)


aespinosa:
So. restarting the chef-client doesn't necessarily mean it can start back up again :(

```
cat /var/log/chef/client.log
[2018-04-01T06:35:22-04:00] FATAL: Chef is already running pid 43711
[2018-04-01T06:35:22-04:00] WARN: Chef RFC 062 (https://github.com/chef/chef-rfc/blob/master/rfc062-exit-status.md) defines the exit codes that should be used with Chef.  Chef::Application::ExitCode defines valid exit codes In a future release, non-standard exit codes will be redefined as GENERIC_FAILURE unless `exit_status` is set to `:disabled` in your client.rb. (CHEF-2) at /opt/chef/embedded/lib/ruby/gems/2.3.0/gems/chef-12.19.36/bin/chef-client:26:in `<top (required)>'.
Please see https://docs.chef.io/deprecations_exit_code.html for further details and information on how to correct this problem.
```

PR #1040, opened 2017-12-13
Restart the chef-server everyweek

aespinosa:
Manage the restart over cron because a chef-client restarting its own
chef-server doesn't seem like a good idea.


vt0r:
Please don't forget to backport this to 3.0.x

PR #1044, opened 2017-12-18
Separate management of client.rb and knife.rb

aespinosa:
Manage knife.rb through chef-client::config (except the log location) as well.  Allows the following:

* Remove the need for running `sudo knife ...` in the bootstrap.
* Upload failures in `knife cookbook upload ...` will be in STDOUT instead of being written in `/var/log/chef/client.log`.


vt0r:
This does not fix the log location for knife (to negate the need for using `sudo` to run knife commands), but that's out of scope of this PR anyway, so I'm ok with merging this as is.

http-418:
The code in this pull request is great, but we need detailed block comments explaining what we are doing, and why we are doing it.

chef-client logging vs knife logging, and sudo questions, are both non-obvious looking at the chef code

aespinosa:
Done documenting the `why?` of the change

PR #1051, opened 2017-12-22
Passwd fix

cbaenziger:
This allows one to override databag entries. Specifically this means one can now use the environment file for testing to set a human reasonable default password -- now `hadoopy`!

Also, this improves snapshot handling as it seemed Shoe-Less wasn't getting taken for the bootstrap VM previously; now a library function in `vbox_create` unifies snapshot handling and a state machine is called out in `automated_install.sh` for the snapshot names.

### DIFF
--- a/cookbooks/bcpc/libraries/utils.rb
+++ b/cookbooks/bcpc/libraries/utils.rb
@@ -144,10 +144,26 @@ def get_config(key, item=node.chef_environment, bag="configs")
   end
 
   #
-  # We should always provide the vault item if possible.
-  # If that fails, fall back to the data bag.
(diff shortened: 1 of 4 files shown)


cbaenziger:
Note: Just to be clear the feature allowing databag lookups in `getconfig` to be overriden by the environment is for human and test-kitchen testing. This feature should not be used normally as the reliance on security sensitive attributes in the environment will lead them to be more easily leaked (e.g. source control of the environment file or by machines which have no business being permissioned for a secret).

cbaenziger:
I've built both clusters with the cobbler password override set and clusters without it set.

PR #1052, opened 2017-12-26
Rearrange the prefixing of graphite and statsd

aespinosa:
The Influx+StatsD protocol will actually be used so tagging is
supported.  The prefixes are only needed in graphite.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/jmxtrans_agent.rb
+++ b/cookbooks/bcpc-hadoop/recipes/jmxtrans_agent.rb
@@ -1,12 +1,12 @@
 node['bcpc']['hadoop']['jmxtrans_agent'].tap do |agent|
   graphite = agent['graphite'].dup
-  def graphite.with_prefix(component)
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc-hadoop/recipes/jmxtrans_agent.rb
```diff
@@ -1,12 +1,12 @@
...
-  def graphite.with_prefix(component)
-    self.merge('namePrefix' => "#{component['name_prefix']}.#escaped_hostname#.")
+  def graphite.with_prefix(component, cluster)
+    self.merge('namePrefix' => "#{component['name_prefix']}.#{cluster}.#escaped_hostname#.")
   end
 
   statsd = agent['statsd'].dup
   def statsd.with_prefix(component)
-    self.merge('metricName' => "#{component['name_prefix']}.#escaped_hostname#")
+    self.merge('metricName' => "bach.#{component['name_prefix']}")
```
leochen4891: Does the metricName need to be distinguished by hostname?

PR #1056, opened 2017-12-28
Set the default jmx datatype to be gauges

aespinosa:
JMXTrans-agent sets them to counters by default.  We mostly use gauges
since we've been using graphite.


leochen4891:
What is the benefits of specifying the type to be gauge, and what is the possible impact?
I noticed some of the attributes are actually counters. e.g. NameNodeActivity->GetEditNumOps

aespinosa:
This is more for reporting metrics that is accurate.

So for example, the resulting statsd for jvm.memory.HeapMemoryUsage_committed is displayed 160GB in Statsd but in Graphite it is actually 20GB.  This is because Statsd is actually accumulating all the 20GB data points from JMX

aespinosa:
Done on whatever Lei said should be a counter.

PR #1060, opened 2018-01-02
Travis Style Check Support

cbaenziger:
This provides for us to enable [Travis CI](http://travis-ci.org) for pull request checking. Particularly this adds support for us to start using `rake` along with a few tasks:

* `rake clean` -- A stub to clean the repository up after a cluster build
* `rake style` -- Runs all `style:*` verbos checks
* `rake style:diff` -- Provides if your repo improves or degrades style (a quick go/no-go one can run)
* `rake style:foodcritic` -- foodcritic for all cookbooks
* `rake style:rubocop` -- RuboCop

I have tested this with TravisCI ensuring we improve or at least maintain style quality every commit. As soon as this goes into master, I can work with the Bloomberg organization owners to enable chef-bach for TravisCI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: ruby
+sudo: false
+rvm:
+  - 2.0
+  - 2
+bundler_args: --without cookbooks
+cache: bundler
+gemfile: Gemfile
+before_install:
+  - chef --version &> /dev/null || curl -L https://www.getchef.com/chef/install.sh | sudo bash -s -- -P chefdk -v 1.2.22
+  - eval "$(/opt/chefdk/bin/chef shell-init bash)"
+  - bundle config --local PATH vendor/bundle
+  - bundle config --local DISABLE_SHARED_GEMS true
+before_script:
+  - chef --version
+  - cookstyle --version
+  - foodcritic --version
+script: rake style:diff
--- a/Rakefile
+++ b/Rakefile
@@ -1,0 +1,41 @@
+#!/usr/bin/env rake
+
+require 'mixlib/shellout'
+
+chef_version = '12.19.36'
+namespace :style do
+  desc 'Style check with rubocop'
+  task :rubocop do
+    sh '/opt/chefdk/embedded/bin/rubocop -P'
+  end
+
+  desc 'Style check with foodcritic'
+  task :foodcritic do
+    sh '/opt/chefdk/embedded/bin/foodcritic --role-path ./stub-environment/roles/ --environment-path ./stub-environment/environments/Test-Laptop.json --epic-fail none cookbooks/'
+  end
+
+  desc 'Check style violation difference'
+  task('diff'.to_sym) do
+    sh './compare_style.sh'
+  end
+end
+
+desc 'Run style checks'
+task style: %w(style:rubocop style:foodcritic)
+
+desc 'Clean some generated files'
+task :clean do
+  %w(
+    **/Berksfile.lock
+    .bundle
+    .cache
+    **/Gemfile.lock
+    .kitchen
+    vendor
+    ../cluster
+    vbox
+  ).each { |f| FileUtils.rm_rf(Dir.glob(f)) }
+  # XXX should remove VBox VM's
+end
+
+task :default => 'style'
--- a/compare_style.sh
+++ b/compare_style.sh
@@ -1,0 +1,141 @@
+#!/bin/bash
+
+###########################################
+# Script to compare style check differences
+# between branches of code to make sure we
+# are cleaning up and not regressing code
+###########################################
+
+########################################
+# Function to gather master style output
+# arguments: file path to write rake sytle output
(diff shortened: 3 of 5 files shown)


cbaenziger:
For reference, I had to implement dynamic foodcritic style targets per cookbook due to issues whereby I could not get the FoodCritic raketask to effectively run `foodcritic -E stub-environment/environments/Test-Laptop.json -R stub-environment/roles -c 12.0.0 cookbooks/*`. There is some discussion of this at foodcritic/foodcritic#250 for future following.

.travis.yml
```diff
@@ -0,0 +1,18 @@
+language: ruby
+sudo: false
+rvm:
+  - 2.0
+  - 2
+bundler_args: --without cookbooks
+cache: bundler
+gemfile: Gemfile
+before_install:
+  - chef --version &> /dev/null || curl -L https://www.getchef.com/chef/install.sh | sudo bash -s -- -P chefdk
```
aespinosa: should we pin this to what is set to https://github.com/bloomberg/chef-bach/blob/master/build_bins.sh#L25 ?

Rakefile
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env rake
+
+chef_version = '12.14.89'
```
aespinosa: I think we're at 12.19.36

Rakefile
```diff
@@ -0,0 +1,70 @@
...
+
+  desc 'Run foodcritic checks'
+  task chef: %w(style:chef./) + \
+              Dir.glob('./cookbooks/*').map { |c| 'style:chef' + c }
+end
+
+desc 'Run style checks'
+task style: %w(style:ruby style:chef)
+
+desc 'Clean some generated files'
```
aespinosa: Are we planning to have other non-file cleaning related stuff? If not let's use the existing clean/ clobber facilities: http://docs.seattlerb.org/rake/doc/rakefile_rdoc.html#label-Clean+and+Clobber+Tasks

cbaenziger: Thanks for this pointer; mind if I hold this suggestion until I start the `rspec`/`chefspec` work where I will need to make sure Gem caches get cleaned-up?

.travis.yml
cbaenziger: Fixed to use the same version for now; ideally master should float though.

cbaenziger:
Hi @aespinosa let me know if this looks better and I can hold off the Rake changes for the next update. I verified this ran through on my personal Travis build test.

compare_style.sh
```diff
@@ -0,0 +1,141 @@
+#!/bin/bash
```
aespinosa: This shell script seems to be more of a crutch for Travis since it doesn't support tracking build data.

I think codeclimate will be the better public service tracking linting metrics.

Plus we won't need to maintain another shellscript

Rakefile
```diff
@@ -0,0 +1,70 @@
...
+
+chef_version = '12.19.36'
+namespace :style do
+  require 'rubocop/rake_task'
+  desc 'Style check with rubocop'
+
+  require 'foodcritic'
+  desc 'Style check with foodcritic for root'
+  orig_dir = Dir.pwd
+  FoodCritic::Rake::LintTask.new('chef./'.to_sym) do |t|
```
aespinosa: Rake tasks around the chefdk ecosystem tends to be less maintained than the main CLI scripts.

Problems in the broader community tent to occur in some versions of ChefDK where each individual tool gets rubygem activation conflicts.

CLI applications work around this by using appbundler where each CLI app can activate its own set of dependencies.  (At least that's how I understood how chefdk works)

TLDR it's probably better to shell out

Rakefile
```diff
@@ -0,0 +1,70 @@
...
+        chef_version: chef_version,
+        role_path: ['./stub-environment/roles/*'],
+        cookbook_path: c,
+        environment_path: ['./stub-environments/environments/Test-Laptop.json']
+      }
+    end
+    Dir.chdir(orig_dir)
+  end
+
+  RuboCop::RakeTask.new(:ruby) do |t|
```
aespinosa: shellout here as well

http-418:
As discussed in the 1 Feb 18 code review, changes will be needed on this PR:
- Allan's concerns about rake task wrapper objects should be addressed
- The new shell script should be removed

compare_style.sh
cbaenziger: I appreciate being able to check my changes before submitting to an external service but it is optional for one to use. I would rather we have shared tooling than folks implementing their own ad-hoc methods. Indeed CodeClimate seems like a better service than Travis for tracking metrics on code quality though.

Rakefile
cbaenziger: I have updated to shell out now.

Rakefile
cbaenziger: Ditto on updating for shell out here.

PR #1063, opened 2018-01-04
slider support

pu239ppy:
# Slider support
- YARN settings
- Install slider client on workers
- Update slider-env.sh

# Caveats
- The version of slider shipped currently does not grab HDFS delegation tokens, so even if a long running (> 7 days) process is not required, the cluster will not start, unless a keytab is provided
- Slider does not work with multihomed systems, the URL registered in yarn points to `host` as opposed to `f-host`, one way around this was to issue a `sudo hostname f-bcpc-vm3` on bcpc-vm3 prior to launching the cluster.  Opened https://issues.apache.org/jira/browse/SLIDER-1259

# Testing
Tested using a simple setup in https://github.com/pu239ppy/slider_poc

# What does it look like?
Here I am asking for two instances of a service, and the services are running on ephemeral ports
````
ubuntu   13491     1  0 16:37 ?        00:00:00 /usr/bin/python /disk/2/yarn/local/usercache/ubuntu/appcache/application_1515102595090_0151/container_1515102595090_0151_01_000003/app/install/mysvc.py 60468
ubuntu   13492     1  0 16:37 ?        00:00:00 /usr/bin/python /disk/2/yarn/local/usercache/ubuntu/appcache/application_1515102595090_0151/container_1515102595090_0151_01_000002/app/install/mysvc.py 60391
````

Here is what the exports query returns
````
{
  "host_port" : [ {
    "value" : "f-bcpc-vm3.bcpc.example.com:60468",
    "containerId" : "container_1515102595090_0151_01_000003",
    "tag" : "1",
    "level" : "component",
    "updatedTime" : "Mon Jan 08 11:37:13 EST 2018"
  }, {
    "value" : "f-bcpc-vm3.bcpc.example.com:60391",
    "containerId" : "container_1515102595090_0151_01_000002",
    "tag" : "2",
    "level" : "component",
    "updatedTime" : "Mon Jan 08 11:37:13 EST 2018"
  } ]

````

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/datanode.rb
+++ b/cookbooks/bcpc-hadoop/recipes/datanode.rb
@@ -321,3 +321,5 @@ locking_resource 'hadoop-hdfs-nodemanager-restart' do
   subscribes :serialize, 'log[jdk-version-changed]', :delayed
   subscribes :serialize, node['bcpc']['hadoop']['jmxtrans_agent']['nodemanager']['xml'], :delayed
 end
+
+include_recipe 'bcpc-hadoop::slider'
--- a/cookbooks/bcpc-hadoop/recipes/slider.rb
+++ b/cookbooks/bcpc-hadoop/recipes/slider.rb
@@ -10,6 +10,18 @@ end
 
 hdp_select('slider-client', node[:bcpc][:hadoop][:distribution][:active_release])
 
+site_xml = node.default[:bcpc][:hadoop][:yarn][:site_xml]
+
+slider_properties = {
+  'hadoop.registry.zk.quorum' =>
+    zk_hosts.map{ |h| float_host(h[:hostname]) +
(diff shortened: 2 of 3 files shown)


cbaenziger:
Hi @pu239ppy, is this tested on a cluster?

pu239ppy:
It is testing now

pu239ppy:
@cbaenziger I was able to launch Map/Reduce applications and spark shell

pu239ppy:
Added WIP label again.  While the proposed configuration does not harm YARN operation, it also does not do what is required

pu239ppy:
Tested full build

cbaenziger:
@pu239ppy did you file or find a JIRA on Slider failing to allow the hostname to be set?

pu239ppy:
@cbaenziger No, I am still trying to figure out where this is being set and if we can influence it.  Meanwhille, this will will work in non multi homed environments

PR #1065, opened 2018-01-05
opentsdb service fails to start due to missing JAVA_HOME

vineshcpaul:


cbaenziger:
Hi Vinesh, this should be handled by the wrapper recipe at https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/opentsdb.rb#L26-L27; did this not work for you?

aespinosa:
The wrapper wasn't invoked in https://github.com/bloomberg/chef-bach/blob/master/stub-environment/roles/BCPC-Hadoop-Head-OpenTSDB-Server.json#L5

aespinosa:
Use #1066 instead

PR #1066, opened 2018-01-05
Wrap bach_opentsdb directly

aespinosa:
Fix in favor of #1065 

@vineshcpaul can you try this approach instead and let us know if this works for you?


vt0r:
I jumped the gun a bit on merging this after the review, so I'm going to test it now by starting a build. I'll update once it builds and update the tags accordingly, otherwise I'll revert.

vt0r:
On the first attempt at building, I noticed this PR adds an `include_recipe` to `bcpc` without adding it as a dependency in the metadata file. I've made that addition locally and ran a second build.

The second build attempt failed due to some transient download timeout. Restarted it this morning. If that was the only fix needed, I'll make a new PR for that one-liner when the build completes.

PR #1078, opened 2018-01-17
Enable phoenix Query Server

pu239ppy:
This Pull Request enables the phoenix query server.  In current mode of operation PQS uses the HTTP/ principal to enable SPNEGO and then (sadly) to proxy users.  Also made ubuntu user a superuser to perform HBASE operations.

### Connecting to server
`phoenix-sqlline-thin http://f-bcpc-vm1`

### Interacting with phoenix/PQS
````
create table stuff (id integer primary key, price integer);
upsert into stuff (id, price) values (1,2);
upsert into stuff (id, price) values (3,4);

select * from stuff;

+-----+--------+
| ID  | PRICE  |
+-----+--------+
| 1   | 2      |
| 3   | 4      |
+-----+--------+

````

### Kerberos and PROXY USER
Currently PQS uses `HTTP/` principal for both SPNEGO auth and for proxying users, since access to HTTP/ is fairly common, any system service that uses HTTP/ presents an attack surface, hence the following does not have any attribute defaults and is only added in the Test-Laptop environment for testing purposes

```
       "core": {
           "site_xml": {
             "hadoop.proxyuser.HTTP.hosts": "*",
             "hadoop.proxyuser.HTTP.users": "*"
            }
         },
``` 

In a production environment it is recommended that `hadoop.proxyuser.HTTP.users": "*"` is replaced with an appropriate group permission that will limit proxy user access.
[PHOENIX-4533](https://issues.apache.org/jira/browse/PHOENIX-4533), will hopefully address this in near future.  In preparation for that, the included recipes already generate a `phoenixqs/` principal.  Furthermore for core-site.xml generation, I replaced the mapping of `HTTP/` principal to mapred https://github.com/pu239ppy/chef-bach/blob/f5149e814fd83cefda319de5adf45d68fe0ea1fc/cookbooks/bcpc-hadoop/recipes/core_site.rb#L30 as proxyuser was failing with a non obvious error, as is this does not seem to serve any purpose.

### Process ownership
Currently due to the limitations of the PQS init script, the service has to run as root, the template take a `pqs_runas` user and interpolates into the init script, but then init script does nothing with it.  In the future will work to modify this to improve security.  This means that `default['bcpc']['hadoop']['phoenix']['phoenixqs']['username'] ` is used to assign keytab ownership, but not to launch the process.

### CODE REVIEW

**Comment by RiteshVajariya** [comment]
@bijugs can you please help review this PR?

**Comment by http-418** [comment]
This needs review from @bijugs before merge

File: stub-environment/roles/BCPC-Hadoop-Head-HBase.json [hunk 1]
```diff
@@ -3,7 +3,9 @@
   "json_class": "Chef::Role",
   "run_list": [
     "role[Basic]",
+    "recipe[bcpc-hadoop::phoenixqs_hbase_site]",
     "recipe[bcpc-hadoop::hbase_master]",
+    "recipe[bcpc-hadoop::phoenix_query_server]",
```
bijugs: This will install PQS on all HBase head nodes. Is that the intention?

stub-environment/roles/BCPC-Hadoop-Head-HBase.json
```diff
@@ -3,7 +3,9 @@
   "json_class": "Chef::Role",
   "run_list": [
     "role[Basic]",
+    "recipe[bcpc-hadoop::phoenixqs_hbase_site]",
```
bijugs: Can we include this as part of ``recipe[bcpc-hadoop::phoenix_query_server]`` as we do in [hbase_master.rb](https://github.com/pu239ppy/chef-bach/blob/f5149e814fd83cefda319de5adf45d68fe0ea1fc/cookbooks/bcpc-hadoop/recipes/hbase_master.rb#L3)

cookbooks/bcpc-hadoop/recipes/phoenixqs_hbase_site.rb
```diff
@@ -0,0 +1,14 @@
+# vim: tabstop=2:shiftwidth=2:softtabstop=2
+
+include_recipe 'bcpc-hadoop::phoenixqs_kerberos'
+
+site_xml = node.default[:bcpc][:hadoop][:hbase][:site_xml]
+generated_values = {}
```
bijugs: Can we add the property ``phoenix.queryserver.http.port`` so that the default value can be overwritten when required?

cookbooks/bcpc-hadoop/attributes/phoenixqs.rb
```diff
@@ -0,0 +1,10 @@
+# vim: tabstop=2:shiftwidth=2:softtabstop=2
+
+default['bcpc']['hadoop']['phoenix']['phoenixqs']['username'] = 'phoenixqs'
+default['bcpc']['hadoop']['phoenix']['phoenixqs']['principal'] =\
+  "HTTP/#{float_host(node[:fqdn])}@#{node[:bcpc][:hadoop][:kerberos][:realm]}"
+default['bcpc']['hadoop']['phoenix']['phoenixqs']['keytab'] =\
+  "#{node[:bcpc][:hadoop][:kerberos][:keytab][:dir]}/" +
+  node[:bcpc][:hadoop][:kerberos][:data][:spnego][:keytab]
+default['bcpc']['hadoop']['phoenix']['phoenixqs']['localuser'] = true
+default['bcpc']['hadoop']['phoenix']['phoenixqs']['serialization'] = 'JSON'
```
bijugs: Can we add an attribute for ``phoenix.queryserver.http.port`` with value set to the default of ``8765`` so that it can be overwritten.

cookbooks/bcpc-hadoop/recipes/phoenix_query_server.rb
```diff
@@ -0,0 +1,42 @@
...
+
+
+template '/etc/init.d/pqs' do
+  source 'etc_initd_pqs.erb'
+  variables(qs_runas: qs_runas) 
+  mode 0o755
+  notifies :restart, 'service[pqs]', :delayed
+end
+
+service 'pqs' do
```
bijugs: The ``qs_runas`` user is created and added to ``hadoop`` group when the attribute ``node['bcpc']['hadoop']['phoenix']['phoenixqs']['localuser']`` is set to ``true``. By setting it ``false`` this step will fail since the init script will not be able to start PQS. no?

cookbooks/bcpc-hadoop/recipes/phoenix_query_server.rb
```diff
@@ -0,0 +1,42 @@
...
+
+group "add #{qs_runas} to hadoop" do
+  group_name 'hadoop'
+  members [ qs_runas ]
+  append true
+  action :nothing
+  only_if { node['bcpc']['hadoop']['phoenix']['phoenixqs']['localuser'] }
+end
+
+configure_kerberos 'phoenixqs_keytab' do
```
bijugs: The ``qs_runas`` user is created and added to ``hadoop`` group when the attribute ``node['bcpc']['hadoop']['phoenix']['phoenixqs']['localuser']`` is set to ``true``. Do we need to perform the same condition checking here?

cookbooks/bcpc-hadoop/attributes/phoenixqs.rb
```diff
@@ -0,0 +1,10 @@
+# vim: tabstop=2:shiftwidth=2:softtabstop=2
+
+default['bcpc']['hadoop']['phoenix']['phoenixqs']['username'] = 'phoenixqs'
```
bijugs: Can we add an attribute so that the query server can be enabled/disabled through ``chef`` without performing these actions manually.

stub-environment/roles/BCPC-Hadoop-Head-HBase.json
```diff
@@ -3,7 +3,9 @@
...
     "recipe[bcpc-hadoop::hbase_master]",
+    "recipe[bcpc-hadoop::phoenix_query_server]",
     "recipe[bcpc-hadoop::hbase_repl]",
     "recipe[bcpc::diamond]"
   ],
```
bijugs: We should have PQS behind ``haproxy``` or something similar for service availability since any node can go into maintenance.

cookbooks/bcpc-hadoop/attributes/phoenixqs.rb
pu239ppy: Enablement/disablement is done on a role basis.  For maintenance purposes we would just shut down chef-client

cookbooks/bcpc-hadoop/attributes/phoenixqs.rb
pu239ppy: in the environment yuo can add node['bcpc']['hadoop']['hbase']['site_xml'] as an override without having it declared explicitly

cookbooks/bcpc-hadoop/recipes/phoenix_query_server.rb
pu239ppy: All this does is kinit

bijugs: Sorry the comment was meant for the next step.

cookbooks/bcpc-hadoop/recipes/phoenixqs_hbase_site.rb
pu239ppy: in the environment yuo can add node['bcpc']['hadoop']['hbase']['site_xml'] as an override without having it declared explicitly

stub-environment/roles/BCPC-Hadoop-Head-HBase.json
pu239ppy: Yes

stub-environment/roles/BCPC-Hadoop-Head-HBase.json
pu239ppy: PQS uses zookeeper for location and load balancing, we can add haproxy in a follow up I suppose

ronny-macmaster:
This eventually needs to be refactored to a separate Phoenix cookbook. 
We will merge this for now.

ronny-macmaster:
Don't run this on a physical cluster yet.
Wait until the new hortonworks pqs that includes init scripts.

PR #1081, opened 2018-01-19
Change to make ZK connection more robust. issue 353

bijugs:
**Draft PR for comments**
-----------------------------
Making the ``ZooKeeper`` connect a bit more robust. Based on ``ZOOKEEPER-1057`` looks like if there is a failure with connecting to first node in the ZK host names passed i.e it is down, then connection seem to fail even though there is a ZK quorum.

### DIFF
--- a/cookbooks/bcpc-hadoop/libraries/utils.rb
+++ b/cookbooks/bcpc-hadoop/libraries/utils.rb
@@ -162,15 +162,33 @@ def storage_host(*args)
   end
 end
 
+def zk_connect(zk_host)
+  require 'rubygems'
+  require 'zookeeper'
+  zk_host_array = zk_host.split(",")
+  begin
+    zk_host_array.each_with_index do |d, idx|
+      temp = zk_host_array[idx]
+      zk_host_array[idx] = zk_host_array[0]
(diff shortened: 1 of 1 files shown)


cbaenziger:
@bijugs thanks for these edits, however, as mentioned in #1074, I'd like to see a lot of this code removed in favor of bloomberg/locking_resource-cookbook (e.g.https://github.com/cbaenziger/chef-bach/commit/aa3c213f807ceeaa01cbd9cd903c4a6ed4c5cce3). What are your thoughts?

bijugs:
Hi @cbaenziger , If we are close to incorporating the changes from ``locking-resource-cookbook`` soon then this is not required. If not, it may be good to consider any change to improve ZK connection so that we can help alleviate cluster issues.

cbaenziger:
Hi @bijugs can this code be moved to locking_restart now that all locking code is there? For format zkfc, perhaps we can make a clean API from locking resource for it to use so we don't have duplicate code (or we just accept duplication for now).

cbaenziger:
@bijugs Please commit this to  bloomberg/locking_resource-cookbook; closing here as no-longer applicable.

Issue #1091, opened 2018-02-07
cluster address is blank in vm1 when mysql recipe is run


vineshcpaul:
MySQL fails to start as the cluster address is blank in /etc/mysql/conf.d/wsrep.cnf

[0m    Mixlib::ShellOut::ShellCommandFailed^[[0m^M^M
    ------------------------------------^[[0m^M^M
    Expected process to exit with [0], but received '1'^M^M
^[[0m    ---- Begin output of /etc/init.d/mysql start ----^M^M
^[[0m    STDOUT: * Starting MySQL (Percona XtraDB Cluster) database server mysqld^M^M
^[[0m     * The server quit without updating PID file (/var/run/mysqld/mysqld.pid).^M^M
^[[0m       ...fail!^M^M
^[[0m    STDERR: ^M^M
^[[0m    ---- End output of /etc/init.d/mysql start ----^M^M
^[[0m    Ran /etc/init.d/mysql start returned 1^[[0m^M^M
    ^M^M
^[[0m    Resource Declaration:^[[0m^M^M

aespinosa:
the `wsrep.conf` files are meant to only include "other heads" in the cluster

PR #1092, opened 2018-02-07
MySQL is not starting as vm1 is missing the cluster ip address.

vineshcpaul:
Fix for issue : https://github.com/bloomberg/chef-bach/issues/1091


aespinosa:
See https://github.com/bloomberg/chef-bach/issues/1091#issuecomment-385400123

Issue #1096, opened 2018-02-11
HBASE Issue frequently encountered during C-A-R


pu239ppy:
I am frequently seeing this error when chefing HBase Master

````
  * service[hbase-master] action start
    - start service service[hbase-master]
  * ruby_block[handle_prev_hbase_master_restart_failure] action create

    ================================================================================
    Error executing action `create` on resource 'ruby_block[handle_prev_hbase_master_restart_failure]'
    ================================================================================

    TypeError
    ---------
    no implicit conversion of nil into String

    Cookbook Trace:
    ---------------
    /var/chef/cache/cookbooks/bcpc-hadoop/libraries/utils.rb:341:in `process_restarted_after_failure?'
    /var/chef/cache/cookbooks/bcpc-hadoop/definitions/hadoop_service.rb:38:in `block (3 levels) in from_file'

    Resource Declaration:
    ---------------------
    # In /var/chef/cache/cookbooks/bcpc-hadoop/definitions/hadoop_service.rb

     32:     ruby_block "handle_prev_#{params[:service_name].gsub('-','_')}_restart_failure" do
     33:       block do
     34:         Chef::Log.info "Need to restart #{params[:service_name]} since it failed during the previous run. Another node's restart process failure is a possible reason"
     35:       end
     36:       action :create
     37:       only_if { node[:bcpc][:hadoop][params[:service_name].gsub('-','_').to_sym][:restart_failed] and
     38:               !process_restarted_after_failure?(node[:bcpc][:hadoop][params[:service_name].gsub('-','_').to_sym][:restart_failed_time],"#{params[:process_identifier]}")}
     39:     end
     40:     #
    Compiled Resource:
    ------------------
    # Declared in /var/chef/cache/cookbooks/bcpc-hadoop/definitions/hadoop_service.rb:32:in `block in from_file'

    ruby_block("handle_prev_hbase_master_restart_failure") do
[...]

PR #1098, opened 2018-02-11
Fix issue #1096

pu239ppy:
This is a fix for #1096

### DIFF
--- a/cookbooks/bcpc-hadoop/libraries/utils.rb
+++ b/cookbooks/bcpc-hadoop/libraries/utils.rb
@@ -338,7 +338,7 @@ def process_restarted_after_failure?(restart_failure_time, process_identifier)
   require 'time'
   begin
     start_time = process_start_time(process_identifier)
-    if start_time.nil? && (Time.parse(restart_failure_time).to_i < Time.parse(start_time).to_i)
+    if not start_time.nil? && (Time.parse(restart_failure_time).to_i < Time.parse(start_time).to_i)
       Chef::Log.info("#{process_identifier} seem to be started at #{start_time} after last restart failure at #{restart_failure_time}")
       return true
     end


cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -338,7 +338,7 @@ def process_restarted_after_failure?(restart_failure_time, process_identifier)
   require 'time'
   begin
     start_time = process_start_time(process_identifier)
-    if start_time.nil? && (Time.parse(restart_failure_time).to_i < Time.parse(start_time).to_i)
+    if not start_time.nil? && (Time.parse(restart_failure_time).to_i < Time.parse(start_time).to_i)
```
vt0r: First, please prefer to use `!` over `not`, [due to possible issues with precedence](https://stackoverflow.com/questions/27429711/can-someone-explain-not-vs-in-ruby). Second, Rubocop is a fickle... uh... beast... that will also [fault you for checking to see if something is non nil](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/NonNilCheck), so maybe this should instead read as follows:

``` ruby
if start_time && (Time.parse(restart_failure_time).to_i < Time.parse(start_time).to_i)
```

cbaenziger:
FWIW this code should be deprecated to use `locking_resource` using `hadoop_service` is not recommended.

cookbooks/bcpc-hadoop/libraries/utils.rb
pu239ppy: The is what it originally was I believe

vt0r: @pu239ppy Ok so I found the two commits that modified this. The first one starts at this line:

https://github.com/bloomberg/chef-bach/commit/cf690ae7f5da6b9cf240b075bbbc01df8704c32a#diff-4301562bb3786f09a63001f8d6489b37L382

And the second one is the one you linked, at this line:

https://github.com/bloomberg/chef-bach/commit/a3c0d112c3a29c9d17096933fb4c7c956f2079ea#diff-4301562bb3786f09a63001f8d6489b37R341

It seems to me that simply making this PR change it to the following (same as my comment says above) will do exactly what we want, restoring the original condition as well:

``` ruby
if start_time ...
```

As for the `locking_resource` modification, I do agree it should be moved, but I'd say that's far out of scope of this PR.

PR #1114, opened 2018-02-21
Craft some artisanal hbase custom resources

### LINKED ISSUE #1106: codify HBase namespace tenancy policies
For setting up tenancies like ACLs, (and quotas in the future)

Some permissions:

(Updated)
* `RCWAX` for the actual namespace
* `RX` to `SYSTEM:CATALOG` for Phoenix
* `RX` to `hbase:meta`

aespinosa:
Allows us to create sort of self-checking tenant recipes

Fixes #1106


cbaenziger:
I am pretty strongly opposed to more `hbase shell` work this like the `recipe[bach-hadoop::hdfs_user_directories]` (and groups analog) need to be Oozie workflows in Java. The time to execute these shell actions on clusters with more than a trivial number of tenants is in the tens of minutes. (Launching `hbase shell` takes 46 seconds -- to do nothing!)

Though a lot of work looks to have gone in here, we are at a state where throwing away code to do the right thing is something the team seems to have painfully adopted; please refactor to use the HBase Java API's and ideally from a cluster aware scheduler like Oozie.

aespinosa:
will make a new PR

PR #1115, opened 2018-02-22
Oozie Client Library Cookbook

ronny-macmaster:
This adds a cookbook that contains a simple oozie library client.
It doesn't search for a running oozie host like the helper methods in the smoke tests, but that can be added if necessary.


ronny-macmaster:
updated in the backup pr, #1226

PR #1119, opened 2018-02-23
HBase Whitelist Coprocessors and Use locking_resource

cbaenziger:
This updates HBase to use [HBASE-16700](https://issues.apache.org/jira/browse/HBASE-16700) to allow for coprocessor whitelisting (or as it defaults, to disallow table coprocessors unless on the daemon's classpath).

Also corrected is HBase now uses locking_restart to arbitrate restarts of HBase daemons instead of the now removed `hadoop_service` definition.


cbaenziger:
Tested on a VM cluster, configs are there and cluster builds happily with a running HBase.

RiteshVajariya:
@amithkanand pl review.

PR #1128, opened 2018-03-01
Update service name for kafka broker

http-418:
bcpc_kafka was configured to use a non-existent service name for
Kafka when updating a ulimit.

This patch uses the correct service resource name from the upstream
Kafka cookbook.


cbaenziger:
Looks reasonable!

PR #1131, opened 2018-03-05
Option to enable/disable Jolokia agent to collect JMX stat

bijugs:
Since ``Jolokia`` agent is used to collect JMX stats from ``kafka``, want to use the same for the ``ZooKeeper`` processes on the ``Kafka`` clusters. By default enabling ``Jolokia`` agent is set to ``false`` so that ``jmx-agents`` will be used. When enabled, ``jmx-agents`` need to be [disabled](https://github.com/bloomberg/chef-bach/blob/56c6c208629e34c9a778bad75e489a360bcbcaaf/cookbooks/bcpc-hadoop/attributes/default.rb#L38).

*Note:* Not tested yet. Raised this to get early feedback.

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/zookeeper_impl.rb
+++ b/cookbooks/bcpc-hadoop/recipes/zookeeper_impl.rb
@@ -25,6 +25,11 @@ link "/usr/bin/zookeeper-server-initialize" do
   to "/usr/hdp/current/zookeeper-client/bin/zookeeper-server-initialize"
 end
 
+#Install jolokia's jvm agent to node['bcpc']['jolokia']['path']
+if node[:bcpc][:jolokia][:enable] == true
+  include_recipe 'bcpc-hadoop::jolokia'
+end
+
 template "#{node[:bcpc][:hadoop][:zookeeper][:conf_dir]}/zookeeper-env.sh" do
   source "zk_zookeeper-env.sh.erb"
   mode 0644
--- a/cookbooks/bach_repository/attributes/default.rb
+++ b/cookbooks/bach_repository/attributes/default.rb
@@ -76,6 +76,9 @@ default['bcpc']['hadoop']['java'] = "/usr/lib/jvm/java-8-oracle-amd64"
 default['bach']['repository']['jmxtrans_agent']['download_url'] = 'https://github.com/jmxtrans/jmxtrans-agent/releases/download/jmxtrans-agent-1.2.5/jmxtrans-agent-1.2.5.jar'
 default['bach']['repository']['jmxtrans_agent']['checksum'] = 'd351ac0b863ffb2742477001296f65cbca6f8e9bb5bec3dc2194c447d838ae17'
 
+# jolokia-agent (https://github.com/rhuss/jolokia)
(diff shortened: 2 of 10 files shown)


http-418:
Looks good to me.

cookbooks/bcpc-hadoop/recipes/zookeeper_impl.rb
```diff
@@ -25,6 +25,11 @@
   to "/usr/hdp/current/zookeeper-client/bin/zookeeper-server-initialize"
 end
 
+#Install jolokia's jvm agent to node['bcpc']['jolokia']['path']
+if node[:bcpc][:jolokia][:enable] == true
+  include_recipe 'bcpc-hadoop::jolokia'
```
cbaenziger: This looks duplicated with [bcpc_kafka::zookeeper_server](https://github.com/bloomberg/chef-bach/blob/8952764c95cf77063f2303a0b08231a4599fd1a3/cookbooks/bcpc_kafka/recipes/zookeeper_server.rb#L18)?

bijugs: Thanks @cbaenziger . Updated [bcpc_kafka::zookeeper_server](https://github.com/bloomberg/chef-bach/blob/8952764c95cf77063f2303a0b08231a4599fd1a3/cookbooks/bcpc_kafka/recipes/zookeeper_server.rb#L17)

cbaenziger:
Thanks Biju; please update once tested.

vt0r:
Do we need to use `force_override` in the 3-ish places I think I saw it being used? Shouldn't `force_default` suffice?

bijugs:
Updated PR to download ``Jolokia`` agent jar instead of using ``maven``.

http-418:
I would rather not add anything new to the "bins" tarball.  In a perfect world, the "bins" tarball would not exist.  It is one of the major obstacles to getting rid of "bootstrap" hosts and using shared infrastructure.

Besides, Jolokia is itself distributed as a jar file via maven central.  The maven cookbook is the natural and obvious way to distribute it to our hosts.  Why would we go out of our way to re-host this and do it in an awkward way?

bijugs:
@http-418 , as discussed offline, the code to re-host was done due to concerns raised about not being able to build VM clusters being a proxy which seem to be the cause of [this](https://github.com/bloomberg/chef-bach/commit/8952764c95cf77063f2303a0b08231a4599fd1a3#diff-ab0bd0575946adf51413dd40f07f4324L28) removal earlier. Once we are able to build a VM cluster behind the proxy with out re-hosting Jolokia jar, will remove the last commit from the PR.

http-418:
It appears the hdfsdu PR added a lot of the necessary SSL stuff for maven:
https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/maven_config.rb

I'm still not sure how it works -- are SSL certs downloaded from the bootstrap?

cbaenziger:
The code will setup Maven using the SSL certs if available. (E.g. as we setup on the bootstrap.) But today, no Maven commands are run on the cluster nodes, so no infrastructure yet exists to ship the certs from bootstrap to cluster nodes for them to run Maven.

vt0r:
Closing in favor of #1149

PR #1009, opened 2017-10-05
Reduce reliance on node search

pu239ppy:
Where ever possible remove node search and replace with a custom search that does lookups against cluster.txt

# Changes
- cluster.txt is the source of node information
- Node ID predefined in cluster.txt
- bcpc_8bit_node_number just refers to data from cluster.txt
- rewrote large parts of functionality that depended on node search output
- MySQL since we no longer rely on node search, we now have no idea if any heads are already running MySQL or not.  The only way to do so is ask.  If we find a quorum we join it if we do not we bootstrap, this is a potentially destructive operation as nodes of a preexisting quorum that contain data may simply be down, but is good enough for VM build purposes and is controlled by `node['bcpc']['mysql']['bootstrap_on_error']`

# Mode of operation
ClusterDef class can either access a remote web repository to get cluster.txt (when running in chef context and has access to node attributes) or access cluster.txt locally.  It relies on cluster_data's `BACH::ClusterData::repo_dir` to tell it where to find cluster.txt, when attempting to read cluster.txt locally.

# Assumptions
The web server functionality has to be provided early on in the chef-client run on the bootstrap as later parts (keytab generation rely on this)

# Breaking
This is a BIG change proceed with caution.

PR #1141, opened 2018-03-13
Fix node_number referencing for YARN

aespinosa:
We transform node_id -> node_number in
https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/libraries/utils.rb#L202.


cbaenziger:
@aespinosa I'm not groking the reason for the change from the long `node_number` to short `node_id` is this due to #1009?

aespinosa:
> Re: `node_number`
yes. This is for #1009 

We seem to use them interchangeably.  But right now for the yarn configs to work it should be `node_number` since that's how the hash is being populated as referenced in the link earlier.

> Re: tag

I wanted to track what additional changes we need for #1009 to work

cbaenziger:
Thanks for chatting @aespinosa; as discussed, let us keep labels for long-term state tracking and technologies (since labels are long-term creations) for deployment help. For tracking PR relationships, it seems referencing the issue should give us a nice list by going to the original PR and seeing what other PR's referenced it.

PR #1143, opened 2018-03-14
Reindex by hand

aespinosa:
by setting the update interval to every 1 node object, `c-a-r` runs fast and doesn't wait very long to get a correct index :O

To update the chef-server:

1.  update `/etc/chef-server/chef-server.rb`
2.  run `chef-server-ctl reconfigure`
3.  restart solr `chef-server-ctl restart chef-solr`

Downsides: since indexing (not reindexing) happens more often, the chef-server sometimes responds with a 500.  Saw it once while fiddling with my VM cluster.

### DIFF
--- a/lib/chef_node.rb
+++ b/lib/chef_node.rb
@@ -114,7 +114,9 @@ module BACH
             # iteration.
             elsif i == 179
               raise "Did not find indexed node for #{entry[:fqdn]} " \
-                "after #{timeout} secs! Reindex by hand or wait again"
+                "after #{timeout} secs! Reindex by hand or wait again.\n" \
+                'WARNING: Reindex is a dangerous cluster operation. Refer '\
+                'to runbooks for details.'
             end
             sleep 1
           end
--- a/lib/chef_node.rb
+++ b/lib/chef_node.rb
@@ -108,11 +108,13 @@ module BACH
             puts "Found #{entry[:fqdn]} in search index"
             return
           else
-            # the #times method counts up, not down, so i == 179 on the 180th iteration.
-            reindex_chef_server if i == 179
-
-            if i % 60 == 0
+            if i % 61 == 0
               puts "Waiting for #{entry[:fqdn]} to appear in Chef index..."
+            # the #times method counts up, not down, so i == 179 on the 180th
+            # iteration.
+            elsif ii == 179
+              raise "Did not find indexed node for #{entry[:fqdn]} " \
+                "after #{timeout} secs! Reindex by hand or wait again"
             end
             sleep 1
           end
--- a/lib/chef_node.rb
+++ b/lib/chef_node.rb
@@ -108,11 +108,11 @@ module BACH
             puts "Found #{entry[:fqdn]} in search index"
             return
           else
-            if i % 61 == 0
+            if i % 60 == 0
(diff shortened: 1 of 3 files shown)


setup_chef_server.sh
```diff
@@ -46,6 +46,10 @@ else
   printf "chef_server_webui['enable'] = false\n" >> /etc/chef-server/chef-server.rb
   printf "nginx['enable_non_ssl'] = false\n" >> /etc/chef-server/chef-server.rb
   printf "nginx['non_ssl_port'] = 4000\n" >> /etc/chef-server/chef-server.rb
+  # Configure Solr to index right away when we a new node.  
+  # Reference: https://docs.chef.io/config_rb_server.html#opscode-solr4
+  # Called opscode_solr4 in chef-server 12+
+  printf "chef_solr['max_commit_docs'] = 1\n" >> /etc/chef-server/chef-server.rb
```
vt0r: Nice. [The default](https://docs.chef.io/server_tuning.html#update-frequency) is `1000` or 60 seconds, meaning it's probably always 60 seconds for us.

I think we add objects infrequently enough for this value to make sense, and I can't come up with any downsides at the moment...

MEME THAT DESCRIBES MY FEELINGS

lib/chef_node.rb
```diff
@@ -108,11 +108,13 @@ def reindex_and_wait(entry)
             puts "Found #{entry[:fqdn]} in search index"
             return
           else
-            # the #times method counts up, not down, so i == 179 on the 180th iteration.
-            reindex_chef_server if i == 179
-
-            if i % 60 == 0
+            if i % 61 == 0
```
vt0r: Why `61` here?

lib/chef_node.rb
```diff
@@ -108,11 +108,13 @@ def reindex_and_wait(entry)
...
               puts "Waiting for #{entry[:fqdn]} to appear in Chef index..."
+            # the #times method counts up, not down, so i == 179 on the 180th
+            # iteration.
+            elsif ii == 179
```
vt0r: I don't think `ii` is set in this file

lib/chef_node.rb
aespinosa: oops. fixed

lib/chef_node.rb
aespinosa: oops again

setup_chef_server.sh
aespinosa: finished chef-client runs counts as updates to the node objects as well if i understood the docs correctly.

amithkanand:
**WARN**: Manually re-indexing `chef-server` can still break the cluster.

aespinosa:
Done adding the warnings

setup_chef_server.sh
vt0r: How often were you able to reproduce the 500 errors? Does setting this to maybe `10` or `20` alleviate that somewhat? What's the highest value we can set that to where the result will appear in Solr within about 5 seconds or so?

aespinosa: Talked with @vt0r to leave as is and see how it fares in our deployments.

The worst case from 500 errors would be chef runs aborting.  We'll look at how chef runs fare ad adjust the interval accordingly.

PR #1146, opened 2018-03-15
Add node id to kafka_cluster.txt

bijugs:
Add node id to ``kafka_cluster.txt`` so that latest changes to [parse_cluster_txt](https://github.com/bloomberg/chef-bach/blob/2a713e62996c1b17c37b8e9c943503052c39d673/lib/cluster_data.rb#L133) will function correctly for ``kafka`` cluster builds.


bijugs:
@pu239ppy , mind taking a look.

pu239ppy:
Looks correct, also if you wanted something not to have an id you want a space between `-` and hostname

bijugs:
Thanks @pu239ppy .

PR #1149, opened 2018-03-21
Rollup of changes from dev Kafka

http-418:
This PR contains three minor changes from our discoveries in our dev Kafka environment, and one larger change to simplify VM builds using maven

The minor changes:
- Make Kafka's broker configuration world-readable, as our tenants have repeatedly asked us
- Update the jute.maxbuffer value for the ZK client in Kafka brokers
- Correct Jolokia configuration for Zookeeper nodes

The larger change:

This PR uploads all local CA certificates from the Bootstrap to the Chef server, then has clients download them.  This makes it possible for maven and other tools to pass through a MitM proxy successfully.

This PR includes the changes from Biju's PR #1131, but now with the SSL support needed to avoid adding new things to `bins`.


vt0r:
Supersedes #1131

PR #1158, opened 2018-04-04
Enable table level region balancing

amithkanand:
This PR adds back table level region balancing.


vt0r:
Amit has verified this setting is compatible with the stochastic load balancer.

PR #1164, opened 2018-04-09
Hive CRUD smoke tests

rashidaligee:
Added Hive and Hive2 tests which covers:
1. Oozie Hive and Hive2 actions
2. Creation of Database, Tables in Hive using JDBC and thrift server.
3. Insertion of data into Hive table,
4. Clean up of test data once test is completed.
5. Run tests in parallel.


cookbooks/bcpc-hadoop/recipes/smoke_test_wrapper.rb
```diff
@@ -80,6 +80,17 @@
   user test_user
 end
 
+hive_jdbc_url = 'jdbc:hive2://'
+zookeeperQuorum = node[:bcpc][:hadoop][:zookeeper][:servers].map { |s|
```
cbaenziger: Please check with rubocop. Particularly multiline `{}` blocks are discouraged for `do` blocks instead.

cookbooks/smoke-tests/recipes/oozie_smoke_test.rb
```diff
@@ -92,18 +97,33 @@
...
+execute "upload hive_smoke_test to #{workflow_path}" do
+  command "hdfs dfs -copyFromLocal -f #{cache_dir}/hive_smoke_test.hql \
+  #{workflow_path}"
+  user test_user
+  not_if "hdfs dfs -test -f #{workflow_path}/hive_smoke_test.hql",
+         :user => test_user
+end
+
+execute "upload hive-site to #{workflow_path}" do
+  command "hdfs dfs -copyFromLocal -f /etc/hive/conf/hive-site.xml \
```
cbaenziger: I understand `copyFromLocal` to fail if he file already exists? How is this made idempotent?

cookbooks/smoke-tests/templates/default/hive_smoke_test.hql.erb
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE ${HIVE_SMOKE_TEST_DB};
+USE ${HIVE_SMOKE_TEST_DB};
+CREATE TABLE IF NOT EXISTS ${HIVE_SMOKE_TEST_TBL}(firstName string, lastName string);
+INSERT INTO TABLE ${HIVE_SMOKE_TEST_TBL} VALUES ('firstName1', 'lastName');
+
+SELECT firstName FROM ${HIVE_SMOKE_TEST_TBL};
+
+DROP TABLE ${HIVE_SMOKE_TEST_TBL};
```
cbaenziger: Should this be a prepare step to ensure all's clean at the start of the run incase the last test failed before cleaning up?

cookbooks/smoke-tests/recipes/oozie_smoke_test.rb
rashidaligee: the `-f` option overrides the existing file.

cookbooks/smoke-tests/templates/default/hive_smoke_test.hql.erb
rashidaligee: The intent is that script should only drop what it creates. As we have intentions to run these tests in productions, I would avoid to blind delete/drop statements.

cookbooks/smoke-tests/recipes/oozie_smoke_test.rb
cbaenziger: Cool! Then should line 104's `not_if` continue to exist?

cookbooks/smoke-tests/templates/default/hive_smoke_test.hql.erb
cbaenziger: While I certainly advocate not wiping all databases, the drop on line 8 seems like it'd be better to move to line 1. That way we can investigate the test results after running and the test should not be influenced by the last test, if it were to have failed.

cbaenziger:
Though not your sins, please run rubocop one more time as I still see some stuff in what you are touching that could be cleaned up.

As to the fork to run the CRUD test twice, simultaneously; why run the same test twice on two differently named databases?

RiteshVajariya:
Looks good. Reviewed.

PR #1170, opened 2018-04-11
Changes to enable users to use non default ZK port for Kafka

bijugs:
Currently ``2181`` the default ``ZK`` port is assumed.


RiteshVajariya:
@http-418 reviewed it. Merging it.

PR #1171, opened 2018-04-12
Measure the root queue metrics

aespinosa:
This is how the YARN dashboard and /ws/v1/cluster/metrics endpoint
reports their cluster-wide resource usage.


cbaenziger:
Hi @aespinosa, please refactor to grab all queues under root rather than only root and the users parent queue (we have others too).

cbaenziger:
Does the `%q#%` break the metrics into a folder structure?

aespinosa:
screenshot-2018-4-26 graphite composer

folder structure is `queue_name.user`. If there is none it defalt to `null` (like total for the queue)

cbaenziger:
Awesome! For reference of users, `null` is the total for that queue-level as Allan explains it to me.

PR #1173, opened 2018-04-13
refactored datanode and journalnode ports into attributes.

rashidaligee:
This item was somehow landed in Ambari Epic, instead of moving it around, just implemented it.

### DIFF
--- a/cookbooks/bcpc-hadoop/attributes/hdfs.rb
+++ b/cookbooks/bcpc-hadoop/attributes/hdfs.rb
@@ -13,10 +13,14 @@ default['bcpc']['hadoop']['namenode']['xmx']['max_size'] = 1024
 # # set to nil to calculate dynamically based on available memory
 default['bcpc']['hadoop']['namenode']['xmn']['max_size'] = 128
 default['bcpc']['hadoop']['namenode']['xmx']['max_ratio'] = 0.25
-default['bcpc']['hadoop']['namenode']['jmx']['port'] = 10111
+default['bcpc']['hadoop']['namenode']['jmx']['port'] = 10_111
 default['bcpc']['hadoop']['namenode']['rpc']['port'] = 8020
-default['bcpc']['hadoop']['namenode']['http']['port'] = 50070
-default['bcpc']['hadoop']['namenode']['https']['port'] = 50470
+default['bcpc']['hadoop']['namenode']['http']['port'] = 50_070
+default['bcpc']['hadoop']['namenode']['https']['port'] = 50_470
+
+default['bcpc']['hadoop']['datanode']['ipc']['port'] = 50_020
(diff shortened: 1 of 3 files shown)


cookbooks/bcpc-hadoop/attributes/hdfs.rb
```diff
@@ -13,10 +13,14 @@
 # # set to nil to calculate dynamically based on available memory
 default['bcpc']['hadoop']['namenode']['xmn']['max_size'] = 128
 default['bcpc']['hadoop']['namenode']['xmx']['max_ratio'] = 0.25
-default['bcpc']['hadoop']['namenode']['jmx']['port'] = 10111
+default['bcpc']['hadoop']['namenode']['jmx']['port'] = 10_111
```
cbaenziger: I don't think we should be using JMX port anymore? (We use jmxtrans_agent to send data so we don't need to network poll or expose JMX.)

cookbooks/bcpc-hadoop/recipes/hdfs_site.rb
```diff
@@ -211,13 +211,13 @@
       jn_properties =
         {
          'dfs.journalnode.rpc-address' =>
-           node[:bcpc][:floating][:ip] + ':8485',
+           node[:bcpc][:floating][:ip] + ':' + node['bcpc']['hadoop']['journalnode']['rpc']['port'].to_s,
```
vt0r: For this and other concatenated strings below, could we use interpolation instead? It's arguably *slightly* easier to read, the `.to_s` happens implicitly, and it's also slightly faster in benchmarks (interpolation is faster already, plus it now becomes a single string).

Example:
``` ruby
"#{node['bcpc']['floating']['ip']}:#{node['bcpc']['hadoop']['journalnode']['rpc']['port']}",
```

cookbooks/bcpc-hadoop/attributes/hdfs.rb
rashidaligee: I tried to please rubocop here. My change is just to add `_` in a number.

rashidaligee:
- Interpolation is used to concatenate attributes.

cookbooks/bcpc-hadoop/recipes/hdfs_site.rb
```diff
@@ -262,8 +262,8 @@
 
          'dfs.namenode.shared.edits.dir' =>
            'qjournal://' +
-           zk_hosts.map{ |s| float_host(s[:fqdn]) + ":8485" }.join(";") +
-           '/' + node.chef_environment,
+           zk_hosts.map { |s| float_host(s[:fqdn]) + ':' + node['bcpc']['hadoop']['journalnode']['rpc']['port'].to_s }
```
cbaenziger: Can this be an interpolation too?

PR #1174, opened 2018-04-13
Enable users to use non chef-bach ZooKeeper for Kafka

bijugs:
Users can set ``default[:kafka][:zookeeper_connect] `` with the ``ZK`` connection string in the format ``zkhost:port,zkhost:port,zkhost:port/zknode`` where ``zknode`` is optional if a non ``chef-bach`` ZooKeeper quorum need to be used with ``Kafka``.


cbaenziger:
Mind rubocop'ing the changes @bijugs ?

http-418:
These changes are already in use on a cluster.  I want to move it to a release tarball rather than local changes, so linter changes will have to wait for another day.

PR #1177, opened 2018-04-16
Namenode Ids instead of nodenumbers

rashidaligee:
used node_id instead of node_number which was breaking the ambari cookbook


cbaenziger:
I can confirm this is breaking master as I just hit it.

cbaenziger:
We should not be using node_number since `hdfs-site.xml` already uses node_id and that is what controls.

rashidaligee:
@vt0r I'm not updating the attributes, I'm using it, if this what you are asking.

PR #1183, opened 2018-04-19
Revert "Enable users to use non chef-bach ZooKeeper for Kafka"

http-418:
Reverts bloomberg/chef-bach#1174

These changes turned out to be totally unnecessary.

`node[:kafka][:broker][:zookeeper][:connect]` can be overriden directly in the environment, without a new attribute or any copying inside the chef code.


bijugs:
@http-418 , since this is not required anymore can we close this?

PR #1187, opened 2018-04-24
Remove hypervisor's dependency on the Gemfile

aespinosa:
Tested and confirmed that the cluster.txt generation still works.

Scripts on the hypervisor are supposed to be designed to be minimal
anyway.

### DIFF
--- a/vm-to-cluster.sh
+++ b/vm-to-cluster.sh
@@ -11,16 +11,4 @@ REPO_DIR="`dirname ${BASH_SOURCE[0]}`"
 cd $REPO_DIR
 . ./proxy_setup.sh
 
-# Iterate over a list of "<dpkg name> <gem name>" necessary for the Gemfile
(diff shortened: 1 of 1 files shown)


cbaenziger:
Do rspecs still work? They largely depend on our gemfile. You can pull my chef_12_rebase branch if it's easier to test there.

aespinosa:
yes.  the only hard dependency outside of chefdk is `simplecov`.

http-418:
I am not 100% sure we won't end up doing this _again_, later, to get some special gem, but if this works for now, it is better to have fewer hypervisor deps than more.

cbaenziger:
Allan how did you test this on a fresh Ubuntu machine? I don't quite grok how you can remove these dependencies and still run rspec/chefspec or bundler.

aespinosa:
i only ran specs related to the hypervisor.

i have separate gem_home directories for my cookbooks and bootstrap.

Issue #1144, opened 2018-03-15
Add support for node numbers > 8 bits in cluster.txt


amithkanand:
With latest change to get cluster information statically, it doesn't allow node numbers greater than 8 bits. We need to add support to `cluster.tx`t to allow node numbers that are greater than 8 bit.

pu239ppy:
Why do you need node numbers greater then 8 bits?  as of #1009 Worker nodes no longer receive a node ID, which mean your entire 8 bit ranger (does 0 constitute a valid ID) may be used for head nodes only.  So unless you plan to have more then 8 bits of Head nodes, this is unnecessary.  Furthermore I was under the impression that 8bit IDs was a hard requirement from zookeeper.

aespinosa:
Do you mind expanding the impact of changing node numbers at namenodes and resource managers during the migration?

1.  namenode A (standby) has node_number > 8 bit
2.  chef namenode A.  node_number of A will become < 8 bit . namenode A points to namenode B via namenode B's new < 8bit node_number.  Can namenode A still connect to namenode B and vice versa?

Richard's currently working on testing the upgrade path so we'll know what to expect in the upgrade path.

http-418:
We have to preserve existing node numbers, because otherwise we will experience downtime during the upgrade.

Existing node numbers are often greater than 8 bits, and then they are modulo'd down to the 8 bit range for use by zookeeper.

All future nodes/clusters will have simple node numbers < 255, for simplicity.  But our existing clusters have to retain their current numbers.

pu239ppy:
@http-418 if you have an existing node number > 8bit it will be converted to number mod 8 for zookeeper anyway, why not take that number and use it as a new node number.  The only caveat here is that you will have to suffer restarts for HDFS and YARN that use the full number as opposed to number mod 8.  However the benefit is that you will never accidentally conflict with another nodes node number while picking a new number, and you do not have to check this manually by picking a number and running mod 8 on existing numbers.

bijugs:
``node number > 8bit it will be converted to number mod 8 for `` -> is there a possibility of creating duplicate node id or we are handling that as well?

pu239ppy:
@bijugs C-A-R will warn you that your `node_number mod 8` conflicts with other `node_number mod 8`, however as I said above if you accept the tradeoff of having to re-converge your entire node at least as far as HDFS NN/JN/DN and YARN RM/NM daemons go, then you will not have to deal with this issue

[...]

PR #1188, opened 2018-04-26
Correctly handle node numbers over 255 in ZK templates

http-418:
Applying v3.4.1 to an existing hardware cluster will fail catastrophically due to disagreements between Zookeeper `myid` and greater-than-255 node numbers.

This change prevents the catastrophe.   It is a partial fix for issue #1144


bijugs:
This change looks good to me.
Based on our experience
- What is the recommended way to deploy v3.4 to existing clusters in terms of ``node id``. We update ``cluster.txt`` with existing node id.
- Looking at the code comment in [bcpc_8bit_node_number](https://github.com/bloomberg/chef-bach/blob/2a713e62996c1b17c37b8e9c943503052c39d673/cookbooks/bcpc/libraries/node_numbers.rb#L8-L11), the code may need update to check duplicates in the generated node number unless it is done elsewhere.

PR #1190, opened 2018-04-27
Clean-up unused whisper metric directories

cbaenziger:
This cleans up the stale Graphite directory nodes left behind when machines are removed from clusters, or tables are deleted out of HBase for example.


aespinosa:
should we sync this to run only on when it needs based on the retention period like in #1022

aespinosa:
Note: this is not enough even if we override `node[bcpc][graphite][data][retention] = 1` to stop the TireFire that is the Region and Table metrics,

PR #1194, opened 2018-04-30
Setting HBase GC log file size and number of files for retention

bijugs:
Changes to use GC log rotation, set log file size and number of ``HBase`` GC files for retention.


cbaenziger:
Please also do for HDFS, YARN and other GC components.

PR #1198, opened 2018-05-03
Deploy custom files/jars to the cluster

amithkanand:


cookbooks/bach_deployment/attributes/default.rb
```diff
@@ -0,0 +1,12 @@
+force_default['bach']['deploy']['appfile']['data'] = {
+  appfoo: {
+    repo_url: get_binary_server_url()
```
aespinosa: This is an implicit dependency to the `bcpc` cookbook .  Let's make this to some default URL as well.

PR #1200, opened 2018-05-07
Ambari metrics Installation (embedded mode) in BACH

rashidaligee:
This pull request enables users to install ambari metrics components into bach cluster.


cookbooks/ambari_metrics/recipes/ambari_metrics_grafana_dashboard.rb
```diff
@@ -0,0 +1,194 @@
...
+  port node['ams']['metrics_grafana']['port']
+  admin_user node['ams']['metrics_grafana']['admin_user']
+  admin_password node['ams']['metrics_grafana']['admin_password']
+  dashboard(
+    source: 'grafana-ams-hbase-home',
+    overwrite: true
+  )
+end
+
+grafana_dashboard 'ams-hbase-misc' do
```
RiteshVajariya: create a loop on dashboard list to import them into grafana.

cbaenziger:
Please use an array of dashboard URLs to deploy (e.g. local files or URLs). These dashboards should be populated by the various installation scripts so we only install relevant dashboards per the configuration of that cluster.

cbaenziger:
Can we render `ams-site.xml.erb` like we do our normal site files to allow for attributes to override when necessary?

RiteshVajariya:
Clay is right, values of config params should not be hardcoded into templates if possible.  Look at hbase-site.xml as an example -- a base template takes a hash of keys and values and formats them as xml

cookbooks/ambari_metrics/templates/default/ams-grafana.ini.erb
```diff
@@ -0,0 +1,258 @@
...
+protocol = http
+# The ip address to bind to, empty will bind to all interfaces
+;http_addr =
+
+# The http port  to use
+;http_port = 3000
+http_port = 3000
+
+# The public facing domain name used to access grafana from a browser
+;domain = localhost
```
cbaenziger: Should not this be overrideable for using `node[:bcpc][:domain]`

cookbooks/ambari_metrics/templates/default/ams-grafana.ini.erb
```diff
@@ -0,0 +1,258 @@
...
+
+# Redirect to correct domain if host header does not match domain
+# Prevents DNS rebinding attacks
+;enforce_domain = false
+
+# The full public facing url
+;root_url = %(protocol)s://%(domain)s:%(http_port)s/
```
cbaenziger: Here we likely want to be able to point to a VIP?

cookbooks/ambari_metrics/templates/default/ams-grafana.ini.erb
```diff
@@ -0,0 +1,258 @@
...
+# https certs & key file
+;cert_file =
+;cert_key =
+cert_file = /etc/ambari-metrics-grafana/conf/ams-grafana.crt
+cert_key = /etc/ambari-metrics-grafana/conf/ams-grafana.key
+
+#################################### Database ####################################
+[database]
+# Either "mysql", "postgres" or "sqlite3", it's your choice
+;type = sqlite3
```
cbaenziger: How do we make this HA to not rely on sqlite3 or what gets stored in this database?

rashidaligee: The ams grafana is not used and will be replaced with in-house grafana.

PR #1205, opened 2018-05-09
Split gemfiles

aespinosa:
First phase of having bootstrap, chef nodes and hypervisor having different gem dependencies. will pave way to allow faster cluster builds (not installing `berkshelf` dependencies which take forever to build), prepare for chef 13 (no more `/opt/chef/Gemfile`), and having easier `Gemfile.lock` to debug and reason about.

* bootstrap.gemfile
* nodes.gemfile
* hypervisor.gemfile

Tested `cluster-assign-roles.sh`, `repxe-host.sh`, and `wait-for-host.sh`  (needed by `tests/automated_install.sh`) and verified that they still work

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,11 @@ rvm:
   - 2
 bundler_args: --without cookbooks
 cache: bundler
-gemfile: Gemfile
+gemfile: gemfiles/bootstrap.gemfile
 before_install:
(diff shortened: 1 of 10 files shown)


aespinosa:
Done on `CONTRIBUTING.md`. There is no need to update `create-release.rb` since there is no shellscript that sets up its environment.

http-418:
Looks great to me.  I would like to merge this ASAP.

PR #1206, opened 2018-05-10
Chef 12 Server for Chef-BACH

cbaenziger:
This supports us standing up and running a Chef 12 Server for Chef-BACH. This builds cleanly for a Hadoop cluster at this time.


cookbooks/bcpc/recipes/bach_repository_wrapper.rb
```diff
@@ -0,0 +1,11 @@
+node.force_default['bach']['repository']['gem_server'] = node['bcpc']['bootstrap']['server']
```
aespinosa: why use `force_default` here? 

`force_default` is supposed to be reserved for internal wrappers so that it has the highest precedence.  having it in `chef-bach` will confuse the precedence in that part.

cookbooks/bcpc-hadoop/recipes/opentsdb.rb
```diff
@@ -34,5 +34,3 @@
   "hbase/_HOST@#{krb_realm}"
 node.force_default['bach_opentsdb']['hbase_region_princ'] =
   "hbase/_HOST@#{krb_realm}"
-
-include_recipe 'bach_opentsdb'
```
aespinosa: this only makes the dependency implicit which is more confusing when debugging.

cbaenziger: This allows us to prevent accidental cyclical dependencies or unnecessary dependencies for Chef-BACH as we work to break bcpc up. As this is a wrapper recipe, we know that it needs the base cookbook and like most of our specific uses there is a role [BCPC-Hadoop-Head-OpenTSDB-Server](https://github.com/bloomberg/chef-bach/pull/1206/files#diff-db60cd2ab1f98652f0acda7d4127b893) that provides the ordering.

This was likely needed for simplifying rspec or chef-spec but is not necessary for the Chef12-Server work if you would still like me to revert the changes?

aespinosa: we shouldn't rely on the role ordering for dependencies. it's brittle, unversioned and implicit. it's hard to debug if you have unsynced roles in production.

if we have cyclical dependencies, it will get detected with `berks install` and get the error at the build-level. nothin having this will just move implicit cyclical dependencies in runtime which bites is in production.

cbaenziger: This actually fixes at least one circular dependency with `bach_repository` which berks was not catching.

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -481,3 +490,20 @@ def update_oozie_sharelib(host)
     Chef::Log.info("Sharelibupdate: Oozie server not running on #{host}")
   end
 end
+
+def cluster_nodes
```
RiteshVajariya: Isn't this duplicated by existing helpers from Lev's PR?

cookbooks/bcpc-hadoop/recipes/configs.rb
```diff
@@ -76,7 +76,7 @@
   end
 end
 
-# We need more maps!
+# incrase max_map_count
```
RiteshVajariya: typo

cookbooks/bcpc-hadoop/recipes/hbase_config.rb
```diff
@@ -58,15 +58,23 @@
...
   'hbase.master.hostname' => float_host(node[:fqdn]),
   'hbase.regionserver.hostname' => float_host(node[:fqdn]),
   'hbase.regionserver.dns.interface' =>
-      node["bcpc"]["networks"][subnet]["floating"]["interface"],
+    node["bcpc"]["networks"][subnet]["floating"]["interface"],
   'hbase.master.dns.interface' =>
-      node["bcpc"]["networks"][subnet]["floating"]["interface"],
-  'dfs.client.read.shortcircuit' => node["bcpc"]["hadoop"]["hbase"]["shortcircuit"]["read"].to_s
+    node["bcpc"]["networks"][subnet]["floating"]["interface"],
+  'hbase.regionserver.ipc.address' =>
```
RiteshVajariya: We should find out why these are marked as new -- wasn't this material merged into master in the past?

cookbooks/bcpc-hadoop/recipes/oozie.rb
```diff
@@ -219,7 +219,7 @@
     require 'digest'
     node['bcpc']['hadoop']['oozie_hosts'].each do |oozie_host|
       next unless oozie_running?(float_host(oozie_host['hostname']))
-      node.set['bcpc']['hadoop']['oozie']['sharelib_checksum'] =
+      node.override['bcpc']['hadoop']['oozie']['sharelib_checksum'] =
```
RiteshVajariya: Why are we overriding in a recipe?

Recipe attr levels already have precedence

cookbooks/bcpc-hadoop/recipes/zookeeper_impl.rb
```diff
@@ -4,7 +4,6 @@
 ::Chef::Resource::File.send(:include, BCPC::Utils)
 
 include_recipe 'bcpc-hadoop::zookeeper_packages'
-include_recipe 'bach_krb5::keytab_directory'
```
RiteshVajariya: Isn't this required?

cookbooks/bcpc-hadoop/spec/spec_helper.rb
```diff
@@ -48,23 +48,26 @@
 
 end
 
-berks = Berkshelf::Berksfile.from_file('Berksfile').install()
+Berkshelf.ui.mute do
```
RiteshVajariya: This is clever.

cookbooks/bcpc/recipes/bach_repository_wrapper.rb
```diff
@@ -0,0 +1,42 @@
+require 'socket'
```
RiteshVajariya: Please add comments to explain the broken cyclical dependency

proxy_setup.sh
```diff
@@ -23,8 +23,12 @@ if [ -n "$http_proxy" ]; then
     DOMAIN="$(domainname),"
   fi
 
+  # a string like 127.0.0.1, 10.0.100.3, ..."
   local_ips=$(ip addr list |grep 'inet '|sed -e 's/.* inet //' -e 's#/.*#,#')
-  export no_proxy="$(sed 's/ //g' <<< $local_ips)$(hostname),$(hostname -f),$DOMAIN10.0.100.,10.0.100.*,$no_proxy"
+  # a string like 10.0.2.*, 10.0.100.*, ..., 10.0.2., 10.0.100., ...
+  local_nets=$(sed -e 's/127[.0-9]*, //' -e 's/\.[0-9]*,/.*,/g' <<< $local_ips)$(sed -e 's/127[.0-9]*, //' -e 's/\.[0-9]*,/.,/g' <<< $local_ips)
```
RiteshVajariya: This is a really good example of why we might wish to abandon the multiple NICs in the default setup.

cookbooks/bcpc-hadoop/recipes/opentsdb.rb
aespinosa: the default solver doesn't detect circular dependencies. you would have to add `solver :ruby` (the molinillo gem) to be able to detect it.

cookbooks/bcpc-hadoop/metadata.rb
```diff
@@ -7,9 +7,7 @@
 version          '0.1.0'
 
 depends 'bcpc', '= 0.1.0'
-depends 'bach_krb5', '= 0.1.0'
-depends 'smoke-tests', '= 0.1.0'
-depends 'bach_opentsdb', '= 0.1.0'
```
aespinosa: i thought that `bcpc-hadoop` was like one of the ultimate wrapper cookbooks.  `bcpc-hadoop` should depend on `bach_opentsdb`. not the otherway around.

http-418: I'm not sure there's ever been a clear pattern of what depends on what.  They all depend on the same blob of mutable state, so it's hard to say.

cluster_assign_roles.rb
```diff
@@ -285,6 +293,58 @@ def install_kafka(target_nodes)
     end
   end
 
+  def permission_host(node)
```
aespinosa: Reference for the recommended way to use the old chef-server behaviour that we rely on:

https://blog.chef.io/2014/11/10/security-update-hosted-chef/

build_bins.sh
```diff
@@ -94,7 +96,7 @@ EOF
 # defaults.
 #
 pushd $DIR/vendor > /dev/null
-/opt/chefdk/bin/chef-client -z -r 'recipe[bach_repository]' -c $TMPFILE
+/opt/chefdk/bin/chef-client -z -r 'recipe[bcpc::bach_repository_wrapper],recipe[bach_repository]' -c $TMPFILE
```
aespinosa: i think we should do `include_recipe 'bach_repository'` inside `bcpc::bach_repository_wrapper` for reasons i posted in other comments.

bin/setup_chef_perms.rb
```diff
@@ -0,0 +1,27 @@
...
+Chef::Config.from_string(config, 'Modified knife.rb')
+
+options = {
+  client_name: 'admin',
+  raw_key: File.open('/etc/chef-server/admin.pem').read
+}
+
+rest = Chef::ServerAPI.new(Chef::Config[:chef_server_url])
+
+Chef::Node.list.each do |node|
```
aespinosa: Since we setup the ACLs to be like the old chef-server, is this still needed?

cluster_assign_roles.rb
cbaenziger: Thanks; we should file this as an issue unless this will not work to get us off the ground?

bin/setup_chef_perms.rb
cbaenziger: To allow the bootstrap host proper permissions I need this in [setup_chef_cookbooks.sh](https://github.com/bloomberg/chef-bach/pull/1206/files/73d22bfce559dacbb3ff7685655c228f06cd3b22#diff-95d29768b114a017d5a3307f0e57eb09R67)

lib/chef_node.rb
```diff
@@ -125,7 +125,22 @@ def reindex_and_wait(entry)
         raise "Did not find #{entry[:fqdn]} in Chef index after 180 seconds!"
       end
 
+      # If running on a machine with /etc/chef-server/admin.pem
+      # will use that user
       def ridley
+        # use admin account if available
```
aespinosa: Can you list operations where we need the pivotal/ admin user for the chef-server?

We should avoid this as possible as it keeps tying us on to having to run our own chef-server.

bootstrap.md
```diff
@@ -408,7 +408,7 @@ index d844783..9a9e53a 100644
 +      "include_source_packages": false
 +    },
      "chef_client": {
-       "server_url": "http://10.0.100.3:4000",
+       "server_url": "http://bcpc-bootstrap.bcpc.example.com:4000",
```
aespinosa: should be `https://bcpc-bootstrap.bcpc.example.com/organizations/theorg` now

cluster_assign_roles.rb
```diff
@@ -221,6 +222,8 @@ def install_bootstrap(target_nodes)
                            search: 'role:BCPC-Hadoop-Head')
 
     target_head_nodes.each do |node|
+      puts "#{node[:fqdn]}: Permissioning head node"
```
aespinosa: this should no longer be needed when we setup the permissions to look like the old chef-server right?

Plus this is only needed for VM clusters/ new clusters.

cookbooks/bcpc-hadoop/Gemfile
```diff
@@ -1,6 +1,16 @@
 ruby RUBY_VERSION
+if File.exist?('/opt/chef/Gemfile')
```
aespinosa: this file no longer exists. let's stop doing this.

cookbooks/bcpc-hadoop/recipes/oozie.rb
aespinosa: we should keep everything in chef-bach at `default` so that internal wrappers will work.

cookbooks/bcpc-hadoop/recipes/smoke_test_principal.rb
```diff
@@ -62,11 +62,10 @@
 end
 
 chef_vault_secret 'test_user_keytab' do
-  provider ChefVaultCookbook::Provider::ChefVaultSecret
   data_bag 'os'
   raw_data ( lazy {{ 'keytab' => node.run_state['test_user_base64_keytab']}})
   search '*:*'
-  admins "#{ nodes },#{ bootstrap }"
+  admins "#{ nodes },#{ bootstrap },admin"
```
aespinosa: this ties us to having running our own chef-server. let's try to keep interacting with the chef-server only through the organization's admin account (which is a normal user with an "admin-level" access for that organization only.

tests/automated_install.sh
```diff
@@ -132,7 +132,7 @@ vagrant ssh -c "cd chef-bcpc; source proxy_setup.sh; ./wait-for-hosts.sh ${VM_LI
 snapshotVMs "${SNAP_POST_PXE}"
 
 printf "#### Chef the nodes with Basic role\n"
-vagrant ssh -c "cd chef-bcpc; ./cluster-assign-roles.sh $BACH_ENVIRONMENT Basic"
+vagrant ssh -c "cd chef-bcpc; sudo ./cluster-assign-roles.sh $BACH_ENVIRONMENT Basic"
```
aespinosa: we just removed the need to `sudo` all bootstrap-related commands because it considered a bug. why did we put it back?

setup_chef_server.sh
```diff
@@ -36,25 +39,51 @@ pushd /home/vagrant/chef-bcpc/lib/cluster-def-gem > /dev/null
...
+  printf "nginx['non_ssl_port'] = 4000\n" >> /etc/opscode/chef-server.rb
   # we can take about 45 minutes to Chef the first machine when running on VMs
   # so follow tuning from CHEF-4253
-  printf "erchef['s3_url_ttl'] = 3600\n" >> /etc/chef-server/chef-server.rb
-  export NO_PROXY=${NO_PROXY-127.0.0.1}
+  printf "opscode_erchef['s3_url_ttl'] = 3600\n" >> /etc/opscode/chef-server.rb
+  export NO_PROXY=${NO_PROXY-127.0.0.1,$(hostname),$(hostname -f)}
   chef-server-ctl reconfigure
+
+  # Create Chef Admin User
```
aespinosa: We currently have the bootstrap as the "admin" user. Why are we using the pivotal user in the chef-server as the admin for the org?

If we want a separate admin account we should recreate it. So the bootstrap will now have 3 certs:


1.  pivotal (ishould be optional) - for the `chef-server-ctl` commands. we shouldn't touch this so we can move to a chef-server we don't manage someday.
2.  bootstrap node's client.pem - to converge the bootstrap node.
3.  bootstrap's knife client.pem - to do "admin-ish" operations on the organization like create databags, assign runlists, etc.

cookbooks/bcpc-hadoop/recipes/opentsdb.rb
cbaenziger: Restored to old behavior

cookbooks/bcpc-hadoop/libraries/utils.rb
cbaenziger: This must have been a merge turd as nothing calls it; removed

cookbooks/bcpc-hadoop/recipes/configs.rb
cbaenziger: Corrected

cookbooks/bcpc-hadoop/metadata.rb
aespinosa: then we should move out all the wrapper cookbooks out of bcpc-hadoop then

moving to roles just makes the dependency implicit.  which are then harder to debug.

http-418: Not only should we not be adding new implicit dependencies to roles, In a perfect world we would using roles entirely.

cookbooks/bcpc-hadoop/recipes/oozie.rb
cbaenziger: Corrected

cookbooks/bcpc-hadoop/recipes/zookeeper_impl.rb
cbaenziger: See circular dependency issue elsewhere. I will leave this recipe in `barch_krb5` for now but we can open a future PR if we want to refactor the bach_krb5 cookbook.

build_bins.sh
cbaenziger: So we use the wrapper recipe as a wrapper recipe for more than [`build_bins`](https://github.com/bloomberg/chef-bach/pull/1206/files#diff-241f80a9f414eb8c74ac43a71e8570e7R99). We also use `bach_repository_wrapper` in [`recipe[bcpc::chef_vault_download]`](https://github.com/bloomberg/chef-bach/pull/1206/files#diff-12a3fefb1f76b03d12b9f4e3d4636874) which does not then call `recipe[bach_repository::default]` but instead calls `recipe[bach_repository::gems]`. This would require us calling all of bach_repositroy every Chef run we download Vault. Also we use `recipe[bach_respository_wrapper]` when we then call [`recipe[bach_repository::apt]`](https://github.com/bloomberg/chef-bach/pull/1206/files#diff-89d187ed8768ffde26c53a9c5a96e17eR62) in the Chef-Server.

This is a wrapper before we call into any point in bach_respository -- it is not an entrypoint to `recipe[bach_respository::default]` as such it would be in appropriate to force a specific bach_repository entry point at this time.

cookbooks/bcpc-hadoop/Gemfile
cbaenziger: Removed

build_bins.sh
aespinosa: I guess this is another set of implicit dependencies that this PR introduces.  After this PR, most of us will forget that this relationship exists and will take a lot of time for us to debug what is going on.

cookbooks/bcpc/recipes/bach_repository_wrapper.rb
cbaenziger: This is wrapping `bach_repository`'s defaults; as such I can not use the `default` precedence can I?

aespinosa: using explicit recipe wrapping like having `include_recipe 'bach_repository::the_recipe_being_wrapped'` at the end of this recipe should make that work.

cluster_assign_roles.rb
cbaenziger: Issue #1240 submitted

lib/chef_node.rb
cbaenziger: Yes, we need a "human" account for interactive operations. We do not use the bootstrap server's identity in these updates to run things like knife, etc. for uploading roles, environment and cookbooks.

bootstrap.md
cbaenziger: Thanks, fixed that but ew, that file is still scary out of date sadly.

cluster_assign_roles.rb
cbaenziger: For the current process, if you are creating new databag entries this is needed.

cookbooks/bcpc-hadoop/recipes/smoke_test_principal.rb
cbaenziger: At this time there is only this admin account which I have created; I think?

tests/automated_install.sh
cbaenziger: Oh awesome! I figured we still had the Gem-hell and was not trying to solve that here. If the Gem-hell's fixed we can chmod the key and all should be good. I'll see where things are at.

setup_chef_server.sh
cbaenziger: I certainly agree! This is the plumbing to get us to 12 but it does not do 12 right; this simply gets us to a point we can leverage a 12 server. Let us design the correct permissioning in #1240. Copied your points there.

cbaenziger:
For reference, I have been testing this via the following build configuration:
```
# System needed
PATH=/opt/chefdk/embedded/bin:$PATH
http_proxy=...
https_proxy=...
# Chef-BACH specific
BACH_CLUSTER_PREFIX=rebase
BOOTSTRAP_VM_CPUs=2
BOOTSTRAP_VM_MEM=5048
CLUSTER_VM_EFI=false
CLUSTER_VM_CPUs=5
CLUSTER_VM_MEM=5048
```

I have needed to add a second worker VM due to recent smoke-test changes; that was done via adding the following line to `stub-environment/hadoop_cluster.txt`:
```+- bcpc-vm5 08:00:27:C0:FF:EE 10.0.101.15 - cobbler_profile bcpc.example.com role[BCPC-Hadoop-Worker],recipe[bcpc-hadoop::copylog]```

Issues I have seen are:
* I sometimes see issues with Gems failing to download when behind a proxy when setting up the bootstrap. Re-running `automated_install.sh` usually runs through fine.
* HBase RS and HBase Master sometimes are not noticed being started by Chef (e.g. it will error that they did not start in 600 seconds). I see this on Chef-11 based master as well. A rerun of `automated_install.sh`/`cluster-assign-roles.sh`/`chef-client` then runs through on that machine.

ronny-macmaster:
Hypervisor hosts need to be updated to use a modern version of the ChefDK (chefdk-2.5.3)

Issue #1049, opened 2017-12-20
OS Installation stalled


amithkanand:
While building a VM cluster I found out that OS installation doesn't move forward until you connect to it via serial console to watch the progress.  I am using `Vagrant 2.0.1` with `Virtualbox  5.2.2-119230`.  During my cluster build with 6 VMs I had to connect to each of them using the [virtualbox_serial_console.sh](https://github.com/bloomberg/chef-bach/blob/master/tests/virtualbox_serial_console.sh) script. Also, I got below message when I connected to the VM:
```
$ ./virtualbox_serial_console.sh bach-bcpc-vm4
/home/aanand59/chef-bcpc/virtualbox_env.sh: line 19: bc: command not found
ERROR: VirtualBox 5.2 is less than 4.3.x!
  Only VirtualBox >= 4.3.x is officially supported.
```

aespinosa:
seems like a virtualbox thing. happens in CentOS too https://www.spinics.net/lists/centos-devel/msg15834.html

http-418:
Judging from the mailing list post, we need to manually add a serial port to each vm in `vbox_create.sh`

For the version check, I have a fix in #1209

cbaenziger:
This looks related to #1228. Closing to have a single issue.

PR #1209, opened 2018-05-14
Use Ruby gem library to correctly evaluate vbox version

http-418:
Partial fix for #1049

### DIFF
--- a/virtualbox_env.sh
+++ b/virtualbox_env.sh
@@ -14,9 +14,10 @@ if [[ -z "$VBM" ]]; then
 
     function check_version {
         local MIN_VERSION="4.3"
-        local version=`VBoxManage --version | perl -ne 'm/(\d\.\d)\./; print "$1"'`
+        local MY_VERSION=`VBoxManage --version | perl -ne 'm/(\d\.\d)\./; print "$1"'`
+	local VERSION_CHECK=`ruby -e "puts Gem::Version.new($MY_VERSION) > Gem::Version.new($MIN_VERSION)"`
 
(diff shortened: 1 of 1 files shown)


virtualbox_env.sh
```diff
@@ -14,9 +14,10 @@ if [[ -z "$VBM" ]]; then
 
     function check_version {
         local MIN_VERSION="4.3"
-        local version=`VBoxManage --version | perl -ne 'm/(\d\.\d)\./; print "$1"'`
+        local MY_VERSION=`VBoxManage --version | perl -ne 'm/(\d\.\d)\./; print "$1"'`
+	VERSION_CHECK=`ruby -e "puts Gem::Version.new($MY_VERSION) > Gem::Version.new($MIN_VERSION)"`
```
cbaenziger: Why not `local` here?

http-418: no reason at all

virtualbox_env.sh
```diff
@@ -14,9 +14,10 @@ if [[ -z "$VBM" ]]; then
 
     function check_version {
         local MIN_VERSION="4.3"
-        local version=`VBoxManage --version | perl -ne 'm/(\d\.\d)\./; print "$1"'`
+        local MY_VERSION=`VBoxManage --version | perl -ne 'm/(\d\.\d)\./; print "$1"'`
+	local VERSION_CHECK=`ruby -e "puts Gem::Version.new($MY_VERSION) > Gem::Version.new($MIN_VERSION)"`
```
cbaenziger: It looks like an indent inconsistency here?

cbaenziger:
Nice change!

PR #1211, opened 2018-05-15
Fixes for deployment issues in production

http-418:
I had to make these two fixes to deploy 3.5.0 to internet-disconnected hosts.

The first is a problem in the java cookbook.   It still attempts a download despite a local copy existing.  We copy it to the cache ahead of the cookbook to avoid the internet dependency.

The second is an issue in archive_bins.sh.  The escaping rules on `vagrant ssh -c` are complicated, and the wildcard is not expanded by the remote shell.  By removing the wildcard entirely, we can sweep up the split gemfiles.

### DIFF
--- a/cookbooks/bach_repository/recipes/oracle_java.rb
+++ b/cookbooks/bach_repository/recipes/oracle_java.rb
@@ -33,10 +33,19 @@ end
 java_tgz_name = Pathname.new(node['bach']['repository']['java']['jdk_url'])\
                         .basename.to_s
 
-remote_file "#{bins_dir}/#{java_tgz_name}" do
+jdk_local_path = "#{bins_dir}/#{java_tgz_name}"
+
+remote_file jdk_local_path do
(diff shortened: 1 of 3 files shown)


cookbooks/bach_repository/recipes/oracle_java.rb
```diff
@@ -33,10 +33,19 @@
...
   group 'root'
   mode 0444
   checksum node['bach']['repository']['java']['jdk_checksum']
 end
+
+ruby_block 'Copy JDK to file cache' do
+  block do
+    require 'fileutils'
+    ::FileUtils.cp(jdk_local_path, Chef::Config[:file_cache_path])
+  end
```
aespinosa: can we add `creates ` to be minimally idempotent?

http-418: Copying a file is already idempotent.

PR #1217, opened 2018-05-24
Automated serial console connections, Cluster prefix idempotence, General shell script linting and documentation

ronny-macmaster:
Most of this pull request is general linting and documentation that I've made to my fork to better understand the vm cluster build process. The pull request also includes some key functionality changes to automate the serial console connections and retain idempotence in prefixed-cluster builds.

The linting changes are not crucial to be merged, but I think they help to make the shell scripts a little more understandable. 

**After these changes, I was able to automatically build a vm cluster with a prefix in 3hrs 15mins.** 
Waiting for the hosts to install took 9 checks before proceeding.

1) Automates the serial console connections to the bcpc-vms in tests/automated_install.sh. 
These changes are mainly in lines 120-140 of tests/automated_install.sh. The previous screen session either approach failed to connect to the serial consoles or failed to kill them properly.
The main difference here is the use of the screen command argument, bash -c "$WAIT_FOR_HOSTS && $KILL_SCREENS".  This feature builds on the proposed changes of https://github.com/bloomberg/chef-bach/pull/868

2) Retains idempotence in prefixed-cluster builds
After changing the format of cluster.txt to include a node index as the first field, some of the previous installation scripts break idempotence. Lines 74-77 in vbox_create.sh and line 74 in vm_to_cluster.rb now check that the hostname starts with the prefix. Previous versions would always prepend a non empty prefix to the cluster.txt file.

3) General shell scirpt linting and documentation
I tried to refactor some of the shell script segments in a simpler way while retaining the original functionality. The pull request makes it look more complicated than it is. It may be a good idea to pull down the pull request branch and actually look at the scripts, mainly tests/automated_install.sh and vbox_create.sh. I realize that changes that make the scripts clearer and simpler to me may not necessarily do the same for others.

### DIFF
--- a/vbox_create.sh
+++ b/vbox_create.sh
@@ -69,8 +69,10 @@ require './lib/cluster_data.rb';
 include BACH::ClusterData;
 cp=ENV.fetch('BACH_CLUSTER_PREFIX', '');
 cp += '-' unless cp.empty?;
-vms = parse_cluster_txt(File.readlines('cluster.txt'))
-puts vms.map{|e| cp + e[:hostname]}.join(' ')
+vms = parse_cluster_txt(File.readlines('cluster.txt')).map do |e|
+  (e[:hostname] !~ /^#{cp}/) ? (cp + e[:hostname]) : e[:hostname]
+end.join(' ')
+puts vms
 "
 export VM_LIST=( $(/usr/bin/env ruby -e "$code_to_produce_vm_list") )
 
--- a/vbox_create.sh
+++ b/vbox_create.sh
@@ -113,29 +113,6 @@ function snapshotVMs {
   wait && printf "Done Snapshotting\n"
 }
 
-# ################################################################################
-# # Function to enumerate VirtualBox hostonly interfaces
-# # in use from VM's.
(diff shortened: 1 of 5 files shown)


vbox_create.sh
```diff
@@ -107,28 +113,28 @@ function snapshotVMs {
...
-  for vm in $($VBM list vms | sed -e 's/^[^{]*{//' -e 's/}$//'); do
-    ifs=$($VBM showvminfo --machinereadable $vm | \
-      egrep '^hostonlyadapter[0-9]*' | \
-      sed -e 's/^hostonlyadapter[0-9]*="//' -e 's/"$//')
-    for interface in $ifs; do
-      used_ifs[$interface]=$((${used_ifs[$interface]} + 1))
-    done
-  done
-}
+# ################################################################################
```
aespinosa: we can remove completely instead of commenting things out.

ronny-macmaster:
Going to close this PR in favor of resolving #1228 with a different approach. 
These changes should work for anyone who still wants to take the screened serial consoles approach in a private setting.

PR #1218, opened 2018-05-24
Override the Berkshelf solver through the environment

aespinosa:
So that we can start working on debugging/untangling circular dependencies

Reference: https://docs.chef.io/berkshelf.html#solver-keyword

### DIFF
--- a/Berksfile
+++ b/Berksfile
@@ -3,6 +3,11 @@ source 'https://supermarket.chef.io'
 
 metadata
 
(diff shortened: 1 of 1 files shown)


http-418:
I have never tried the alternate solver, but nothing could be worse than the default.

aespinosa:
The nice thing about the ruby solver is that it's sort of 'stricter' when it comes to circular dependencies.  it will simply refuse to solve and break.

PR #1226, opened 2018-06-05
Backup

ronny-macmaster:
Here's the first iteration of my `backup` cookbook.
This cookbook will provide schedule periodic HDFS directory backups on storage clusters.
The BACH-Backup role should be run on a storage cluster worker node.
The backup schedule can be modified with the [:backup][:hdfs][:schedules] attribute.

The cookbook has been tested on VM cluster builds, but it still needs to be tested on a physical lab cluster.
Please share any comments or concerns.


cookbooks/bcpc-hadoop/attributes/kerberos.rb
```diff
@@ -128,6 +128,15 @@
...
   },
+  backup: {
+    principal: node[:bcpc][:hadoop][:backup][:user],
+    keytab: 'backup.service.keytab',
+    owner: node[:bcpc][:hadoop][:backup][:user],
+    group: node[:bcpc][:hadoop][:backup][:user],
+    princhost: '_HOST',
+    perms: '0440',
+    spnego_keytab: 'spnego.service.keytab'
+  },
```
ronny-macmaster: This creates a new kerberos keytab.  This process is autmoated on VM clusters, but not in production.  That makes this PR a breaking change.

It should be tagged "breaking" and made a milestone for 4.0

tests/automated_install.sh
```diff
@@ -1,77 +1,84 @@
 #!/bin/bash
```
ronny-macmaster: duplicated changes should be taken out before merging.

PR #1229, opened 2018-06-07
Modify BCPC-Hadoop and the stub-environment to support the BACH-Backup role on BACH clusters.

ronny-macmaster:
These changes modify the `bcpc-hadoop`  cookbook to support the `backup` cookbooks. 
They work fine on a VM cluster, but the creation of a new kerberos keytab will cause breaking changes in certain production cluster environments.


ronny-macmaster:
These Breaking changes are required by #1226 to be useful anyway, so I will re-include them in that PR.

PR #1231, opened 2018-06-18
Fix Oozie high-availability hosts.

ronny-macmaster:
The current method for detecting hosts that run Oozie servers is decentralized, nonstandard, and inaccurate. Many of the auxiliary cookbooks that depend on Oozie (running oozie jobs) have their own method for setting the Oozie and ResourceManager URLs. These cookbooks (hdfsdu, ambari, backup, etc.) must call the `set_hosts` bcpc-hadoop helper before attempting to set the URLs. They also rely on the `set_hosts` function to properly set the `node.default[:bcpc][:hadoop][:oozie_hosts]` attribute. 

Currently, an oozie_host is detected as any node running the role[BCPC-Hadoop-Head-MapReduce]. Any node that runs the recipe[bcpc-hadoop::oozie] recipe w.o. running the MapReduce role goes undetected as an Oozie host. 

These changes centralize the initialization of two new attributes, `default[:bcpc][:hadoop][:oozie_url]` and `default[:bcpc][:hadoop][:rm_address]`. They also check for nodes running recipe[bcpc-hadoop::oozie] or role[BCPC-Hadoop-Head-MapReduce] to determine the correct oozie_hosts attribute. 

The auxiliary cookbooks will continue to work as before with these changes, but they can be refactored in the future to source these centralized attributes.

### DIFF
--- a/cookbooks/bcpc-hadoop/libraries/utils.rb
+++ b/cookbooks/bcpc-hadoop/libraries/utils.rb
@@ -186,35 +186,18 @@ end
 
   
 #
-# Function to retrieve commonly used node attributes so that the call to chef server is minimized
+# Function to retrieve commonly used node attributes.
+# Minimizes calls to the chef server.
 #
 def set_hosts
-  # Hadoop service to chef role mapping used by set_hosts to perform lookups
-  service_role = {
-    'zookeeper'=> 'role[BCPC-Hadoop-Head]',
-    'journal_node' => 'role[BCPC-Hadoop-Head]',
-    'resource_manager' => 'role[BCPC-Hadoop-Head-ResourceManager]',
-    'job_history_server' =>  'role[BCPC-Hadoop-Head-MapReduce]',
-    'oozie_server' => 'role[BCPC-Hadoop-Head-MapReduce]',
-    'hadoop_worker' => 'role[BCPC-Hadoop-Worker]',
-    'hadoop_datanode' => 'role[BCPC-Hadoop-Worker]', 
-    'hadoop_head' => 'role[BCPC-Hadoop-Head]',
-    'hive_server' => 'role[BCPC-Hadoop-Head-Hive',
-    'hbase_head' => 'role[BCPC-Hadoop-Head-HBase]'
-  }
-
-  service_recipe = {
-    'oozie_server' => 'recipe[bcpc-hadoop::oozie]'
-  }
-
   if node.run_state['cluster_def'].nil? then
     node.run_state['cluster_def'] = BACH::ClusterDef.new(node_obj: node)
   end
-  nodes = node.run_state['cluster_def'].fetch_cluster_def
+  hosts = node.run_state['cluster_def'].fetch_cluster_def
 
   # host search helper lambdas
-  runs_role = Proc.new { |host, role| host[:runlist].include? service_role[role] }
-  runs_recipe = Proc.new { |host, recipe| host[:runlist].include? service_recipe[recipe] }
+  runs_role = Proc.new { |host, role| host[:runlist].include? role }
+  runs_recipe = Proc.new { |host, recipe| host[:runlist].include? recipe }
 
   # mapped host objects
   to_host = Proc.new do |host| {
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -203,59 +203,94 @@ def set_hosts
     'hbase_head' => 'role[BCPC-Hadoop-Head-HBase]'
   }
 
+  srvc2recipe = {
```
aespinosa: can you rename to `service_recipe` ?

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -203,59 +203,94 @@ def set_hosts
...
   end
-  cd = node.run_state['cluster_def']
+  nodes = node.run_state['cluster_def'].fetch_cluster_def
+
+  # host search helper lambdas
+  runs_role = Proc.new { |host, role| host[:runlist].include? srvc2role[role] }
+  runs_recipe = Proc.new { |host, recipe| host[:runlist].include? srvc2recipe[recipe] }
+
+  # mapped host objects
+  to_host = Proc.new do |host| {
```
aespinosa: why the need for a similar hash with less keys?

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -203,59 +203,94 @@ def set_hosts
...
+        "#{node[:bcpc][:hadoop][:oozie_port]}/oozie"
+  end
+
+  # set the resourcemanager_url (rm_address)
+  rm_hosts = node[:bcpc][:hadoop][:rm_hosts]
+  if rm_hosts.length > 1
+    # high-availability
+    node.default[:bcpc][:hadoop][:rm_address] = node.chef_environment
+  else
+    node.default[:bcpc][:hadoop][:rm_address] =
```
aespinosa: we always use HA for YARN these days. i don't think handling a non-HA case is necessary.

cookbooks/bcpc-hadoop/libraries/utils.rb
ronny-macmaster: I wondered the same thing. I was trying to maintain the original functionality of the code, but utilizing a single hash doesn't hurt here.

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -203,59 +203,84 @@ def set_hosts
...
+  to_host = Proc.new do |host| {
+    'hostname' => host[:fqdn],
+    'node_number' => host[:node_id],
+    'zookeeper_myid' => nil
+  } end
   
   node.default[:bcpc][:hadoop][:zookeeper][:servers] = 
-    cd.fetch_cluster_def.select { |hst| hst[:runlist].include? srvc2role['zookeeper'] }.map {
-      |hst| { 'hostname' => hst[:fqdn], 'node_number' => hst[:node_id], 'zookeeper_myid' => nil } }
+    nodes.select { |n| runs_role.call(n, 'zookeeper') }.map { |n| to_host.call(n) }
```
cbaenziger: Can you iterate over a list of roles here rather than hardcoding them? (Ideally can you introspect the roles to see which ones include Oozie?)

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -203,59 +203,84 @@ def set_hosts
...
+    nodes.select { |n| runs_role.call(n, 'hbase_head') }.map { |n| to_host.call(n) }
 
   # different flavors of hive
   node.default[:bcpc][:hadoop][:hive_hosts] = 
-      cd.fetch_cluster_def.select { |hst| hst[:runlist].include? srvc2role['hive_server'] }.map {
-      |hst| { 'hostname' => hst[:fqdn]} }
+    nodes.select { |n| runs_role.call(n, 'hive_server') }.map { |n| to_host.call(n) }
 
-  # BCPC-Hadoop-Head-MapReduce
+  # (the BCPC-Hadoop-Head-MapReduce role nests the oozie recipe)
```
cbaenziger: I am not groking what the relevance of the MapReduce role is to anything here?

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -203,59 +203,84 @@ def set_hosts
...
+  elsif oozie_hosts.length == 1
+    # single oozie host
+    node.default[:bcpc][:hadoop][:oozie_url] =
+      "http://#{float_host(oozie_hosts.first[:hostname])}:" +
+        "#{node[:bcpc][:hadoop][:oozie_port]}/oozie"
+  end
+
+  # set the HA resourcemanager_url (rm_address)
+  rm_hosts = node[:bcpc][:hadoop][:rm_hosts]
+  node.default[:bcpc][:hadoop][:rm_address] = node.chef_environment
```
cbaenziger: Should not this check that we have more than one RM?

ronny-macmaster: That's how I originally had it, but this should work if we want to throw errors when there's no HA RM. 
I can change it back to work with a single RM too. @aespinosa thoughts?

cookbooks/bcpc-hadoop/libraries/utils.rb
ronny-macmaster: The MapReduce role enforces that every node running a JobHistoryServer also runs an Oozie server. 
https://github.com/bloomberg/chef-bach/blob/master/stub-environment/roles/BCPC-Hadoop-Head-MapReduce.json

We might be better off changing this.

cookbooks/bcpc-hadoop/libraries/utils.rb
ronny-macmaster: Yea, I can probably arrange these individual statements into a clever loop. I'll make that change.

cookbooks/bcpc-hadoop/libraries/utils.rb
```diff
@@ -186,76 +186,66 @@ def znode_exists?(znode_path, zk_host = 'localhost:2181')
...
+      "http://#{first_host}/#{oozie_port}/oozie"
+    end
+
+  # set the resourcemanager_url (rm_address)
+  rm_hosts = node['bcpc']['hadoop']['rm_hosts']
+  first_host = float_host(rm_hosts.first['hostname'])
+  rm_port = node['bcpc']['hadoop']['yarn']['resourcemanager']['port']
+
+  node.default['bcpc']['hadoop']['rm_address'] =
+    if rm_hosts.length > 1
```
aespinosa: Add comment:

```
# FIXME if we just use plain HA ResourceManagers in yarn_site.rb we can simply remove this if statement
```

https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/yarn_site.rb#L37

ronny-macmaster:
We need to eventually remove the checks for >= 2 RM and Oozie hosts running scattered throughout BCPC-Hadoop and various service cookbooks. Ideally, all of these urls should be behind a VIP.

cookbooks/bcpc-hadoop/libraries/utils.rb
cbaenziger: No we should not always assume the RM's are HA; test-kitchen style tests will not be.

cbaenziger:
Ron says all tests pass and this has been through many codereviews. Let's clean this out.

cookbooks/bcpc-hadoop/libraries/utils.rb
aespinosa: I don't understand why not.  The failover proxy provider clients can support having only one RM defined behind its configuration.  test-kitchen should still work.

cbaenziger: Ah I had not realized one could run a single RM without triggering an error that multiple were needed. It does still pull in Zookeeper though.

cbaenziger:
Just an aside but re-reviewing this for #1288, we can simply remove `set_hosts` now; if we wish. Originally `set_hosts` was a caching function due to calls to the Chef server being slow in early Chef-11.

Now that we have removed the Chef-Server in the look-up, it should be possible to write a utility function for the lookup of hosts running a role. (And perhaps another that would ease recipe to role lookups.) Then we can replace the node attributes with those lookups; if that pattern would be desired.

PR #1236, opened 2018-07-03
Remove the hdp-utils-gpl bcpc_repo

ronny-macmaster:
Building off #1235 , the bcpc_repo resource should also be removed.


aespinosa:
Thanks! didn't realized I didn't commit that

PR #1238, opened 2018-07-11
Tune jute max buffer

amithkanand:
With `cluster replication` and `region replication` enabled we are observing a large number of child `zondes` under `/hbase/replication zonde`. The value of `5million` seems to be small and causes `hbase` to fail to connect with `Zookeeper` while trying to read replicaiton related inforamtion. Increating the value to `20 million` seems to have worked well in couple of tested environments.


aespinosa:
Cherry-picked in 3.3

PR #1239, opened 2018-07-16
repair BCPC-Hadoop roles

ronny-macmaster:
Previously, we merged the PR for bach backups, #1226.
It turns out that the modifications to the BCPC roles cause the builds to crash and burn on clusters that utilize LDAP for user and group information. 

We should revert these role modifications immediately.
I need to devise a way to completely contain the needed backup functionality in the BACH-Backup role that works on both VM and Physical clusters.


vt0r:
Closing per Ron's comment about no longer needing this PR.

PR #1241, opened 2018-07-16
Forked Smoke-Tests 

NinjaLabib:
Currently tests are serialized so if one fails, all following tests won't even be run. We want to make sure every test is able to be run to test every service. The best way to do this is using fork + join + decision. If one of these fails, it won't goto a fail path and end oozie execution because all fail/success paths goto the join. Once all tests are done, a decision is used to check if lastErrorNode() was populated or not which would indicate whether or not something failed. 

Also removed a duplicate line from job properties.


bijugs:
LGTM

ronny-macmaster:
LGTM

Issue #1228, opened 2018-06-07
wait-for-hosts loops forever until the user manually connects to serial consoles


ronny-macmaster:
Sal and Ron have found that connecting to each serial console in turn will fix the shell script checks for living VMs.

We should probably drop serial console support entirely.  We can't troubleshoot boot problems on VMs or physical hardware via serial console, so the first step in every troubleshooting process is to open a video console.

vt0r:
Fixed in #1242

PR #1242, opened 2018-07-23
Kerberos support for Kafka

http-418:
This is my second attempt to Kerberize BACH Kafka clusters.

Zookeeper is kerberized using the canned support built for `bcpc-hadoop`.  Kafka is kerberized by an independent JAAS template, and has an authenticated session with Zookeeper.

TODO:
* Find out whether Kafka will still permit an insecure link to Zookeeper even with a "Client" section in the JAAS.


cookbooks/bcpc_kafka/attributes/default.rb
```diff
@@ -52,9 +49,9 @@
   broker[:dual][:commit][:enabled] = false
   broker[:offsets][:storage] = 'kafka'
 
-  # Default to a 0.10.0 protocol for 0.11.x upgrades.
-  broker[:inter][:broker][:protocol][:version] = '0.10.0'
-  broker[:log][:message][:format][:version] = '0.10.0'
+  # Default to a 1.1.0 protocol.
+  broker[:inter][:broker][:protocol][:version] = '1.1.0'
+  broker[:log][:message][:format][:version] = '1.1.0'
```
cbaenziger: Should the 1.1.0 be another attribute like our active release notion in the Hadoop code?

cookbooks/bcpc_kafka/attributes/default.rb
```diff
@@ -66,16 +63,46 @@
...
+  # We default to using PLAINTEXT for inter-broker communication, then
+  # migrate to SASL after the cluster is up.
+  #
+  broker[:security][:inter][:broker][:protocol] = 'PLAINTEXT'
+
+  #
+  # Deprecated values used to generate listeners in the past.  Still
+  # required by the upstream cookbook, but largely ignored by Kafka itself.
+  #
+  broker[:advertised_host_name] = node[:fqdn] # Deprecated.
```
cbaenziger: I do not see `advertised` in the Kafka cookbook.

cookbooks/bcpc_kafka/recipes/kafka.rb
```diff
@@ -107,9 +107,69 @@
...
+  action :create
+  username 'kafka'
+  home '/var/empty/kafka'
+  shell '/sbin/nologin'
+  #
+  # Chef does not have a :create_if_missing for the user resource, so
+  # we avoid trying to alter an existing user by refusing to converge
+  # if a kafka user exists.
+  #
+  not_if 'id kafka'
```
cbaenziger: Though `id` is more concise we have used `getent` elsewhere (e.g. [here](https://github.com/bloomberg/chef-bach/blob/3be0bf618c2cb195c10ff403f858bdb58b13249a/cookbooks/bcpc/recipes/nscd.rb#L40)).

cookbooks/bcpc_kafka/recipes/kafka.rb
```diff
@@ -107,9 +107,69 @@
...
+configure_kerberos 'kafka_kerb' do
+  service_name 'kafka'
+end
+
+#
+# Here is the Kafka-specific Kerberos material: we create a JAAS
+# configuration for Kafka, and add it to the Kafka command line.
+#
+# The configuration is in /etc/bach because the final configuration
+# directory created by the upstream Kafka cookbook isn't available
```
cbaenziger: Can't we simply do a directory resource to keep everything colocated, and since the path should be available as an attribute?

vt0r:
Fixes #1228

PR #1254, opened 2018-08-08
fix ldapsearch for hdfs dir creation

ronny-macmaster:
These changes should allow hdfs user and group directory provisioning to work in bcpc-hadoop with ldapsearch commands. Some hdfs attributes have been refactored to logically group them in an ldap.rb attributes file. I have tested these changes on a hardware cluster.


cbaenziger:
Feel free to call this a bike shed, but `dirs.rb`->`create_dirs.rb`?

cookbooks/bcpc-hadoop/files/default/dirs.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      base = s.sub(/[kmgtp]/, '').to_i
+      prefix = prefixes[s.sub(/[0-9.]*/,'').downcase]
+
+      return base * prefix
+    end
+
+    # set space quota
+    if quotas[dir]
+      max_quota = HdfsConstants::QUOTA_DONT_SET
```
cbaenziger: Any reason to not use `HdfsConstants.QUOTA_RESET`?

PR #1255, opened 2018-08-08
utilize the quota_reset hdfs constant

ronny-macmaster:


cbaenziger:
Yeah, I was wondering what was used by `hdfs dfsadmin -clrQuota`/`hdfs dfsadmin -clrSpaceQuota`. @ronny-macmaster all tested out okay?

PR #1257, opened 2018-08-09
fix role account filter

ronny-macmaster:
We should expand the definition of what constitutes a "role account".


cbaenziger:
Looks like what I would expect. For future reference, do you have a role account for each use case?

cbaenziger:
Looks great to me too

PR #1271, opened 2018-09-21
Refactor the bcpc::ssh recipe. Remove 'failed to start' error.

ronny-macmaster:
Here's a simplification the SSH recipe.
We can fetch key items from the vault at compile rather than converge time.
If done this way, we can take advantage of the recipe DSL instead of creating resources with Chef::Resource.

Also, on our current version of chef, the init provider seems to work fine with upstart.
If we utilize the service resource :start action, the 'failed to start ssh' error on every chef run goes away.

The PR is broken up into two commits. 
The first commit can be cherry picked to only fix the 'failed to start' error.
The second commit is recommended to simplify the cognitive complexity of the recipe.
It shortens the recipe by 25%, and it removes some unnecessary levels of nesting.
The diff looks a little nasty, but viewing the full recipe looks a lot nicer.
This is the typically the first recipe that cluster_assign_roles.rb runs on a newly provisioned node,
so we should start the chef-run off right.

This has been tested under the following situations:
* ssh is not running
* ssh is already running

* host keys already exist in the vault
  should save the key values in the vault over the existing host keys on the node.
* host keys don't exist in the vault, but they exist on the node.
  create and update the vault with the existing host key values from the node.
* host keys exist in neither the vault nor on the node.
  now, it logs warning messages. previously, this was unhandled in the original recipe.

If the vault item does not yet exist, and the client does not have admin privileges to create the vault item, an exception is immediately raised at compile time rather than proceeding with a improperly configured ssh service.

### DIFF
--- a/cookbooks/bcpc/recipes/ssh.rb
+++ b/cookbooks/bcpc/recipes/ssh.rb
@@ -36,21 +36,7 @@ package 'openssh-server' do
 end
 
 service 'ssh' do
-  action :enable
-  ignore_failure true
-end
-
-#
-# The Init::Debian provider fails to start an already-started
-# service on Ubuntu 14.04, due to bugs in their sysvinit
-# compatibility scripts.
-#
-execute '/etc/init.d/ssh start' do
-  action :run
-
-  if node[:lsb][:codename] == 'trusty'
-    ignore_failure true
-  end
+  action [:enable, :start]
 end
 
 template '/etc/ssh/sshd_config' do
--- a/cookbooks/bcpc/recipes/ssh.rb
+++ b/cookbooks/bcpc/recipes/ssh.rb
@@ -42,7 +42,6 @@ end
 template '/etc/ssh/sshd_config' do
   source 'sshd_config.erb'
   mode 00644
-  notifies :restart, 'service[ssh]', :immediately
   variables lazy { {
     address_family: node['bcpc']['ssh']['address_family'],
     listen_address: node[:bcpc][:management][:ip],
@@ -50,6 +49,7 @@ template '/etc/ssh/sshd_config' do
   } }
 
   # Don't rewrite the file unless we know the listen address is valid!
+  notifies :reload, 'service[ssh]'
   only_if {
     bound_addresses = node[:network][:interfaces]
       .map { |_, ii| ii[:addresses] }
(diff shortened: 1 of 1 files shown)


cookbooks/bcpc/recipes/ssh.rb
```diff
@@ -74,118 +60,82 @@
...
+        content vault_item[key_type]
+        sensitive true
+      end
+
+      # Generate a new public key when private key is updated
+      execute "generate-#{key_type}-public-key" do
+        umask 0222
+        command "ssh-keygen -y -f #{private_key_path} > #{public_key_path}"
+        subscribes :run, "file[#{private_key_path}]", :immediately
+        notifies :reload, 'service[ssh]'
```
cbaenziger: Is this an immediate or delayed reload you want?

cookbooks/bcpc/recipes/ssh.rb
```diff
@@ -74,118 +60,82 @@
...
+      Chef::Log.warn("Can't find #{key_type} key in vault!")
+      Chef::Log.warn("Creating #{key_type} key in vault.")
+      vault_item[key_type] = existing_key
+      vault_item.save
     else
-      Chef::Log.warn("No vault item found for ssh_host_keys/#{node[:fqdn]}!")
+      # Neither vault item nor node key exists
+      Chef::Log.warn("Can't find ssh_host_keys/#{node[:fqdn]} in vault!")
+      Chef::Log.warn("#{private_key_path} does not exist either!")
+      Chef::Log.warn("Unsupported Key Type: #{key_type}")
```
cbaenziger: I do not understand what unsupported key type means here?

cookbooks/bcpc/recipes/ssh.rb
```diff
@@ -74,118 +60,82 @@
...
-        Chef::Log.warn('Failed to create new vault ' \
-                       "ssh_host_keys/#{node[:fqdn]}. " \
-                       "Is this node an admin?\n#{eee}")
-      end
-    end
-  end
-  not_if { node[:fqdn] == get_bootstrap }
+require 'chef-vault'
+
+def load_ssh_keys
```
cbaenziger: Please add a method comment on what you intend this function to do, return and consume.

cookbooks/bcpc/recipes/ssh.rb
```diff
@@ -74,118 +60,82 @@
...
-          # to notify the SSH service.
-          #
-          self.notifies :restart, 'service[ssh]', :immediately
-          self.resolve_notification_references
-        end
+if node[:fqdn] != get_bootstrap
+  # First, try to retrieve existing host keys from the vault.
+  # If we fail to load from the vault, log the failure and create an item.
+  # If we also fail to create a vault item, this node may not be an admin.
+  vault_item = load_ssh_keys || create_ssh_keys
```
cbaenziger: This logic seems unobvious; what return do you expect from a successful `create_ssh_keys` call?

cookbooks/bcpc/recipes/ssh.rb
```diff
@@ -74,118 +60,82 @@
...
-          vault_item[key_type] = ::File.read(private_key_path)
-          vault_item.save
-        end
-      end
-    else
-      Chef::Log.warn("No vault item found for ssh_host_keys/#{node[:fqdn]}!")
-    end
-  end
-  not_if { node[:fqdn] == get_bootstrap }
+def create_ssh_keys
```
cbaenziger: Please add a method comment on what you intend this function to do, return and consume.

cookbooks/bcpc/recipes/ssh.rb
```diff
@@ -74,118 +60,82 @@
...
+  vault_item = ChefVault::Item.new('ssh_host_keys', node[:fqdn])
+  vault_item.admins([get_bootstrap, node[:fqdn], 'admin'].join(','))
+  vault_item.search("fqdn:#{node[:fqdn]}")
+  vault_item.save
+rescue Exception => eee
```
cbaenziger: I do not think `eee` is common in our code for exceptions. Don't we use `e` or `ex` normally?

http-418:
Using chef-vault at compile time is unsafe and will break the debugger if anything goes wrong.

We should move things _out_ of the compile phase rather than into it.

ronny-macmaster:
putting this on the backburner for now...

PR #1278, opened 2018-09-26
Ensure unprivileged users have unique tmp and home dirs

cbaenziger:
This PR introduces [pam_namespaces](https://linux.die.net/man/8/pam_namespace) to ensure that each user gets its own kernel usernamespace for its potential state directories that are managed outside YARN. One further addition is automatic creation and clean-up of directories as needed using [pam_exec](https://linux.die.net/man/8/pam_exec).

Providing per-user namespaces means we can prevent accidental data sharing and we will prevent many imporoper uses of temporary directories by pro-actively cleaning them up. We have seen many users impact themselves with both anti-patterns and similarly when a user disconnects we will ensure they stop taking up space on our root partition for what should be their ephemeral home directory. *We will need to notify users of this change!*

The directories affected are listed in [`/etc/secuirty/namespaces.conf](https://github.com/cbaenziger/chef-bach/blob/tmp_dirs/cookbooks/bcpc/templates/default/pam_namespace.conf.erb#L8-L11). Users listed in that file are excluded from that line.

The directories created in [`inst_dir.sh.erb`](https://github.com/cbaenziger/chef-bach/blob/tmp_dirs/cookbooks/bcpc/templates/default/inst_dir.sh.erb) are the necessary directories for pam_namespace to create directories and when the user is no longer running any processes and the user terminations its session the script is called to remove the directories. (If one uses `nohup` or such you can hold your directory around I think but I am not too concerned about that right now; we can write clean-up logic if need be later; similarly no quotas are introduced here but use of ZFS or possibly LVM could allow us to easily do that.)

The structure under the polyinstantion directory looks like follows; and inside each directory is what the user see:
```
ubuntu@iptables-bcpc-vm2:~$ sudo ls -l /inst-dirs
total 4
drwxrwxrwt 14 root root 4096 Sep 26 00:50 tmp_myuser
drwxrwxrwt  3 root root 4096 Sep 26 00:50 var_tmp_myuser
ubuntu@iptables-bcpc-vm2:~$ sudo ls -l /dev/shm/inst-dirs
total 0
drwxrwxrwt 2 root root 40 Sep 26 00:50 inst_myuser
```


cbaenziger:
Oh, also to explain; we override `/etc/security/namespace.init` to correct [Ubuntu Bug 1081323|https://bugs.launchpad.net/ubuntu/+source/pam/+bug/1081323].

cookbooks/bcpc/templates/default/inst_dir.sh.erb
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# This script is managed by Chef
+# It is designed to setup directories using pam_exec
+# for the pam_namespaces module to isolate user-direcotires
+
+if [ -z "$PAM_TYPE" -o "$PAM_TYPE" == "open_session" ]; then
+  # need to setup the polyinstantation directory for login
+  mkdir -pm 000 "/dev/shm/<%= @shm_polyinstantion_dir %>" \
```
cbaenziger: Sal asks for an extra 0

cookbooks/bcpc/templates/default/inst_dir.sh.erb
```diff
@@ -0,0 +1,20 @@
...
+                "<%= @polyinstantion_dir %>"
+elif [ -z "$PAM_TYPE" -o "$PAM_TYPE" == "close_session" ]; then
+  # see if user still has any prescence on this machine
+  pgrep -u "$PAM_USER" && exit 0
+  # if no prescense of user, remove all trace of their activity
+  rm -rf "<%= @polyinstantion_dir %>/home_$PAM_USER" \
```
cbaenziger: Chris asks please ensure that hardlink traversal isn't a risk here.

cookbooks/bcpc/recipes/pam.rb
```diff
@@ -0,0 +1,86 @@
...
+  mode 500
+  variables(shm_polyinstantion_dir: shm_polyinstantion_dir,
+            polyinstantion_dir: polyinstantion_dir)
+end
+
+template '/etc/security/namespace.conf' do
+  source 'pam_namespace.conf.erb'
+  mode 500
+  variables(lazy {{ real_home_dir_users:
+                     node['bcpc']['pam_namespace']['real_home_dir_users'].join(','),
```
cbaenziger: Ron asks, please `.uniq` on `node['bcpc']['pam_namespace']['real_home_dir_users']`.

cookbooks/bcpc/templates/default/pam_namespace.init.erb
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
cbaenziger: Please pull in comment https://github.com/bloomberg/chef-bach/pull/1278/#issuecomment-427468873 to the script so it persists with the file -- and we remove this file on upgrade of Ubuntu.

stub-environment/roles/BCPC-Hadoop-Worker.json
```diff
@@ -13,8 +13,9 @@
     "recipe[bach_krb5::krb5_client]",
     "recipe[hdfsdu::create_user]",
     "recipe[bcpc-hadoop::bach_backup_wrapper]",
-    "recipe[bcpc-hadoop::configs]",
+    "recipe[bcpc::pam]",
```
cbaenziger: Ron asks why did configs move?

Since there's no longer any dependency on pam in configs, we can move configs to be closer to the Hadoop components which use it.

cookbooks/bcpc/recipes/cronjobs.rb
```diff
@@ -18,9 +18,12 @@
...
 
+polyinstantion_dir = node['bcpc']['pam_namespace']['polyinstantion_dir']
+shm_polyinstantion_dir = node['bcpc']['pam_namespace']['shm_polyinstantion_dir']
+
 clear_tmp = node['bcpc']['cronjobs']['clear_tmp']
-execute 'clear /tmp' do
-  command '/usr/bin/find /tmp -type f '\
+execute 'clear tmp dirs' do
+  command '/usr/bin/find /tmp #{polyinstantion_dir} /dev/shm/#{shm_polyinstantion_dir} -type f '\
           "-atime +#{clear_tmp['atime_age']} -delete && "\
```
vt0r: `/tmp` is usually `tmpfs`, but we should verify this is not mounted with `noatime`, as this has likely been broken forever if so.

cookbooks/bcpc/templates/default/inst_dir.sh.erb
vt0r:
```suggestion
  mkdir -pm 0000 "/dev/shm/<%= @shm_polyinstantion_dir %>" \
```

cookbooks/bcpc/recipes/cronjobs.rb
cbaenziger: On a VM cluster I see that root has atime enabled and /tmp is on root:
```ubuntu@tmpdirs-bcpc-vm4:~$ cat /etc/mtab 
/dev/mapper/tmpdirs--bcpc--vm4--vg-root / ext4 rw,errors=remount-ro 0 0
proc /proc proc rw,noexec,nosuid,nodev 0 0
sysfs /sys sysfs rw,noexec,nosuid,nodev 0 0
none /sys/fs/cgroup tmpfs rw 0 0
none /sys/fs/fuse/connections fusectl rw 0 0
none /sys/kernel/debug debugfs rw 0 0
none /sys/kernel/security securityfs rw 0 0
udev /dev devtmpfs rw,mode=0755 0 0
devpts /dev/pts devpts rw,noexec,nosuid,gid=5,mode=0620 0 0
tmpfs /run tmpfs rw,noexec,nosuid,size=10%,mode=0755 0 0
none /run/lock tmpfs rw,noexec,nosuid,nodev,size=5242880 0 0
none /run/shm tmpfs rw,nosuid,nodev 0 0
none /run/user tmpfs rw,noexec,nosuid,nodev,size=104857600,mode=0755 0 0
none /sys/fs/pstore pstore rw 0 0
tracefs /var/lib/ureadahead/debugfs/tracing tracefs rw,relatime 0 0
/dev/sda1 /boot ext2 rw 0 0
systemd /sys/fs/cgroup/systemd cgroup rw,noexec,nosuid,nodev,none,name=systemd 0 0
/dev/sdb1 /disk/0 xfs rw,noatime,nodiratime,inode64 0 0
/dev/sdc1 /disk/1 xfs rw,noatime,nodiratime,inode64 0 0
/dev/sdd1 /disk/2 xfs rw,noatime,nodiratime,inode64 0 0
/dev/sde1 /disk/3 xfs rw,noatime,nodiratime,inode64 0 0```

File: cookbooks/bcpc/recipes/pam.rb (hunk 3 above)
Comment (vt0r) [reply cookbooks/bcpc/recipes/pam.rb]: Possibly also sort.

Comment (vt0r) [inline cookbooks/bcpc/recipes/pam.rb]: Not sure if this would work, since I haven't tested it:

```suggestion
                     node['bcpc']['pam_namespace']['real_home_dir_users'].uniq.sort.join(','),
```

vt0r: Seems ok

``` ruby
[1] pry(main)> blah = %w(hdfs hdfs yarn hbase mapred hdfs hbase yarn mapred mapred hdfs hbase hadoop)
=> ["hdfs", "hdfs", "yarn", "hbase", "mapred", "hdfs", "hbase", "yarn", "mapred", "mapred", "hdfs", "hbase", "hadoop"]
[2] pry(main)> blah.uniq.sort.join(',')
=> "hadoop,hbase,hdfs,mapred,yarn"
```

cbaenziger:
All requested changes made

cookbooks/bcpc/recipes/pam.rb
```diff
@@ -0,0 +1,90 @@
...
+
+# Chef recipe to implement pam_namespace polyinstantiated directories
+# This will provide users the appearance they are the only user with data
+# in the affected directories -- also this will clean-up their data when
+# leaving the machine
+
+directory '/inst-dirs' do
+  user 'root'
+  group 'root'
+  mode 0o000
```
vt0r:
```suggestion
  mode 0o0000
```

cookbooks/bcpc/templates/default/inst_dir.sh.erb
vt0r: Looks like this got merged out:
```suggestion
  mkdir -pm 0000 "/dev/shm/<%= @shm_polyinstantion_dir %>" \
```

PR #1285, opened 2018-10-03
Allow fair_share_queue overrides

ronny-macmaster:
These changes will allow an attribute to override / define fair scheduler queue properties. 

Tested on both the VM and dev hardware cluster.

**Purpose:**
Deploying a release on existing hardware clusters quickly can sometimes prove challenging.
We'd like to avoid deploying new releases to push out small and urgent compute budget changes.
These changes will allow us to override the budgets codified in tenant recipes with environment attributes.

**Usage:**
Override the node['bcpc']['hadoop']['yarn']['fair_scheduler_queue'] in the environment json.
The data structure takes the following form:

```
node['bcpc']['hadoop']['yarn']['fair_scheduler_queue'] = {
  queue_name[string]: {
    parent_resource: [chef_resource_string],
    minResources: [string ~= /(\d+)mb (\d+)vcores/],
    maxResources: [string ~= /(\d+)mb (\d+)vcores/],
    maxRunningApps: [integer],
    maxAMShare: [float],
    weight: [float],
    schedulingPolicy: [string ~= /(fair|fifo|drf)/i],
    aclSubmitApps: [string (csv)],
    aclAdministerApps: [string (csv)],
    minSharePreemptionTimeout: [integer],
    fairSharePreemptionTimeout: [integer],
    fairSharePreemptionThreshold: [float]
  },
  ...
}
```

See the second comment below for an example.

**Reference:**
1. [fair_share_queue resource](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/libraries/fair_share_queue.rb)
1. [fair_share_queue helper methods](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/libraries/hadoop_helpers.rb)
1. [yarn_schedulers recipe](https://github.com/bloomberg/chef-bach/blob/master/cookbooks/bcpc-hadoop/recipes/yarn_schedulers.rb)
1. [fair_scheduler reference](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/FairScheduler.html)

### DIFF
--- a/cookbooks/bcpc-hadoop/recipes/yarn_schedulers.rb
+++ b/cookbooks/bcpc-hadoop/recipes/yarn_schedulers.rb
@@ -12,10 +12,20 @@ fair_share_queue "default" do
 end
 
 file "/etc/hadoop/conf/fair-scheduler.xml" do
-  content lazy {fair_scheduler_xml(node.run_state[:fair_scheduler_queue] || [],
-    node[:bcpc][:hadoop][:yarn][:fairSchedulerOpts],
-    node[:bcpc][:hadoop][:yarn][:queuePlacementPolicy])}
   mode 0644
+  content lazy {
+    # Merge the fair_scheduler_queue environment overrides
+    # into the definitions from fair_scheduler_queue resources.
+    # Definitions in the node attribute take precedence over the resources.
+    queue_defs = node.run_state[:fair_scheduler_queue] || {}
+    queue_overrides = node[:bcpc][:hadoop][:yarn][:fair_scheduler_queue] || {}
+
+    fair_scheduler_xml(
+      queue_defs.merge(queue_overrides),
+      node[:bcpc][:hadoop][:yarn][:fairSchedulerOpts],
+      node[:bcpc][:hadoop][:yarn][:queuePlacementPolicy]
+    )
+  }
 end
 
 template "/etc/hadoop/conf/capacity-scheduler.xml" do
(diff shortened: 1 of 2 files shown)


ronny-macmaster:
hadoopy

ronny-macmaster:
```
Test-Laptop.json:
....
        "yarn": {
          "fair_scheduler_queue": {
            "user": { },
            "groups": { },
            "rmacmaster": {
              "minResources": "1300mb, 5vcores",
              "parent_resource": "fair_share_queue[groups]"
            },
            "chef-bach": {
              "minResources": "2600mb, 10vcores",
              "parent_resource": "fair_share_queue[groups]"
            }
          },
...
```

aespinosa:
This is the type of change that I would like to have chefspec laid out.

aespinosa:
Can you expound on the need to have it available in the environment? Is this a hack so that we won't need to cut releases?

cookbooks/bcpc-hadoop/recipes/yarn_schedulers.rb
```diff
@@ -12,10 +12,20 @@
 end
```
aespinosa: Add a comment on the PR description on how the environment should be set.

vt0r:
> 
> 
> Can you expound on the need to have it available in the environment? Is this a hack so that we won't need to cut releases?

As far as I understand, yes. This is one of the last remaining items we write tenant recipes for in the wrapper (gross), and we'd like to stop doing that if possible.

ronny-macmaster:
> 
> 
> > Can you expound on the need to have it available in the environment? Is this a hack so that we won't need to cut releases?
> 
> As far as I understand, yes. This is one of the last remaining items we write tenant recipes for in the wrapper (gross), and we'd like to stop doing that if possible.

Yes, this is a hack to avoid cutting releases when changing compute queues. I added some documentation in the PR description to elaborate on this.

Managing the queues through tenant recipes can remain the preferred approach because these recipes provide a single source of truth across environments. However, environment overrides will give us the flexibility to quickly turnaround compute queue changes while tenant recipe changes pend in the next release.

Issue #1288, opened 2018-10-07
Master fails to populate node[:bcpc][:hadoop][:hs_hosts] 


cbaenziger:
Trying to build master and debug why the Hive smoke tests fail I see that master seems to fail in populating `node[:bcpc][:hadoop][:hs_hosts] ` this is used by:
* [yarn_site.rb](https://github.com/bloomberg/chef-bach/blob/3f768e717853507bc71b7b724b3fe6a8f726cdb1/cookbooks/bcpc-hadoop/recipes/yarn_site.rb#L248)
* [mapred_site.rb](https://github.com/bloomberg/chef-bach/blob/3f768e717853507bc71b7b724b3fe6a8f726cdb1/cookbooks/bcpc-hadoop/recipes/mapred_site.rb#L8)

The `yarn-site.xml` value `yarn.log.server.url` and `mapred-site.xml` values `mapreduce.jobhistory.address` and `mapreduce.jobhistory.webapp.address` come from this node attribute.

cbaenziger:
The last place I see we were setting `node[:bcpc][:hadoop][:hs_hosts] was in the `set_hosts` function rewritten in [#1231](https://github.com/bloomberg/chef-bach/pull/1231/files#diff-4301562bb3786f09a63001f8d6489b37L223).

cbaenziger:
#1290 should have fixed this now.

PR #1290, opened 2018-10-10
Fix for #1288 - Master fails to populate node[:bcpc][:hadoop][:hs_hosts]

cbaenziger:
This removes the unused `jn_hosts` entry and corrects it to `hs_hosts`; I have verified that it runs through on a VM cluster happily and Oozie smoke-tests and HDFSDU now work again after this change.


ronny-macmaster:
looks great to me!

PR #1296, opened 2018-10-26
Update the iPXE efi image from http://boot.ipxe.org

aespinosa:
Upstream build 2018 Oct 26


vt0r:
I watched this boot said actual hardware, so I approve.

PR #1300, opened 2018-11-02
Pin seven_zip cookbook for Chef-Client 12

cbaenziger:
The [seven_zip](https://supermarket.chef.io/cookbooks/seven_zip/versions/3.1.0) cookbook has moved on to requiring Chef-Client 13 so pin to version ~3.0.0 which is the last to support Chef-Client 12.


aespinosa:
Did the feedback myself in https://github.com/bloomberg/chef-bach/pull/1304/commits/29b8680a384016926af39ad95e28fb020e474ed6

PR #1303, opened 2018-11-02
Fix the zookeeper environment jvm flags.

ronny-macmaster:
The previous method of assignment does not work in bash.
The alternative to the new approach is to use backslashes within one multi-line string.


ronny-macmaster:
This is not a crucial bugfix, and merging #1302 would render this irrelevant.

cbaenziger:
For reference, this is still a relevant fix in master @ronny-macmaster.

cbaenziger:
This looks good to me too

vt0r:
Is this PR still necessary after #1302 ?

ronny-macmaster:
Yes, apparently #1302 doesn't revert any changes to [zk-zookeeper-env.sh.erb](cookbooks/bcpc-hadoop/templates/default/zk_zookeeper-env.sh.erb). We still need this change for this env script to actually work.

ronny-macmaster:
This will break zookeeper if libjvmkill.so doesn't exist on the cluster.

PR #1304, opened 2018-11-06
move to vagrantfile

aespinosa:
highlights:

* use a proper `Vagrantfile` instead of `VBoxManage` shell scripts
* convert most ssh invocations to modern vagrant-isms
* download boxes via `vagrant box add`
* install chefdk through the omnibus installer
* delete `vm-to-cluster.sh`

TODO:

- [x] parse cluster.txt to create nodes in the Vagrantfile


aespinosa:
This got me to boot a bootstrap and run `c-a-r Basic`.  some networking bits needs to be fixed though. which is probably in Andrew's branch.

aespinosa:
Got `car basic` to work.  But `bcpc::networking` blows out the NAT device in the vagrant box losing its upstream connection.

Will probably need to setup ip forwarding in the bootstrap properly and then use the bootstrap as the VM's gateway once the NAT devices disappear.

aespinosa:
After configuring iptables masquerade and port-forwarding in the bootstrap. Basic now works.

Was able to build a hadoop cluster (sans hdfsdu, zabbix) with this.

cookbooks/bcpc/recipes/ufw.rb
```diff
@@ -1,78 +0,0 @@
-#
```
aespinosa: This is needed by the VM cluster so that the bootstrap can act as the fake router for VMs.

PR #1308, opened 2018-11-21
Disable `pcp` and chef-client for Kafka clusters

http-418:
These two changes are lessons learned rolling out the v3.6.3 release.

* Performance Co-pilot (`pcp`) is a set of daemons that come with Ubuntu 14.04 by default.  100% CPU consumption is common.  

  As far as I know, we don't use the data collected by these daemons, so there's no reason to put up with the CPU usage.

* Chef-client configuration was recently migrated from roles to recipes.  Correspondingly, Kafka clusters must now disable the chef-client service explicitly, instead of omitting it from a role.

### DIFF
--- a/cookbooks/bcpc/recipes/packages.rb
+++ b/cookbooks/bcpc/recipes/packages.rb
@@ -28,7 +28,6 @@
   libsasl2-modules-gssapi-mit
   lzop
   p7zip-full
-  pcp
   powertop
   silversearcher-ag
   sysstat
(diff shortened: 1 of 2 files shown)


cookbooks/bcpc/recipes/packages.rb
```diff
@@ -40,3 +40,11 @@
...
   end
 end
+
+#
+# One particularly unhelpful system package -- frequently consumes
+# 100% of CPU, but we don't send the metrics anywhere.
+#
+package 'pcp' do
+  action :purge
+end
```
vt0r: In addition to purging it here, can we also remove it from line 31 so it no longer gets installed at all?

PR #1315, opened 2018-11-29
Bootstrap local

aespinosa:
While debugging with @bijugs , `bcpc::ssl_certs_deploy` needs to write to `/usr/loca/share/ca-certificates`. this fails because that directory is already mounted in cluster VMs as read-only.

Restricting it to the bootstrap should solve this.  We don't need to propagate hypervisor CAs to the actual VMs anyway.


bijugs:
Building a new VM cluster with this change.

PR #1317, opened 2018-11-30
Drop diamond

aespinosa:
Remove disk monitoring in graphite since the disk metrics are supplied
by diamond.


leochen4891:
Diamond not only provides disk metrics, but also [CPU|https://github.com/bloomberg/chef-bach/pull/1317/files#diff-9abfd340ffd8489bd5f812ee7e0b10feL186], memory, and network.
Some of the alarms we currently have in zabbix is based on memory. 
I would suggest disabling disk metrics collector rather than drop diamond as a whole.

aespinosa:
dropped the memory queries in zabbix too.

We don't have zabbix configured to alert on CPU or network so it's safe to drop as well.

PR #1325, opened 2018-12-10
Databag credentials (bootstrap)

aespinosa:
Create `bach_roles::admin` to orchestrate all the databag generation

recipe to generate machine-specific databags (mysql passwords, ssh host keys) will be in a different PR


vt0r (approved):
LGTM 👍

